PR #33, opened 2024-03-05
Update setup.py qiskit version

Nik-Stein:
qiskit version set to 0.44 in setup.py


bastibock:
Install older version of Qiskit in setup.py to fix compatibility issue. Will be changed do latest Qiskit version, when compatibility is again ensured.

PR #84, opened 2024-10-09
Add documentation and dependencies to IQM Integration

manzanillo:
This PR adds
* documentation for how to use an IQM backend in Qrisp, and what parameters to provide
* input validation for API token, device id and availability of iqm relevant dependecies
* Define IQM depencies as an extra package that can be installed using `pip install qrisp[iqm]`

I further suggest to change the default number of shots (100.000) for the measurement to a smaller number as most physical devices have a maximum number of shot that is below that (e.g. for IQM devices it is 20.000). This would result in the user not needing to specify the number of shots when using `IQMBackend`. The alternative is (as I did currently) to mention the limit in the documentation and always provide a shots parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ pip install qrisp
 ```
 Qrisp has been confirmed to work with Python version 3.8, 3.9 & 3.10.
 
+If you want to work with IQM quantum computers as a backend, you need to install additional dependencies using
+```bash
+pip install qrisp[iqm]
+```
+
 ## First Quantum Program with Qrisp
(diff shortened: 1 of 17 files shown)


manzanillo:
After speaking with @positr0nium I moved the default number of shots to the backends. As, currently and in the future, the shots can be passed down towards the backend, every layer above needs to set None as default to keep it optional and conditionally set the shots number in the backend with an if statement.

The only alternative would be to either make the shot number a parameter for the backend (which would be strange as one could intend to run multiple different experiments with different shot numbers) or to use kwargs that are passed down, which might cause unwanted side effects currently not considered.

PR #150, opened 2025-03-07
Fix problems with import

manzanillo:
If one imports the VQE algorithm as in the example given here https://qrisp.eu/reference/Algorithms/vqe/VQEProblem.html it fails as it cannot import from here:
`from qrisp.algorithms.vqe.problems import *`

This can be fixed by making the problems directory a module.


positr0nium:
This is very strange - apparently the PyPi version can not be imported but the GitHub version has worked for everyone and the testing pipeline also didn't fail

positr0nium:
Anyway thanks :)

PR #151, opened 2025-03-07
Khattar mcp implementation

diehoq:
Implement mcp based on khattar mcx, implement the test pipeline for it
Edit khattar mcx to use margolus gates and gidney mcx 
Swap ctrl qubits to reduce depht when using multiple margolus

### DIFF
--- a/src/qrisp/alg_primitives/mcx_algs/khattar.py
+++ b/src/qrisp/alg_primitives/mcx_algs/khattar.py
@@ -115,14 +115,11 @@ def cca_mcx(ctrls, target, anc):
     else:
         xrange = jrange
     # STEP 1
-    # This operation is always executed regardless of the lenght of ctrls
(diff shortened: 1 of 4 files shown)


ACE07-Sev:
Could I see the paper reference for this compilation method?

positr0nium:
There are some additional optimizations we found: In particular it is possible to use the Margolus gate (https://arxiv.org/abs/quant-ph/0312225) for most toffolis because the mcx has the form of a conjugation circuit: U V U^dg, implying the Margolus phases cancel out. Furthermore it's possible to gain some depth advantage by aligning the control knobs of the Margolus gate with it's implementation to avoid delay.
Finally, the mcp implementation is also based on the insight that the mcx is a conjugation but instead of conjugating a Toffoli gate, we conjugate a phase gate.

PR #155, opened 2025-03-20
2nd order trotterization option

Sam-XiaoyueLi:
This PR added 2nd order trotterization as an optional parameter `order`, where the 2nd order Trotter-Suzuki formula is given by

$$
e^{it(A+B)} \approx(e^{i\frac{t}{2N}A} e^{i\frac{t}{N}B} e^{i\frac{t}{2N}A})^N 
$$

### DIFF
--- a/src/qrisp/operators/qubit/qubit_operator.py
+++ b/src/qrisp/operators/qubit/qubit_operator.py
@@ -1654,7 +1654,7 @@ class QubitOperator(Hamiltonian):
     # Trotterization
     #
 
-    def trotterization(self, method='commuting_qw', forward_evolution = True):
+    def trotterization(self, order=1, method='commuting_qw', forward_evolution = True):
         r"""
         .. _ham_sim:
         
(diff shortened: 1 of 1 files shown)


positr0nium:
Thanks for this contribution! :)
I think it would be good if reduce the redundancy of the implementation a bit - this could potentially make it easier for us to debug/maintain/optimize the code in the future.
If I understand it correctly the main difference between the first order and the second order is that there are two blocks that essentially execute the order 1 `trotter_step` function where one is executed in reversed order. 
Please correct me if I'm wrong but wouldn't this perform the same circuit:

```
trotter_step(qarg, t, steps*2)
with invert():
    trotter_step(qarg, -t, steps*2)
```

Where `trotter_step` is still the first order trotter step.

Sam-XiaoyueLi:
> Thanks for this contribution! :) I think it would be good if we reduce the redundancy of the implementation a bit - this could potentially make it easier for us to debug/maintain/optimize the code in the future. If I understand it correctly the main difference between the first order and the second order is that there are two blocks that essentially execute the order 1 `trotter_step` function where one is executed in reversed order. Please correct me if I'm wrong but wouldn't this perform the same circuit:
> 
> ```
> trotter_step(qarg, t, steps*2)
> with invert():
>     trotter_step(qarg, -t, steps*2)
> ```
> 
> Where `trotter_step` is still the first order trotter step.

Thanks Raphael! I've made changes according to your suggestion. Indeed, it works exactly the same and is much shorter. However, I was not able to find the test file to cover it and I'm not sure if it's required.

PR #181, opened 2025-04-16
Switch upgrades

renezander90:
case_function_list can be list[callable] or callable


positr0nium:
The case distinction for tracing mode is not necessary, or? Please check and afterwards merge - thanks!

cjn10:
I added another qswitch implementation, see #184.

PR #202, opened 2025-05-27
Qswitch tree case number

cjn10:
Adds an argument to the `qswitch` to specify the number of cases. Defaults to `None` to automatically set the number of cases. In the case where the case number is not a power of two,  the tree traversal properly terminates saving resources compared to a full tree traversal.


positr0nium:
Nice work! :) I am not sure however if this directly works like this with the ``parallel`` method. Could you check that? You might need to use the ``permit_missmatching_size`` keyword.

PR #205, opened 2025-06-03
Jasp-compatible Montgomery arithmetics

cjn10:
Adds jasp-compatible (and non-jasp-compatible) functions for montgomery arithmetics. Additionally, removes the length-check in the `qq_gidney_adder`.

### DIFF
--- a/documentation/source/reference/Jasp/Scalable Integer Type.rst
+++ b/documentation/source/reference/Jasp/Scalable Integer Type.rst
@@ -27,6 +27,30 @@ Here's a basic example that demonstrates the use of ``BigInteger``:
 
 Note that we use the ``__call__`` method (``x()``) to approximate the ``BigInteger`` as a nicely-readable float.
 
+The ``BigInteger`` natively works with the ``QuantumModulus``:
+
+.. code-block:: python
+
+    from qrisp import *
+
+    @boolean_simulation
+    def main():
+        N = BigInteger.create_static(340282366762482138434845932244680310783, size=4)
+        a = BigInteger.create_static(131313, size=4)
+
+        qm = QuantumModulus(N)
+        qm[:] = 1
+
+        for i in jrange(10):
+            qm *= a
+
+        return measure(qm)()
+
+    print(main())
+    # Output: 1.63367308e+38
+
+Here, we use the ``create_static`` method to transform a Python integer that is larger than 64 bits into a ``BigInteger``.
+
 Class Documentation
 -------------------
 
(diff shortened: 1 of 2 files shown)


positr0nium:
Regarding the big_int initialization: You're right, the presented example doesn't reproduce the error. No idea why I posted that. This one however gives me an error:
```
b_int = 532_323_451_122_411_593_564_574_574
```

renezander90:
There is a jasp_cq_gidney_adder and jasp_cq_gidney_adder and gidney_adder for non-jasp mode.

We could introduce a jasp_gidney_adder which handles the case distinction between qq and qc, or do this case distinction within the gidney_adder (with some check_for_tracing_mode()). **Then we do not have to make any distinction elsewhere.**
(Note that using the same adder for jasp and non-jasp likely could cause issues for edge cases .)

renezander90:
An example like this, demonstrating how BigInteger can be used with QuantumModulus, would be helpful in the Scalable Integer Type documentation.

```
from qrisp import *

@boolean_simulation
def main():

    N = BigInteger.create_static(340282366762482138434845932244680310783)
    a = BigInteger.create_static(131313)

    qm = QuantumModulus(N)
    qm[:] = 1

    for i in jrange(10):
        qm *= a

    return measure(qm)()

main()
```

However, even running a smaller example, like this:
```
from qrisp import *

@boolean_simulation
def main():

    N = BigInteger.create_static(340282366762482138434845932244680310783)

    qm = QuantumModulus(N)
    qm[:] = 1
    qm *= 21

    return measure(qm)()

main()
```

Yields the correct result, but: 
```
WARNING: Faulty uncomputation found during simulation.
WARNING: Faulty uncomputation found during simulation.
WARNING: Faulty uncomputation found during simulation.
WARNING: Faulty uncomputation found during simulation.
WARNING: Faulty uncomputation found during simulation.
WARNING: Faulty uncomputation found during simulation.

Array(21., dtype=float64)
```

src/qrisp/core/quantum_array.py
renezander90, line 1: The cases `isinstance(self.qtype, QuantumModulus)` and `isinstance(self.qtype, QuantumFloat)` in matmul must be swapped since a QuantumModulus is a subtype of QuantumFloat. But, after doing this new issues occur for cq matmul with QuantumArray of type QuantumModulus (e.g. for examples provided comments).

cjn10: Fixed in f77dd8c946f1d7424b829ecd6ed9c3df072457ca.

PR #219, opened 2025-06-24
Update iqm client dependecy

manzanillo:
This PR updates the IQM client dependency


positr0nium:
Fixed in a separate merge - thanks for the contribution :)

PR #227, opened 2025-07-02
Batched backend

positr0nium:
This class tackles the problem that many physical backends have a high overhead regarding individual circuit execution. This overhead typically comes from finite network latency, authentication procedures, compilation steps etc. Typically this overhead is remedied through supporting the execution of batches of circuits, which however doesn't really fit that well into the Qrisp programming model, which shields the user from handling individual circuits and automatically decodes the measurement results into human readable labels.
    
In order to bridge these worlds and still allow automatic decoding, the ``BatchedBackend`` allows Qrisp users to evaluate measurements from a multi-threading perspective. The idea is here that the circuit batch is collected through several threads, which each execute Qrisp code until a individual backend call is required. This backend call is then saved until the batch is complete. The batch can then be sent through the ``.dispatch`` method, which resumes each thread to execute the post-processing logic.

Besides these features, this PR implements batched measurement behavior for QubitOperator expectation_value measurements.

Example: 

```
def run_func_batch(batch):
    """
    Parameters
    ----------
    batch : list[tuple[QuantumCircuit, int]]
        The circuit and shot batch indicating the backend queries.

    Returns
    -------
    results : list[dict[string, int]]
        The list of results.

    """
    
    results = []
    for i in range(len(batch)):
        qc = batch[i][0]
        shots = batch[i][1]
        results.append(qc.run(shots = shots))
        
    return results

from qrisp import *
from qrisp.interface import BatchedBackend
# Create some backend calls        
a = QuantumFloat(4)
b = QuantumFloat(3)
a[:] = 1
b[:] = 2
c = a + b

d = QuantumFloat(4)
e = QuantumFloat(3)
d[:] = 2
e[:] = 2
f = d + e

# Set up batched backend
bb = BatchedBackend(run_func_batch)

# Create threads
import threading

results = []
def eval_measurement(qv):
    results.append(qv.get_measurement())

thread_0 = threading.Thread(target = eval_measurement, args = (c,))
thread_1 = threading.Thread(target = eval_measurement, args = (f,))

# Start the threads
thread_0.start()
thread_1.start()

# Call the dispatch routine
# The min_calls keyword will make it wait 
# until the batch has a size of 2
bb.dispatch(min_calls = 2)

# Wait for the threads to join
thread_0.join()
thread_1.join()

# Inspect the results
print(results)
```

### DIFF
--- a/documentation/source/reference/Backend Interface/BatchedBackend.rst
+++ b/documentation/source/reference/Backend Interface/BatchedBackend.rst
@@ -1,0 +1,9 @@
+.. _BatchedBackend:
+
+BatchedBackend
+==============
+
+.. currentmodule:: qrisp.interface
+
+.. autoclass:: BatchedBackend
+
--- a/documentation/source/reference/Backend Interface/index.rst
+++ b/documentation/source/reference/Backend Interface/index.rst
@@ -8,6 +8,7 @@ Backend Interface
    BackendServer
    BackendClient
    VirtualBackend
+   BatchedBackend
    DockerSimulators
    QiskitBackend
    IQMBackend
--- a/documentation/source/reference/Utilities.rst
+++ b/documentation/source/reference/Utilities.rst
@@ -8,6 +8,7 @@ Utilities
    :toctree: generated/
 
    multi_measurement
+   batched_measurement
    gate_wrap
    custom_control
    lifted
--- a/documentation/source/reference/generated/qrisp.batched_measurement.rst
+++ b/documentation/source/reference/generated/qrisp.batched_measurement.rst
@@ -1,0 +1,6 @@
+﻿qrisp.batched\_measurement
+==========================
+
+.. currentmodule:: qrisp
+
+.. autofunction:: batched_measurement
--- a/src/qrisp/circuit/compilation_acceleration.py
+++ b/src/qrisp/circuit/compilation_acceleration.py
@@ -17,6 +17,7 @@
 """
 
 import numpy as np
+import threading
 
 from qrisp.circuit.quantum_circuit import QuantumCircuit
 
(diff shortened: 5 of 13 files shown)


renezander90:
From a user's perspective the following function would be helpful:

```
def batched_measurement(variables, batched_backend, shots=None):
    """
    This functions facilitates the measurement of multiple QuantumVariables with a BatchedBackend.

    Parameters
    ----------
    variables : list[QuantumVariable]
        A list of QuantumVariables.
    batched_backend : BatchedBackend
        The backend to evaluate the compiled QuantumCircuits on. 
    shots : int, optional
        The amount of shots to perform. The default is given by the backend used.
        
    Returns
    -------
    results : list[dict]
        The list of results.
    
    """

    import threading

    results = []
    def eval_measurement(qv):
        results.append(qv.get_measurement(backend = batched_backend, shots = shots))

    threads = []
    for var in variables:
        thread = threading.Thread(target = eval_measurement, args = (var,))
        threads.append(thread)

    # Start the threads
    for thread in threads:
        thread.start()

    # Call the dispatch routine
    # The min_calls keyword will make it wait 
    # until the batch has a size of number of variables
    batched_backend.dispatch(min_calls = len(variables))

    # Wait for the threads to join
    for thread in threads:
        thread.join()

    # Inspect the results
    return results
```

Then the batched backend can be used as follows:

```
from qrisp import *
from qrisp.interface import BatchedBackend

def run_func_batch(batch):
    # Parameters
    # ----------
    # batch : list[tuple[QuantumCircuit, int]]
    #     The circuit and shot batch indicating the backend queries.

    # Returns
    # -------
    # results : list[dict[string, int]]
    #     The list of results.

    results = []
    for i in range(len(batch)):
        qc = batch[i][0]
        shots = batch[i][1]
        results.append(qc.run(shots = shots))

    return results

# Set up batched backend
bb = BatchedBackend(run_func_batch)

a = QuantumFloat(4)
b = QuantumFloat(3)
a[:] = 1
b[:] = 2
c = a + b

d = QuantumFloat(4)
e = QuantumFloat(3)
d[:] = 2
e[:] = 3
f = d + e

batched_measurement([c,f], batched_backend=bb)
```

How would the integration with hardware backends work? Will there be a backend and batched backend for each provider, e.g., IQMBackend, IQMBatchedBackend, or will the function creating the provider backend receive an argument indicating whether it is a batched backend?

positr0nium:
Made the following updates:
1. Implemented the feature suggested by @renezander90 (https://github.com/eclipse-qrisp/Qrisp/pull/227/commits/0147f194bd6bb550773bad33e476dfc8e8a3b06b)
2. If .run is called from the main thread, all queries are dispatched. BatchedBackends therefore behave like normal backends if queried from the main thread (https://github.com/eclipse-qrisp/Qrisp/pull/227/commits/547f5309cf42f5763d2a61267e4e5622d754a04f)
3. The IQMBackend now returns a BatchedBackend by default.

PR #231, opened 2025-07-06
Add isqrt algorithm

GeorgeKarlinzer:
Added implementation of the algorithm https://arxiv.org/abs/1712.08254. The algorithm computes a square root and a remainder of a number with no garbage and is T-Count optimized.

### DIFF
--- a/tests/test_quantum_integer_square_root.py
+++ b/tests/test_quantum_integer_square_root.py
@@ -21,7 +21,7 @@ from qrisp import QuantumFloat, multi_measurement
 import math
 
 def test_quantum_square_root():
-    for a in range(6, 26):
+    for a in range(1, 26):
         expected_root = int(math.sqrt(a))
         expected_remainder = a - expected_root**2
 


renezander90:
Thank you! This looks very good.

However, I have still some further questions/comments:

1) The q_isqrt should work for any given QuantumFloat without the user having to think about its size.
Additionally, it should work independently of the particular state of the input QuantumFloat. The state of the input QuantumFloat R could be a superposition state.

Note that QuantumVariables can be extended/reduced:
https://qrisp.eu/reference/Core/generated/qrisp.QuantumVariable.extend.html#qrisp.QuantumVariable.extend
https://qrisp.eu/reference/Core/generated/qrisp.QuantumVariable.reduce.html#qrisp.QuantumVariable.reduce

and unsigned QuantumFloat can be turned into signed QuantumFloats:
https://qrisp.eu/reference/Quantum%20Types/generated/qrisp.QuantumFloat.add_sign.html#qrisp.QuantumFloat.add_sign

```
from qrisp import *
R = QuantumFloat(8)
R[:] = 131

R.extend(1)
print(R)
print(R.size)

R.add_sign()
print(R)
print(R.size)
```
> {131: 1.0} 
> 9
> {131: 1.0}
> 10

2) Why do we require a signed QuantumFloat as input? I would rather think that the sqrt should require an unsigned QuantumFloat?

3) Could you please test if it works on superpositions of inputs, e.g., 

```
R = QuantumFloat(8)
R[:] = {131: 1, 81 : 1}
print(R)

F = R + 2 
multi_measurement([R,F])
```
> {81: 0.5, 131: 0.5}  
> {(81, 83): 0.5, (131, 133): 0.5}

4) The import `from qrisp import *` in the isqrt_algorithm.py could cause some issues. We should rather use:

```
from qrisp.circuit import Qubit
from qrisp.qtypes import QuantumFloat
from qrisp.core.gate_application_functions import x, cx, mcx, swap
```

GeorgeKarlinzer:
@renezander90  You're absolutely right, algorithm should work for any quantum variable, I've focused so much on the original paper, that the condition engraved in my mind as an absolute rule. I've modified the code a little to allow users to pass any quantum variable. `extend` feature comes very handy for this matter. Also added couple tests for superpositions input (works as intended).

Couple things that I'm not sure about:

- Algorithm proposed by Thapliyal et al. assumes that the first qubit is in 0 state and the total qubit amount is even, therefore to ensure correctness of the algorithm I always extend input variable by one or two qubits, so the condition stays satisfied. After the computation I reduce it to the initial form. If you think that this is a problem, I can try to modify the original algorithm, to work without extending. 
- About the second comment, now algorithm works with any (signed or unsigned) QuantumFloat, I don't want to disallow signed variables as an input, because it could be as well a positive number. Currently algorithm will work for negative numbers as well, but the output will be incorrect. We can try to perform conditional operation based on the sign value and do nothing if the sign qubit is 1, but it may increase complexity of the implementation. I would appreciate hearing any solutions you might have for such cases.

Summary of changes:

1. Refactor code to allow users to pass any QuantumFloats
2. Refactor code to allow users to pass signed or unsigned inputs
3. Test superposition input, everything works fine, also add a test case
4. Refactor imports

PR #243, opened 2025-07-14
Cleanup Qrisp packaging pipeline

JLenssen:
### Overview

Qrisp currently relies on a mix of `pyproject.toml` and the older `setup.cfg` for packaging. This PR aims at moving the logic to `pyproject.toml` which is [recommended](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/). Furthermore, the linting configuration is added, we switch to [ruff](https://docs.astral.sh/ruff/), and enforce the rules via pre-commit.

### List of changes

- Add: base requirements, and requirements for CI/CD and optional IQM dependency.
- Add: [pip-tools](https://pip-tools.readthedocs.io/en/latest/) to pin the versions of the requirements above.
- Add: [ruff](https://docs.astral.sh/ruff/) for linting and move config to `pyproject.toml`.
- Add: Enforcing of linting rules via [pre-commit](https://pre-commit.com/).
- Add: pre-commit format check to CI/CD pipeline (**Note:** only applies to subset of files at the moment).
- Add: Contribution guide for developers in `CONTRIBUTE.md`.
- Refactor: `qrisp_test.yml`, by renaming it to `main.yml` and adding pre-commit job.
- Remove: `.gitlab-ci.yaml` as GitLab is no longer used.

### Tasks for future PRs
Some necessary tasks were postponed to future PRs to reduce the number of files changed

- Apply linting rules to all files and correct errors thrown by ruff.
- Review when CI/CD pipeline jobs run as they take too long.
- Manage releases via tags instead of `version` in `pyproject.toml` by using [setuptools-scm](https://setuptools-scm.readthedocs.io/en/stable/usage/) and push automatically to PyPI.


ntchol:
=> The changes  to the README file should not  only contain  an explanation about the IQM module but also about the other available backends

=> The changes to the testing and CI/CD pipeline should be agreed among all contributing  partners

=> We should wait with changes to the key processes  and code/package  management  until  we  have agreed  on a governance  structure and roadmap... otherwise we will end up in chaos 

Governance structure  is WiP

Thank you!

JLenssen:
Hi @ntchol , thanks for reviewing. 

> => The changes to the README file should not only contain an explanation about the IQM module but also about the other available backends

Agreed, there are other backends like `qunicorn` that introduce dependencies (e.g. `flask`) that would be worth making optional and installable via `qrisp[qunicorn]`. The exact list of optional dependencies should be discussed.

> => The changes to the testing and CI/CD pipeline should be agreed among all contributing partners
> 
> => We should wait with changes to the key processes and code/package management until we have agreed on a governance structure and roadmap... otherwise we will end up in chaos

Understandable. I'll keep the PR in draft for now and wait for further input. Happy to pick this up when governance structure is set in stone.

I would still recommend pinning the test dependencies as those are currently only documented in the CI pipeline. Happy to open another PR for that.

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Run pre-commit
+        # FIXME: Does only run on subset of files since not all files are linted yet
+        run: pre-commit run --files pyproject.toml .pre-commit-config.yaml requirements/* --show-diff-on-failure
+
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        python-version: ["3.10.*"]
```
purva-thakre, line 32: minor suggestion: you should add more python versions given that the readme states that Qrisp is compatible with 3.10, 3.11 & 3.12.

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+          python -m pip install --upgrade pip
+          pip install -e .[dev]
+
+      - name: Run pre-commit
+        # FIXME: Does only run on subset of files since not all files are linted yet
+        run: pre-commit run --files pyproject.toml .pre-commit-config.yaml requirements/* --show-diff-on-failure
+
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
purva-thakre, line 29: I have never used this. What's the motivation behind using `timeout` in a testing workflow? The unit tests could take longer than 60 minutes, right?

requirements/base.txt
```diff
@@ -0,0 +1,119 @@
+# This file was autogenerated by uv via the following command:
+#    uv pip compile requirements/base.in -o requirements/base.txt
+blinker==1.9.0
```
purva-thakre, line 3: it is better to use inequality dependency specifiers for version syntax rather than `==`. 

https://packaging.python.org/en/latest/specifications/dependency-specifiers/

If a [dependabot workflow](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) is added in the future, `==` can create excessive dependabot PRs.

PR #246, opened 2025-07-23
Convert tutorials to ipynb

manzanillo:
This PR convertts the tutorials to ipynb and makes them downloadable as those, will ping when ready

### DIFF
--- a/documentation/source/general/tutorial/Shor.ipynb
+++ b/documentation/source/general/tutorial/Shor.ipynb
@@ -26,17 +26,6 @@
   },
   {
    "cell_type": "markdown",
-   "id": "bfa082dd",
-   "metadata": {},
-   "source": [
-    ".. _shor_tutorial:\n",
-    "\n",
-    "# asdfa\n",
-    "(shor_tutorial)="
-   ]
-  },
-  {
-   "cell_type": "markdown",
    "id": "86e33752",
    "metadata": {},
    "source": [


manzanillo:
@renezander90 @MatP1337 @positr0nium notebooks all converted, download link added, i recommend we add a pipeline that continously runs them (e.g. I corrected the missing shape parameter for QuantumArray, these things would then become directly apparent), i say this should be a new PR though

renezander90:
Thank you! This looks very good:).

However, there are still some issues:

1) When building the documentation I receive 177 warnings, mostly of the following two types:

a) Links from a tutorial to another page of the documentation:
> WARNING: File not found: '/reference/Quantum Types/QuantumModulus.html' [nbsphinx.localfile]

I think we must change these links in the notebooks to the full weblink, i.e.,
https://www.qrisp.eu/reference/Quantum%20Types/QuantumModulus.html ?

b) Links form the documentation to the tutorials:
> WARNING: undefined label: 'shor_tutorial'

I tried to fix this by adding a raw field with the label at the beginning of the notebook, but it did not work (maybe I was doing something wrong?) Do you know how to fix this?

2) There are some minor issues in the notebooks, such as redundant cells (i.e., we have the output cell for the code, and then an additional hardcopy of the output below).
**I would review all notebooks and correct these issues.**

I think it will be more practical if we first merge this PR into the doc_ipynb_new branch, and after addressing all issues it will be merged into main.

PR #249, opened 2025-07-30
Jaspr to ClosedJaxpr

positr0nium:
This PR changes the Jasp architecture with regard to the class inheritance of ``Jaspr``. The ``Jaspr`` class no longer inherits from regular ``Jaxpr`` but ``ClosedJaxpr`` now. 

What is the difference? When importing a numpy array, a simple list or similar things into Jax, this is traced as "constants" that are treated similar to the function parameters of the ``Jaxpr`` class.

```
from jax import make_jaxpr
import jax.numpy as jnp

def main(b):
    a = jnp.array([1,2,3])
    return a+b

closed_jaxpr = make_jaxpr(main)(4)
print(closed_jaxpr)
```
gives 
```
{ lambda a:i64[3]; b:i64[]. let
    c:i64[] = convert_element_type[new_dtype=int64 weak_type=False] b
    d:i64[3] = add a c
  in (d,) }
```
The constants variables are specified in the signature before the semicolon (i.e. ``a:i64[3]``). They can be accessed through the ``constvars`` attribute.
```
print(closed_jaxpr.jaxpr.constvars)
# Yields [Var(id=2154348981440):int64[3]]
```
The difference between a ``Jaxpr`` and a ``ClosedJaxpr`` is now that the latter also contains concrete values that go into the constant variables:
```
print(closed_jaxpr.consts)
# Yields [array([1, 2, 3])]
```
Previously the ``Jaspr`` class was inherting from ``Jaspr`` and handle the concrete values of constants separately. This was bad design, which is fixed now, i.e. ``Jaspr`` inherits from ``ClosedJaxpr``.

### DIFF
--- a/src/qrisp/jasp/jasp_expression/centerclass.py
+++ b/src/qrisp/jasp/jasp_expression/centerclass.py
@@ -45,7 +45,7 @@ class Jaspr(ClosedJaxpr):
     `established, classical compilation infrastructure <https://mlir.llvm.org/>`_,
     implying state-of-the-art compilation speed can be reached.
 
-    As a subtype of ``jax.core.Jaxpr``, Jasprs are embedded into the well matured
+    As a subtype of ``jax.extend.core.ClosedJaxpr``, Jasprs are embedded into the well matured
     `Jax ecosystem <https://github.com/n2cholas/awesome-jax>`_,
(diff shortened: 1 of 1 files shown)


src/qrisp/jasp/jasp_expression/centerclass.py
renezander90, line 1: Line 48: As a subtype of ``jax.core.ClosedJaxpr``

PR #250, opened 2025-07-31
Rename QITE → DB_QITE to support multiple QITE methods

Sam-XiaoyueLi:
This PR renames the existing `QITE` implementation to `DB_QITE` (Double-Bracket Quantum Imaginary-Time Evolution) to clarify that it’s one specific QITE variant. The old name "QITE" could be misleading if Qrisp plans to introduce additional imaginary-time evolution methods, so this change makes room for a top-level QITE factory that can dispatch on `method=` in the future.


src/qrisp/algorithms/qite.py
```diff
@@ -24,7 +24,18 @@
 import jax.numpy as jnp
 
 
-def QITE(qarg, U_0, exp_H, s, k, method="GC"):
+def QITE(qarg, procedure_type=None, *kwargs):
+    r"""
+    To anticipate other methods implementing QITE
```
marekgluza, line 29:
```suggestion
    To anticipate other methods implementing QITE.
```

Issue #252, opened 2025-07-31
Bug: Interactive kernel in documentation fails to launch


purva-thakre:
Each tutorial in the docs, has a link on the right to start an interactive kernel. This does not work on any of the tutorial pages.

<img width="251" height="171" alt="Image" src="https://github.com/user-attachments/assets/2df2b769-6b8b-46d1-8ce3-a849c4c72326" />

I have checked this is not an issue related to my browser as I can activate the interactive portion in thebe's documentation. 

https://thebe.readthedocs.io/en/stable/

positr0nium:
This is fixed here: 833b75109f428555cf595ec9d4344abfa72b126c
However there is still some style issue. Namely starting the thebe kernel renders the code cells in an extra html class, which results in the code being "framed" twice. If anyone knows how to fix this feel free (@renezander90 @manzanillo). Otherwise I think it is so minor that we can ignore it.

PR #253, opened 2025-08-01
use only input fields for thebe, remove border

manzanillo:
Adresses #252 
and also adresses a bug where also output elements are used and rendered for thebe (executing them would just replicate the output, but is not helpful, so I made the selector to be more specific)


positr0nium:
Gracias :)

PR #274, opened 2025-09-15
run tests on pull requests

cjn10:
Changes the trigger in the GitHub workflow to run tests for pushes and pull requests targeting `main`. Also, running tests is limited to `main` to avoid duplicate runnings. To run tests for a branch, create a (draft) pull request for this branch.


renezander90:
Actually, running the pipeline for push to any branch is a useful. Can we change to always run for push and run for PR to main? This should cause only a few duplicate runs, i.e., if a PR from a branch to main is created.

PR #279, opened 2025-09-22
Arithmetic tools (jasp-compatible)

diehoq:
This PR introduces a set of arithmetic utility functions for QuantumFloats

**New Functions Added:**
1. q_min – returns the minimum of two QuantumFloat values
2. q_max – returns the maximum of two QuantumFloat values
3. q_floor – computes the floor of a QuantumFloat
4. q_ceil – computes the ceiling of a QuantumFloat
5. q_fractional – extracts the fractional part of a QuantumFloat
6. q_round – rounds a QuantumFloat to the nearest integer
7. q_modf – splits a QuantumFloat into integer and fractional components

**Additional Updates**:

Comprehensive documentation for each new function
Corresponding unit tests

### DIFF
--- a/tests/jax_tests/test_jasp_QuantumFloat.py
+++ b/tests/jax_tests/test_jasp_QuantumFloat.py
@@ -1,0 +1,34 @@
+"""
+********************************************************************************
+* Copyright (c) 2025 the Qrisp authors
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+def test_jasp_QuantumFloat():
+
+    # Test decoder for QuantumFloat (Issue #271) 
+    from qrisp import QuantumFloat, h
+    from qrisp.jasp import terminal_sampling
+
+    @terminal_sampling
+    def main():
+        a = QuantumFloat(3, -2, signed = True) 
+
+        h(a)
+
+        return a
+    
+    res = main()
+    assert res == {-2.0: 0.0625, -1.75: 0.0625, -1.5: 0.0625, -1.25: 0.0625, -1.0: 0.0625, -0.75: 0.0625, -0.5: 0.0625, -0.25: 0.0625, 0.0: 0.0625, 0.25: 0.0625, 0.5: 0.0625, 0.75: 0.0625, 1.0: 0.0625, 1.25: 0.0625, 1.5: 0.0625, 1.75: 0.0625}


src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
```diff
@@ -0,0 +1,329 @@
...
+    >>> b = q_floor(a)
+    >>> b.get_measurement()
+    {1.0: 0.75, 0.0: 0.25}
+    """
+
+    b = a.duplicate()
+    for i in jrange(-a.exponent, a.size):
+        cx(a[i], b[i])
+
+    return b
```
positr0nium, line 205: By using duplicate, you create a QuantumFloat that could potentially have non-integer values. The result of floor can however be an integer only. I guess it depends on the usecase whether this is unnecessary. What do you think @renezander90 @cjn10?

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
```diff
@@ -0,0 +1,329 @@
...
+    >>> b = QuantumFloat(2)
+    >>> a += 2
+    >>> h(a[0])
+    >>> b+=1
+    >>> res_max = q_max(a,b)
+    >>> multi_measurement([a,b,res_max])
+    {(2, 1, 2): 0.5, (3, 1, 3): 0.5}
+    """
+
+    res = QuantumFloat(a.size)
```
positr0nium: I think this can be made a bit more robust :)
```
res_mshape = [jnp.minimum(a.mshape[0], b.mshape[0]), jnp.maximum(a.mshape[1], b.mshape[1])]
new_msize = res_mshape[1] - res_mshape[0]
new_exponent = res_mshape[0]
res = QuantumFloat(new_msize, new_exponent)
```

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
```diff
@@ -0,0 +1,329 @@
...
+    >>> a = QuantumFloat(2)
+    >>> b = QuantumFloat(2)
+    >>> a += 2
+    >>> h(a[0])
+    >>> b+=1
+    >>> res_min = q_min(a,b)
+    >>> multi_measurement([a,b,res_min])
+    {(2, 1, 1): 0.5, (3, 1, 1): 0.5}
+    """
+    res = QuantumFloat(a.size)
```
positr0nium: Same here as in `q_max`

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
```diff
@@ -0,0 +1,329 @@
...
+
+    res = QuantumFloat(a.size)
+    c = a >= b
+
+    with control(c):
+        for i in jrange(a.size):
+            cx(a[i], res[i])
+    with control(c, 0):
+        for i in jrange(b.size):
+            cx(b[i], res[i])
```
positr0nium: This might be more efficient like this:

```
cx(a, res)

with conjugate(cx)(a,b):
    with control(c, 0):
        cx(b, res)
```

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
```diff
@@ -0,0 +1,329 @@
...
+
+    compare_func = lambda x, y: x <= y
+
+    with invert():
+        (c << compare_func)(a, b)
```
positr0nium: Might be more elegant and give an in-built custom control to do this via conjugation

positr0nium:
Nice work! Added a few comments within the changed files :) Thanks

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
renezander90: Yes :) This will now work for QuantumFloats of different shape. (But, not yet for signed.)

Could work for signed with:
```
res = QuantumFloat(new_msize, new_exponent, signed=a.signed | b.signed)
```

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
renezander90: But, this wouldn't work if an and b have different shape. With the proposed changes from above, this should work for QuantumFloats of different shape:

```
    with control(c):
        for i in jrange(a.size):
            cx(a[i], res[i + a.exponent - new_exponent])
    with control(c, 0):
        for i in jrange(b.size):
            cx(b[i], res[i + b.exponent - new_exponent])
```

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
diehoq: I've thought about this and I think that, since the operation is out of place and the initial variable still exists, it makes more sense to preserve the same structure in the output variable in order to allow operations between them.

src/qrisp/alg_primitives/arithmetic/arithmetic_tools.py
positr0nium: The efficient approach should work on the digits where ``a`` and ``b`` overlap. On the disjoint digits, the straightforward way has to be employed.

diehoq: The extreme cases of the approach you suggest - no overlap and maximal overlap - will lead to a situation in which the cx is called with two empty lists of qubits. In this case the cx should not be appended to the circuit, but as of now this behaviour is simply not allowed.

PR #293, opened 2025-10-16
Improving `Qrisp` -> `Pennylane` conversion function

PietropaoloFrisoni:
**Context**: The conversion function from a Qrisp to a PennyLane circuit could be improved. The purpose of this PR is to improve the existing conversion process.

**Description of the Change**: As above.

- Simplified logic and improved performance (good speed up for the previously existing randomized test)
- Support for abstract parameters, mid-circuit measurements, controlled and adjoint gates
- Refactored the test structure and added several more tests
- Fixed a bug that caused an error with NumPy within PL (thanks @positr0nium)

**Benefits**: This conversion function is more robust and efficient

**Possible Drawbacks**: Several additional features could be implemented on top of those mentioned above (further gates, support for control-flow operations, etc.). We leave all these cases for future development.

The same goes for the tests. I tried to test several cases as systematically as possible, but a more organized and logical structure would be required. It is probably premature to overthink this step for this specific task now

**Related GitHub Issues**: None

### DIFF
--- a/tests/interface_tests/test_converter_qrisp_to_qml.py
+++ b/tests/interface_tests/test_converter_qrisp_to_qml.py
@@ -21,7 +21,7 @@ import pennylane as qml
 import pytest
 from sympy import symbols
 
-from qrisp import QuantumVariable
+from qrisp import QuantumFloat, QuantumVariable, auto_uncompute, h, z
 from qrisp.circuit.standard_operations import (  # Barrier,
     CPGate,
     CXGate,
@@ -49,6 +49,7 @@ from qrisp.circuit.standard_operations import (  # Barrier,
     ZGate,
     u3Gate,
 )
+from qrisp.grover import diffuser
 from qrisp.interface import qml_converter
 
 SINGLE_GATE_MAP = [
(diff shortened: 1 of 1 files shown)


tests/interface_tests/test_converter_qrisp_to_qml.py
```diff
@@ -0,0 +1,888 @@
...
+    (SGate, qml.S, []),
+    (TGate, qml.T, []),
+    (HGate, qml.Hadamard, []),
+    (u3Gate, qml.U3, [np.pi / 2, np.pi / 4, np.pi / 8]),
+    (IDGate, qml.Identity, []),
+    (U1Gate, qml.RZ, [np.pi / 6]),
+]
+
+# These are gates that map to the same PennyLane gate but with different parameters
+SINGLE_GATE_MAP_DIFF_PARAMS = [
```
PietropaoloFrisoni, line 72: `SINGLE_GATE_MAP ` and `SINGLE_GATE_MAP_DIFF_PARAMS ` could be combined into a unique test.

There are several potential improvements like this, but it does not seem like a big priority right now. Pleaes let me know what you think : )

src/qrisp/interface/converter/pennylane_converter.py
```diff
@@ -17,201 +17,218 @@
 """
 
 import types
+from dataclasses import dataclass
+from typing import Callable, Optional
 
 import numpy as np
+import pennylane as qml
```
positr0nium: By having this import here, we basically make pennylane a requirement, no? We should either move the import inside the conversion function or move the import of the conversion funtion into ``QuantumCircuit.to_pennylane``.

src/qrisp/interface/converter/pennylane_converter.py
```diff
@@ -17,201 +17,218 @@
...
-    import pennylane as qml
-
-    def circuit(*args, wires=None):
-
-        circFlag = False
-
-        qbit_dic = {}
-        # add qubits
+@dataclass
+class QMLGateDescriptor:
```
positr0nium, line 46: Thats a good way to handle it :)

PietropaoloFrisoni: It's not super general, but it avoids creating special cases after the first (and unique) assignment of parameters 👍

src/qrisp/interface/converter/pennylane_converter.py
PietropaoloFrisoni: Thanks, good point!

There is also another way, which is the trick that we used in PennyLane (I used it, for example, in [this file](https://github.com/PennyLaneAI/pennylane/blob/702c81dc0e0e740167ad93283297110faeabf53a/pennylane/devices/default_tensor.py#L56) as `DefaultTensor` requires `quimb`, which is not a PL dependency).

This allows having imports always on top, but the user would get an error message [like this](https://github.com/PennyLaneAI/pennylane/blob/702c81dc0e0e740167ad93283297110faeabf53a/pennylane/devices/default_tensor.py#L392) if he/she tries to use `qml_converter` without PennyLane installed. What do you think?

positr0nium: Yes that looks like a good solution :)

positr0nium:
Ahh one more thing: I think it would be good to test the conversion in some "larger" algorithm (for instance the Grover's algorithm from the 101 tutorial). Testing these larger instances usually immediately covers a lot of edge-cases.

PR #294, opened 2025-10-17
CKS algorithm

MatP1337:
This PR implements the Childs–Kothari–Somma (CKS) linear systems algorithm using the Chebyshev approach.

To-do:
-------
- [x]  ~Improve documentation~
    - [x] ~proofread docstrings~
    - [x] ~add examples to CKS function (one for A unitary, one for A as a block encoding tuple)~
    - [x] ~fix the compiled html (CKS.rst)~
    - [x] ~extend the description in CKS.rst~
    - [x] ~add example to inner_CKS showing how to run CKS with post-selection~

### DIFF
--- a/src/qrisp/algorithms/cks.py
+++ b/src/qrisp/algorithms/cks.py
@@ -713,7 +713,7 @@ def CKS(A, b, eps, kappa=None, max_beta=None):
     corresponds to, e.g., an Ising or a Heisenberg Hamiltonian, the number of Pauli terms may not scale favorably in general.
     For certain sparse matrices, their structure can be exploited to construct more efficient block-encodings.
 
-    In this example, we construct a block-encoding representation for a tridiagonal 3-sparse matrix defined by
+    In this example, we construct a block-encoding representation for the following tridiagonal sparse matrix:
 
     ::
 
@@ -726,6 +726,16 @@ def CKS(A, b, eps, kappa=None, max_beta=None):
         N = 8
         A = tridiagonal_shifted(N, mu=3)
         b = np.random.randint(0, 2, size=N)
+
+        print(A)
+        # [[ 5.  0.  0.  0. -2.  0.  0.  0.]
+        #  [ 0.  5.  0.  0.  0. -2.  0.  0.]
+        #  [ 0.  0.  5.  0.  0.  0. -2.  0.]
+        #  [ 0.  0.  0.  5.  0.  0.  0. -2.]
+        #  [-2.  0.  0.  0.  5.  0.  0.  0.]
+        #  [ 0. -2.  0.  0.  0.  5.  0.  0.]
+        #  [ 0.  0. -2.  0.  0.  0.  5.  0.]
+        #  [ 0.  0.  0. -2.  0.  0.  0.  5.]]
 
     This matrix can be decomposed using three unitaries: the identity $I$, and two shift operators $V\colon\ket{k}\\rightarrow-\ket{k+N/2 \mod N}$ and $V^{\dagger}\colon\ket{k}\\rightarrow-\ket{k-N/2 \mod N}$.
     We define their corresponding functions:


src/qrisp/algorithms/cks.py
```diff
@@ -0,0 +1,528 @@
...
+                RU(in_case, operand)
+            for i in jrange(1, j_0):
+                z(out_case[i])
+                with control(out_case[i]):
+                    RU(in_case, operand)
+                    RU(in_case, operand)
+
+    return operand, in_case, out_case
+
+def inner_CKS_wrapper(qlsp, eps, kappa=None, max_beta=None):
```
MatP1337, line 481: @positr0nium is there another way to pass numpy arrays without having to use this wrapper workaround?

src/qrisp/operators/qubit/qubit_operator.py
```diff
@@ -2048,3 +2053,143 @@ def state_prep(case):
...
+
+    def pauli_block_encoding(self):
+        r"""
+        Returns a block encoding of the operator.
+
+        A block encoding of a Hamiltonian $H$ (acting on a Hilbert space $\mathcal H_s$) is a pair of unitaries $(U,G)$, 
+        where $U$ is the block encoding unitary acting on $\mathcal H_a\otimes H_s$ (for some auxiliary Hilbert space $\mathcal H_a$), 
+        and $G$ prepares the block encoding state $\ket{G}_a=G\ket{0}_a$ in the auxiliary variable such that $(\bra{G}_a\otimes\mathbb I_s)U(\ket{G_a}\otimes\mathbb I_s)=H$.
+        Here $\mathbb I_s$ denotes the identity acting on $\mathcal H_s$.
+
```
positr0nium, line 2066: While this defintion is of course mathematically sound, to me it was very helpful to understand that a block encoding is a non-unitary matrix $H$ embedded as a block matrix
```math
\begin{pmatrix}
 H & *\\
 * & * 
 \end{pmatrix}
 ```
Maybe it could be helpful to elaborate this here.

EDIT: This is only true, if |G> = |0>. But maybe it could still help understanding.

File: src/qrisp/operators/qubit/qubit_operator.py [hunk 3]
```diff
@@ -588,7 +588,7 @@ def from_numpy_array(cls, numpy_array, threshold=np.inf):
         return H
 
     @classmethod
-    def from_matrix(self, matrix):
+    def from_matrix(self, matrix, reversed=False):
```
positr0nium: Maybe rename this parameter to ``reverse_endianness``?

src/qrisp/operators/qubit/qubit_operator.py
```diff
@@ -2048,3 +2053,143 @@ def state_prep(case):
...
+        and $G$ prepares the block encoding state $\ket{G}_a=G\ket{0}_a$ in the auxiliary variable such that $(\bra{G}_a\otimes\mathbb I_s)U(\ket{G_a}\otimes\mathbb I_s)=H$.
+        Here $\mathbb I_s$ denotes the identity acting on $\mathcal H_s$.
+
+        For a Pauli block encoding, consider an $n$-qubit Hamiltonian expressed as linear combination of Pauli operators
+
+        .. math::
+    
+            H = \sum_{i=0}^{T-1}\alpha_iP_i
+
+        where $\alpha_i$ are real coefficients, $P_i$ are Pauli operators, and $T$ is the number of terms.
```
positr0nium: Pauli operator -> Pauli string

src/qrisp/operators/qubit/qubit_operator.py
```diff
@@ -2048,3 +2053,143 @@ def state_prep(case):
             coefficients.append(np.abs(coeff_))
 
         return unitaries, np.array(coefficients, dtype=float)
+
+    def pauli_block_encoding(self):
+        r"""
```
positr0nium, line 2058: It would be good if there is some paper reference here to compare the conventions etc.

src/qrisp/operators/qubit/qubit_operator.py
```diff
@@ -2048,3 +2053,143 @@ def state_prep(case):
...
+            A function ``U(operand, case)`` applying the block encoding unitary $U$ to ``operand`` and ``case`` QuantumVariables.
+        state_prep : function
+            A function ``state_prep(case)`` preparing the block encoding state $\ket{G}$ in an auxiliary ``case`` QuantumVariable.
+        num_qubits : int
+            The number of qubits of the auxiliary ``case`` QuantumVariable.
+
+        Examples
+        --------
+
+        We apply a Hermitian matrix to a quantum state via a Pauli block encoding.
```
positr0nium, line 2116: It would probably be helpful to explain a bit more why applying RUS here gives the observed behavior

src/qrisp/operators/qubit/qubit_operator.py
```diff
@@ -2048,3 +2053,143 @@ def state_prep(case):
...
+        The belonging block encoding state is
+
+        .. math::
+
+            \ket{G} = \sum_{i=0}^{T-1}\sqrt{\alpha_i}\ket{i}.
+       
+        Returns
+        -------
+        U : function
+            A function ``U(operand, case)`` applying the block encoding unitary $U$ to ``operand`` and ``case`` QuantumVariables.
```
positr0nium: Many other functions in Qrisp that have one permeable and one "operated on" type of argument follow the convention that the permeably argument comes first in the signature (for instancee ``cx`` or every adder). Could help to continue this :).
Otherwise I really like the way that you abstracted the concept of a Pauli-Block encoding into code. The interface is well designed :)

src/qrisp/algorithms/cks.py
```diff
@@ -0,0 +1,821 @@
...
+    if max_beta == None:
+        beta = kappa**2 * np.log(kappa / eps)
+    else:
+        beta = min(kappa**2 * np.log(kappa / eps), max_beta)
+    j_0 = np.sqrt(beta * np.log(4 * beta / eps))
+    return int(j_0), int(beta)
+
+
+def cheb_coefficients(j0, b):
+    """
```
positr0nium, line 117: It would be good to have some reference here

src/qrisp/algorithms/cks.py
```diff
@@ -0,0 +1,821 @@
...
+    measure,
+)
+from qrisp.alg_primitives.reflection import reflection
+from qrisp.alg_primitives.prepare import prepare
+from qrisp.jasp import jrange, RUS
+from qrisp.operators import QubitOperator
+
+
+def CKS_parameters(A, eps, kappa=None, max_beta=None):
+    """
```
positr0nium, line 40: Reference would also be nice here

src/qrisp/algorithms/cks.py
```diff
@@ -0,0 +1,821 @@
...
+    this approach enables scalable gate count estimation for linear systems of arbitrary size.
+    
+    **Example 2: Using a custom block encoding**
+
+    The previous example displays how to solve the linear system for any Hermitian matrix :math:`A`,
+    where the block-encoding is constructed from the Pauli decomposition of :math:`A`. While this approach is efficient when :math:`A`
+    corresponds to, e.g., an Ising or a Heisenberg Hamiltonian, the number of Pauli terms may not scale favorably in general.
+    For certain sparse matrices, their structure can be exploited to construct more efficient block-encodings.
+
+    In this example, we construct a block-encoding representation for a tridiagonal 3-sparse matrix defined by
```
positr0nium: It would be helpful to have an example matrix - seeing the code is also good but an example might help even more.

src/qrisp/operators/qubit/qubit_operator.py
renezander90: This is a good point. I will change this here. However, note that we also do not follow this convention for the qswitch, and changing this in the future will be a breaking change.

src/qrisp/algorithms/cks.py
positr0nium: You can pass numpy arrays by using the static feature of RUS and converting the array to a tuple before:
```
from qrisp import *
import numpy as np

@RUS(static_argnums = 0)
def inner_rus(array):
    
    qv = QuantumVariable(2)
    h(qv[array[0]])
    
    return measure(qv) == 0, qv

@jaspify
def main():
    array = np.ones(2)
    array_tuple = tuple(array)
    res_qv = inner_rus(array_tuple)
    return measure(res_qv)

print(main())
```
This snippet needs the latest commit to be bug-free 😇

src/qrisp/operators/qubit/qubit_operator.py
renezander90: Done in 6f9ce42ba3ca7883e4979922f7b769e82980c9cf.

src/qrisp/operators/qubit/qubit_operator.py
renezander90: Done in b91ed663d7f34aed1f29ccaa2edd8b0187f97d99.

src/qrisp/operators/qubit/qubit_operator.py
renezander90: Done in b91ed663d7f34aed1f29ccaa2edd8b0187f97d99. I think it is true, if |G> = G|0>, i.e., the state preparation acts on |0>.

src/qrisp/operators/qubit/qubit_operator.py
renezander90: Done in b91ed663d7f34aed1f29ccaa2edd8b0187f97d99.

src/qrisp/algorithms/cks.py
renezander90: Yes, but for 2D arrays it's more complicated. Then `tuple(arr)`will be a tuple of arrays.
Since `A` can also be a block encoding and `b` a function, I would keep this as it is. Otherwise, it would require further case distinctions.

src/qrisp/operators/qubit/qubit_operator.py
renezander90: Signature of block encoding was changed to `U(case, operand)`in 8c92a363d0ad66a7b1e913dc4a4521ba5896dc96

src/qrisp/algorithms/cks.py
MatP1337: Done in [295fb4b](https://github.com/eclipse-qrisp/Qrisp/pull/294/commits/295fb4b8f6847530171880b76018be45e007792b)

src/qrisp/algorithms/cks.py
MatP1337: Done in [295fb4b](https://github.com/eclipse-qrisp/Qrisp/pull/294/commits/295fb4b8f6847530171880b76018be45e007792b)

src/qrisp/operators/qubit/qubit_operator.py
MatP1337: This was done by @renezander90 in [b91ed66](https://github.com/eclipse-qrisp/Qrisp/pull/294/commits/b91ed663d7f34aed1f29ccaa2edd8b0187f97d99)

src/qrisp/algorithms/cks.py
MatP1337: Done in [c99afb9](https://github.com/eclipse-qrisp/Qrisp/pull/294/commits/c99afb91930483a3df434050f3518496a76bb8ac)

PR #305, opened 2025-11-07
Add `mcz`, `mcp` and `crz` to `QuantumCircuit`

purva-thakre:
The Qrisp documentation provides [a list of operations](https://qrisp.eu/reference/Primitives/Gate%20functions.html) that can be added to an object of the `QuantumCircuit` class. 

If the `mcz`, `mcp` or `crz` gates are added to a quantum circuit, we get `AttributeError: 'QuantumCircuit' object has no attribute 'mcz'/'mcp'/'crz'`. This PR adds the missing gates as methods of the `QuantumCircuit`class. 

```py

from qrisp import QuantumCircuit
qc = QuantumCircuit(4)
qc.cx(0, range(1,4))
qc.mcz([0, 1], 3)
print(qc)

```
In the above code block, replace `mcz` with `mcp` or `crz`. 

~~Edit: Spoke with Raphael on this. The gates that can be applied to the `QuantumCircuit` class are different from the gate functions on the linked page in the documentation. I don't know if it's worth adding these gates....~~

~~To do: mcrx exists in `standard_operations.py` but not as a method of the `QuantumCircuit` class.~~ No need as crx is a version of this gate but under a different name.


positr0nium:
Thanks for tackling this! To align the implementation of ``QuantumCircuit.mcp`` and ``QuantumCircuit.mcz`` more with the standalone functions ``mcp`` and ``mcz``, I think it would be beneficial to remove the ``target`` argument 👍

PR #306, opened 2025-11-10
Dynamic State Preparation

PietropaoloFrisoni:
**Context**: This PR implements the state preparation algorithm based on the technique using multiplexers. However, instead of the latter, we use the `qswitch` function.

**Description of the Change**: As above. For a description of the algorithm, we refer to the documentation.

**Benefits**: Qrisp can now initialize a quantum circuit without resorting to `qiskit`.

**Possible Drawbacks**: None that I can think of, although we can probably implement a more efficient algorithm in the future. The reason is that the `qswitch` function introduces several auxiliary qubits, which can easily cause overhead.

**Related GitHub Issues**: None

### DIFF
--- a/documentation/source/reference/Primitives/generated/qrisp.init_state_qswitch.rst
+++ b/documentation/source/reference/Primitives/generated/qrisp.init_state_qswitch.rst
@@ -1,6 +1,0 @@
-﻿qrisp.init\_state\_qswitch
-==========================
-
-.. currentmodule:: qrisp
-
-.. autofunction:: prepare_qswitch
--- a/documentation/source/reference/Primitives/generated/qrisp.prepare.rst
+++ b/documentation/source/reference/Primitives/generated/qrisp.prepare.rst
@@ -1,6 +1,0 @@
-﻿qrisp.prepare
-=============
-
-.. currentmodule:: qrisp
-
-.. autofunction:: prepare
--- a/documentation/source/reference/Primitives/generated/qrisp.prepare_qswitch.rst
+++ b/documentation/source/reference/Primitives/generated/qrisp.prepare_qswitch.rst
@@ -1,6 +1,0 @@
-﻿qrisp.prepare\_qswitch
-======================
-
-.. currentmodule:: qrisp
-
-.. autofunction:: prepare_qswitch
--- a/documentation/source/reference/Primitives/prepare.rst
+++ b/documentation/source/reference/Primitives/prepare.rst
@@ -5,8 +5,4 @@ Quantum State Preparation
 
 .. currentmodule:: qrisp
 
-.. autosummary::
-   :toctree: generated/
-
-   prepare_qswitch
-   prepare
+.. autofunction:: prepare


src/qrisp/alg_primitives/state_preparation.py
```diff
@@ -0,0 +1,335 @@
...
+
+    # These imports are here to avoid circular dependencies
+    from qrisp import gphase, qswitch, ry, u3
+    from qrisp.misc.utility import bit_reverse
+
+    target_array = jnp.asarray(target_array, dtype=jnp.complex128)
+    # n is static (known at compile time), so we can use normal numpy here
+    n = int(np.log2(target_array.shape[0]))
+
+    thetas, u_params, phases = _preprocess(target_array)
```
PietropaoloFrisoni, line 297: This classical preprocessing function for `n=20` takes 0.3 seconds on my laptop (each layer is vectorized with `vmap`). Therefore, most of the time is spent in the quantum part, which basically just calls `qswitch` (with a bit reverse operation, as we cannot slice a `DynamicQubitArray` backward)

src/qrisp/core/quantum_variable.py
```diff
@@ -718,6 +718,49 @@ def encode(self, value, permit_dirtyness=False):
 
         int_encoder(self, self.encoder(value))
 
+    def init_state_qswitch(self, state_array, method="auto"):
```
PietropaoloFrisoni: @positr0nium Instead of having a separate `init_state_qswitch` method, what about modifying the existing `init_state` in such a way that the workflow is dispatched with different implementations depending on the type of the provided argument? 

That is, if a `dict` is provided, the `qiskit` implementation is used, which is currently based on a different (and currently more efficient) algorithm, while if an array is provided we dispatch to the implementation introduced in this PR. Please let me know what you think : )

src/qrisp/core/quantum_variable.py
```diff
@@ -718,6 +718,49 @@ def encode(self, value, permit_dirtyness=False):
...
+        if len(state_array) != expected_length:
+            raise ValueError(
+                f"Length of statevector must be {expected_length} for {self.size} qubits, got {len(state_array)}."
+            )
+
+        norm = np.linalg.norm(state_array)
+        if np.isclose(norm, 0.0):
+            raise ValueError("The provided state vector has zero norm.")
+        if not np.isclose(norm, 1.0):
+            warnings.warn(
```
positr0nium: Not sure if having a warning here is necessary. It clutters the output and it is clear that the normalization happens. What do you think @renezander90 ?

src/qrisp/core/quantum_variable.py
positr0nium: Having two methods of QuantumVariable that do the same thing seems unintuitive for me. It would be better if ``QuantumVariable.init_state`` would be passed an optional keyword argument ``method`` that indicates, which method should be used.
This should also not depend on whether a dict or an array is passed.

renezander90: I agree, and we should explain in the documentation that only the qswitch-based method works for dynamic input, i.e., jax.numpy.ndarray.

Please also note that we have a (sightly redundant) additional `prepare` function:
https://qrisp.eu/reference/Primitives/prepare.html
https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/alg_primitives/prepare.py

It is there for two reasons: 1) init_state does raise an error if the variable is not in state |0> (is this necessary?), and 2) `prepare` is recognizable by users (e.g., LCU = PREP * SEL * PREP_dg) and more convenient to use than `init_state` in situations like:

```
from qrisp import *

qv = QuantumVariable(2)
b = np.array([1,1,1,1])

with conjugate(prepare)(qv,b):
    x(qv)
```

Hence, now we would have `prepare`(using Qiskit) and the new `state_preparation` (using qswitch). We should also consolidate this in one function via an optional keyword argument `method`. The latter is currently used to specify the `qswitch` method. Is this really required? 
Depending on whether we want to keep the name `prepare` or `state_preparation` we would also need to deprecate the former.

src/qrisp/core/quantum_variable.py
renezander90: Yes, I would just always rescale automatically.

src/qrisp/core/quantum_variable.py
PietropaoloFrisoni: Hi @renezander90 , nice to meet you : )

I briefly returned to this today (I am currently working on other projects). At the moment, I am not convinced that we should completely replace the `qiskit`-base preparation method for this one, as the former is much faster. It is based on [a more efficient algorithm with fewer gates](https://arxiv.org/abs/1501.06911) and it already uses some routines written in lower-level languages from `qiskit._accelerate` (so it's very difficult to do better than that).

It seems to me that the main motivation for this is to have a dynamic (jasp-compatible) preparation algorithm in `qrisp` (which justifies the usage of `qswitch`, which is already jasp-compatible), so potentially we should rely on this method only in tracing mode, and keep using the other implementation in static mode. I agree with you that this should not depend on the provided data structure for the parameters. What do you think?

renezander90: Nice to meet you too @PietropaoloFrisoni :)

Yes, maybe my reply was a bit confusing. I was not suggesting to completely replace the efficient Qiskit state preparation, but to select the method via an optional keyword argument `method`. This could have an option `"auto"` , which uses Qiskit unless we are in tracing mode **and** the array is dynamic.

PietropaoloFrisoni: Having a `method=auto` argument that chooses the qiskit implementation in static mode and the `qswitch`-based implementation in tracing mode seems indeed a very sensible approach to me.

PietropaoloFrisoni: Done. 

The convention on the order of the amplitudes when providing an array is supposed to match the one retrieved from the `statevector` method (as I clarified in the docstrings). 

Edit: I updated the implementation so that `prepare` is automatically used in case the user wants to choose the old `qiskit` route. This avoids redundancies, and it is still compatible with the existing usage. I did not touch the logic when a `dict` is provided to `init_state` or when one uses `prepare` explicitly (it seems to me this is out of scope for this PR).

Honestly, I didn't even notice the existence of `prepare` when I started to work on this : )

src/qrisp/core/quantum_variable.py
PietropaoloFrisoni: I removed the warning : )

PR #307, opened 2025-11-12
Qrisp to Cirq converter

purva-thakre:


src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -0,0 +1,69 @@
...
+    for i, q in enumerate(qrisp_circuit.qubits):
+        qubit_map[q] = cirq_qubits[i]
+
+    for instr in qrisp_circ_ops_data:
+        
+        # get the gate name, qubits it is acting on 
+        # and parameters if there are any
+        op_i = instr.op.name
+        if op_i == 'barrier':
+            raise ValueError("Qrisp circuit contains a barrier which is unavailable in Cirq.")
```
positr0nium: Maybe we just ignore the barrier? It doesn't change anything about the unitary. Raising the exception here could make it annyoing if the user's circuit is generated by some external code.

ACE07-Sev: Can be ignored for now, I'm actually working with cirq team to add barrier, shouldn't take too long for it to be added.

Maybe just add a TODO note for it.

purva-thakre: > Maybe we just ignore the barrier? It doesn't change anything about the unitary. Raising the exception here could make it annyoing if the user's circuit is generated by some external code

yes, I know it doesn't change anything about the unitary. but in case they were expecting a similar visual barrier in the cirq circuit, I thought it would be better to raise an error. 

Now that I think of it, it might be better to raise an exception that proceeds with the circuit conversion rather than terminating after raising the error. 

Edit: Never mind. i think it's better to remove it as suggested.

purva-thakre:
> For global phase, what is the issue exactly?

Cirq's global phase gate's parameter is incompatible with the input parameter of Qrisp's global phase gate. For now, we are temporarily ignoring this gate in the conversion function until I figure out a generic mapping between the two.

ACE07-Sev:
Cirq takes `np.exp(1j*theta)` directly. You need to just pass that instead of pure theta. If you have an assertion for this, I can do it and give you what you'll need to put.

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -0,0 +1,204 @@
...
+            # added op_i != 'id' becasue the identity gate has parameters (0, 0, 0)
+            if op_i == "p":
+                # Cirq does not have a phase gate
+                # for this reason, it has to be dealt with as a special case.
+                # the ZPowGate has a global phase in addition to the
+                # phase exponent. The default is to assume global_shift = 0 in cirq
+                exp_param = params[0]
+                cirq_circuit.append(ZPowGate(exponent=exp_param)(*cirq_op_qubits))
+
+            elif op_i == "rxx":
```
positr0nium: I think the case for gates that have non-trivial ``.definition`` attribute is missing, or? This would already cover the cases of rxx, rzz etc.

tests/interface_tests/test_cirq_converter.py
```diff
@@ -0,0 +1,185 @@
...
+from qrisp.circuit import QuantumCircuit, ClControlledOperation
+
+from qrisp.interface.converter.cirq_converter import convert_to_cirq
+from qrisp import QuantumVariable, mcx, cx, QuantumBool
+from qrisp import h, x, reflection
+
+# define circuits to be used by the unit tests
+
+# 4 qubit circuit containing all single qubit gates
+qc_single_qubit_gates = QuantumCircuit(4)
```
positr0nium: I think it would be good to construct the circuits inside the testing functions. If there is a bug that crashes circuit construction, the whole testing pipeline crashes before a single test is executed.

setup.py
```diff
@@ -22,6 +22,7 @@
                 "numpy>=2.0",
                 "sympy<=1.13",
                 "qiskit>=0.44.0",
+                "cirq",
```
positr0nium: We should leave cirq as an optional dependency - maybe check how Pietro solved it with his PL converter :)

tests/interface_tests/test_cirq_converter.py
purva-thakre: > If there is a bug that crashes circuit construction, the whole testing pipeline crashes before a single test is executed.

I did not think of this. I like to write concise unit tests with the help of pytest's parametrize. This approach favors defining the circuits outside the unit test.

src/qrisp/interface/converter/cirq_converter.py
purva-thakre: > This would already cover the cases of rxx, rzz etc.

Thanks, I was not aware the `definition` attribute could be used in this manner.

tests/interface_tests/test_cirq_converter.py
```diff
@@ -0,0 +1,268 @@
...
+        p(y * 2 * np.pi, qv[1])
+
+    qv = QuantumVariable(2)
+
+    h(qv)
+
+    QPE(qv, U, precision=3)
+    test_circuit = qv.qs.compile()
+    # test_circuit has a composite gate QFT_dg
+    with pytest.raises(
```
positr0nium: I think it would be best, if this does not raise an error. Instead the converter should automatically decompose all composite gates that cause problems.

tests/interface_tests/test_cirq_converter.py
```diff
@@ -0,0 +1,268 @@
...
+    qc_single_qubit_gates.ry(0.4, 1)
+    qc_single_qubit_gates.rz(0.2, 2)
+    qc_single_qubit_gates.s(0)
+    qc_single_qubit_gates.t(1)
+    qc_single_qubit_gates.t_dg(3)
+    qc_single_qubit_gates.s_dg(2)
+    qc_single_qubit_gates.measure(0)
+    qc_single_qubit_gates.reset(0)
+
+    expected_cirq_qc_single_qubit_gates_ops = [
```
positr0nium: I think it would be better if you test for equivalence by computing the unitary. Testing the "hardcoded" gate sequence might cause this test to fail if we change the compiler in a way that modifies the order (it would for instance be perfectly legal to exchange x(1) and y(3).

tests/interface_tests/test_cirq_converter.py
```diff
@@ -0,0 +1,268 @@
...
+        H(LineQubit(3)),
+        CNOT(LineQubit(4), LineQubit(2)),
+        (Z**-0.7853981633974483).on(LineQubit(3)),
+        (Z**0.39269908169872414).on(LineQubit(2)),
+        CNOT(LineQubit(4), LineQubit(2)),
+        CNOT(LineQubit(4), LineQubit(3)),
+        (Z**0.7853981633974483).on(LineQubit(3)),
+        CNOT(LineQubit(4), LineQubit(3)),
+        H(LineQubit(4)),
+    ]
```
positr0nium: Some more tests with slightly more advanced algorithms could be an easy way to make this more solid (for instance the Grover example from the 101 tutorial).

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -0,0 +1,196 @@
...
+        if hasattr(instr.op, "params"):
+            params = instr.op.params
+
+        if op_i not in qrisp_cirq_ops_dict:
+            raise ValueError(
+                f"{op_i} gate is not supported by the Qrisp to Cirq converter."
+            )
+
+        if op_i in ["gphase"]:
+            print(
```
positr0nium, line 105: What exactly is the issue here? Why doesn't the global phase gate work? I know we discussed this but I still didn't fully understand.

tests/interface_tests/test_cirq_converter.py
purva-thakre: > Testing the "hardcoded" gate sequence might cause this test to fail if we change the compiler in a way that modifies the order

This is a good point.

tests/interface_tests/test_cirq_converter.py
purva-thakre: > (for instance the Grover example from the 101 tutorial

Sorry, do you mind clarifying where the 101 tutorial is?

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -0,0 +1,197 @@
...
+            "reset",
+            "id",
+            "p",
+            "sx",
+            "sx_dg",
+        ]
+
+        if (op_i not in cirq_gates_filter) and instr.op.definition:
+            new_circ = instr.op.definition
+            cirq_circuit.append(convert_to_cirq(new_circ))
```
positr0nium: I think there is still an issue here. I'd expect that you somehow have to specify the relevant cirq qubits. This bugged example confirms my suspicion:
```
from qrisp import *

qc = QuantumCircuit(4)
qc.rxx(0.3, 0, 1)
qc.rxx(0.3, 2, 3)
print(qc.to_cirq())
# 0: ───H───@─────────────@───H───H───@─────────────@───H───
#           │             │           │             │
# 1: ───H───X───Z^0.095───X───H───H───X───Z^0.095───X───H───
```

positr0nium (approved):
Nice work :)

PR #315, opened 2025-11-13
Lanczos method

MatP1337:
This pull request add a complete implementation of [Exact and efficient Lanczos method on a quantum computer](https://arxiv.org/pdf/2208.00567) for ground state energy estimation.

The quantum Lanczos method efficiently constructs a Krylov subspace by applying Chebyshev polynomials of the Hamiltonian to an initial state. It avoids the classical barrier of exponential cost in representing Krylov vectors.

Implements the following steps:
      1. ``inner_lanczos``: construct the circuit
      2. ``lanczos_expvals``: runs quantum Lanczos subroutine to obtain Chebyshev expectation values $\langle T_k(H)\rangle$.
      3. ``build_S_H_from_Tk``: builds overlap and Hamiltonian subspace matrices $(\mathbf{S}, \mathbf{H})$.
      4. ``regularize_S_H``: regularizes overlap matrix $\mathbf{S}$ and $\mathbf{H}$ by projecting onto the subspace with well conditioned eigenvalues.
      5. ``lanczos_alg``: utilizes all of the above to solve the generalized eigenvalue problem $\mathbf{H}\vec{v}=\epsilon\mathbf{S}\vec{v}$. Returns lowest eigenvalue $\epsilon_{\text{min}}$ as ground state energy estimate.

##### To-do:
- [x] make compatible with jasp
     - [x] https://github.com/eclipse-qrisp/Qrisp/pull/314 fixes an occurring bug
- [x] update documentation to functions
- [x] add test(s)

### DIFF
--- a/.github/workflows/qrisp_test.yml
+++ b/.github/workflows/qrisp_test.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 90
     strategy:
       matrix:
         python-version: ["3.11.*"]


renezander90:
Auxiliary methods (inner_lanczos, lanczos_expvals) are Jasp compatible and jax versions of classical post processing functions are implemented. However, solving the generalized eigenvalue problems seems to be not implemented for the corresponding jax.numpy/jax.scipy functions (at least for version 0.6). Hence, executing full Lanczos method in Jasp is currently not supported..

src/qrisp/algorithms/lanczos.py
```diff
@@ -0,0 +1,477 @@
...
+    polynomials $T_k(H)$, closely following the algorithm proposed in
+    `"Exact and efficient Lanczos method on a quantum computer" <https://quantum-journal.org/papers/q-2023-05-23-1018/>`_.
+
+    The quantum Lanczos algorithm efficiently constructs a Krylov subspace by applying Chebyshev polynomials
+    of the Hamiltonian to an initial state. The Krylov space dimension $D$ determines the accuracy of ground
+    state energy estimation, with convergence guaranteed when the initial state has a sufficiently large overlap 
+    ($|\gamma_0|=\Omega(1/\text{poly}(n))$ for $n$ qubits) with the true ground state. The Chebyshev approach allows exact
+    Krylov space construction (up to sample noise) without real or imaginary time evolution.
+
+    This algorithm is motivated by the rapid convergence of the Lanczos method for estimating extremal eigenvalues,
```
renezander90: We should also mention the numerical instability of the generalized eigenvalue problem, and the importance of regularization (a wrong result for the energy could be returned depending on the cutoff).

src/qrisp/algorithms/lanczos.py
```diff
@@ -0,0 +1,477 @@
...
+    return S_reg, H_reg
+
+
+def lanczos_alg(H, D, operand_prep, mes_kwargs={}, cutoff=1e-2, show_info=False):
+    r"""
+    Exact and efficient Lanczos method on a quantum computer for ground state energy estimation.
+
+    This function implements the Lanczos method on a quantum computer using block-encodings of Chebyshev
+    polynomials $T_k(H)$, closely following the algorithm proposed in
+    `"Exact and efficient Lanczos method on a quantum computer" <https://quantum-journal.org/papers/q-2023-05-23-1018/>`_.
```
renezander90: We should could add some formulas for derivation of the generalized eigenvalue problem form expectation values <T_k(H)>.

src/qrisp/algorithms/lanczos.py
MatP1337: Done in https://github.com/eclipse-qrisp/Qrisp/pull/315/commits/bdfe76b09d800563fbaf0ad35571ea19852eb1a4

image

src/qrisp/algorithms/lanczos.py
MatP1337: Done in https://github.com/eclipse-qrisp/Qrisp/pull/315/commits/ff60c9118d60824bf64b66c400bc880afff65439

MatP1337:
Perhaps it would also be useful to write a sentence or two about the intuition of the Chebyshev polynomials acting as a filter to amplify the true ground state and/or suppressing the excited (non ground) states.

PR #321, opened 2025-11-19
Minor fixes in QPE and Reflection examples in docs

purva-thakre:
https://www.qrisp.eu/reference/Primitives/QPE.html
https://www.qrisp.eu/reference/Primitives/reflection.html


src/qrisp/alg_primitives/reflection.py
```diff
@@ -81,7 +81,7 @@ def ghz(qv):
             h(qv[0])
 
             for i in range(1, qv.size):
-            cx(qv[0], qv[i])
+                cx(qv[0], qv[i])
```
purva-thakre, line 84: this fixes below:

Image

PR #333, opened 2025-12-02
Fix: Rutorial links in README

JoKaBus:
# Fix Tutorial Links

Tutorial links now point to the Shor tutorial and the index of the tutorial. Before, we had

image


positr0nium:
Cheers :)

Issue #336, opened 2025-12-02
Overload `__eq__` for `QuantumArray`


cjn10:
Currently, the equality operator `__eq__` is not overloaded for `QuantumArray`. This leads to the following behavior since Python falls back to the basic implementation of the operator.

```
from qrisp import *

a_array = QuantumArray(QuantumFloat(4), shape=(5,))
b_array = QuantumArray(QuantumFloat(4), shape=(5,))
r_array = QuantumArray(QuantumBool(), shape=(5,))

r_array = (a_array == b_array)
print(r_array)
>>> False
print(a_array)
>>> {OutcomeArray([0, 0, 0, 0, 0]): 1.0}   
print(b_array)
>>> {OutcomeArray([0, 0, 0, 0, 0]): 1.0}   
```

Implementing the equality operator element-wise should be straightforward.

cjn10:
Similarly, other operators should be implemented element-wise, e.g., `__add__` and `__mul__`.

renezander90:
Also comparisons should work element-wise. E.g., like
```
a = np.array([1,2,3])
b = np.array([1,2,4])

a==b
```
> array([ True,  True, False])

Feel free to create a PR :)

PR #344, opened 2025-12-16
Add injection and element-wise arithmetic for quantum array

cjn10:
Implements various operators for the QuantumArray, see #336.

A few remarks are in order:
- The question of the return type of the non-jasp QuantumFloat addition should be revisited. As of now, it appears that a QuantumFloat of expanded size is returned. This creates problems with injection operator and, possibly, introduces overhead that is not needed by the user. In the jasp-version, the return type is the same as the input type.
- Care should be taken when using `a in a_list` to check if an element `a` is in a list `a`. Internally, Python uses the `__eq__` operator to compare elements. However, for some quantum variables this operation is overloaded and creates a circuit and new quantum variables.
- In same cases, if clauses are used even in jasp-compatible code. This works when the expression is static. However, in some use caess, a static variable is used as a dynamic variable, which results in an error when an if clause is used. See for example, `QuantumFloat.sign()`.


positr0nium:
```python
from qrisp import *

@jaspify
def test():
    a_array = QuantumArray(QuantumFloat(4), shape=(4,4))
    x(a_array)
    b_array = QuantumArray(QuantumFloat(4), shape=(4,4))
    h(b_array)

    # This fails
    # r_array = QuantumArray(qtype = addition_result_type, shape=(4,4))
    
    # This doesn't
    addition_result_type = create_output_qf([a_array.qtype, b_array.qtype], op = "add")
    r_array = QuantumArray(qtype = addition_result_type, shape=(4,4))

    (r_array << (lambda a,b: a+b))(a_array, b_array)
    return measure(r_array), measure(a_array), measure(b_array)

r, a, b = test()
```

The problem is that the qtype of ``r_array`` is too small to hold the result. An addition between two size 4 ``QuantumFloat`` can (worst case) result in 30. This needs at least 5 qubits to prevent overflow.
In principle we can change how out-of-place arithmetic determines output size but it is not really clear to me what the best behavior. Here are the options that I see most viable:

1. No overflow ever (current version). The size will be chosen such that even the worst case is always covered. The disadvantage is that programs, which use a lot of out-of-place arithmetic constantly grow the QuantumFloat size.
2. ``new_size = max(a.size, b.size)``. Potentially another option but here overflow is possible. Compared to classical overflow the issue is even worse because there can be no error-message that the overflow happened.

It should be said that the bug that currently makes the pipeline crash can still happen if the user chooses the target too small - regardless of what behavior we implement.

What do you think is best? Also tagging @purva-thakre since she has become interested in arithmetic topics.

PR #352, opened 2026-01-01
Refactor qswitch interface

renezander90:
**Context:**

The current `qswitch` interface is `qswitch(operand, case, case_function, method="auto", case_amount=None)`.
This has the following disadvantages: 
1) The `operand` argument precedes the `case` argument, i.e., the controlled varaible precedes the control variable, which is contrary to how this is handled in other qrisp methods, e.g., `mcx(ctrl, target)`. 
2) Only one operand variable can be provided (this could be circumvented by passing a list, but is rather inconvenient because the different branches have to be defined in terms of acting on lists of variables). 
3) It does not match the jax.lax.switch (https://docs.jax.dev/en/latest/_autosummary/jax.lax.switch.html) api.

**Description of change**
Changed interface to`qswitch(case, branches, *operands, method="auto", branch_amount=None)` matching the jax.lax.switch api.

**Benefits:** Resolves issues 1), 2), 3). Mainly ensures consistency of interfaces within qrisp.
Additionally, the `qswitch` could also allow the `case` argument to be an integer, and in this case act as qrisp version of the jax.lax.switch (similar as `q_fori_loop`, `q_while_loop`).

**Possible Drawbacks:** 
While the old interface has been deprecated, **it still works** (by proper mapping of inputs) to avoid an immediate breaking change and give the user time to adapt, and a warning is thrown. 
We cannot exclude at 100% edge-case incompatibilities. For example, when non-keyword arguments have been used as if they were keyword arguments (e.g., `qswitch(operand=qv,...)`). 

**Related GitHub Issues:** None

### DIFF
--- a/src/qrisp/alg_primitives/switch_case.py
+++ b/src/qrisp/alg_primitives/switch_case.py
@@ -45,7 +45,7 @@ def _invert_inpl_function(func):
 
 
 def qswitch(
-    operand, case, case_function, method="auto", case_amount=None, inv=False, ctrl=None
+    case, branches, *operands, method="auto", branch_amount=None, case_amount=None, inv=False, ctrl=None
 ):
     r"""
     Executes a switch - case statement distinguishing between a list of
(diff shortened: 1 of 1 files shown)


renezander90:
Part of https://github.com/eclipse-qrisp/Qrisp/pull/374

PR #358, opened 2026-01-05
Dynamic state preparation follow-up

PietropaoloFrisoni:
**Context**:  The endianness situation across `QuantumVariable.init_state`, `prepare`, and`QuantumVariable.statevector_array` is not clear in my opinion. For example, we can specify `reversed` in the state preparation algorithm, but reversed respect to what?  This became even more confusing after this [PR](https://github.com/eclipse-qrisp/Qrisp/pull/306) has introduced state preparation based on an explicit statevector array. 

This PR changes the behaviour of `prepare_qswitch` and tries to clarify the situation across different docstrings (see below for details).

**Description of the Change**: 

- Updated the docstring of `QuantumVariable.init_state` with details, clarifications, and an explicit example.

- Updated the docstring of `prepare` with details, clarifications, and an explicit example.

- Updated the description of `QuantumVariable.statevector_array` so that it is very explicit about the endianness and index-qubit mapping of the returned statevector.

- Changed the behaviour, signature and docstrings of `prepare_qswitch`: after this PR it uses **little-endian** convention by default, instead of the opposite convention (as it did before this PR). We achieved this by reversing the endianness of the input array (the algorithm is exactly the same), which also enables reversed state preparation with this method. It is not a breaking change as this function is not public yet (the [PR](https://github.com/eclipse-qrisp/Qrisp/pull/306) implementing it was merged after the last release).

**Benefits**:  Hopefully users will be less confused about endianness situation.

**Possible Drawbacks:** None that I can think of, as dynamic state preparation with `qswitch` is not part of stable `qrisp` yet, so it cannot be considered a breaking change.

### DIFF
--- a/src/qrisp/alg_primitives/state_preparation/prepare_func.py
+++ b/src/qrisp/alg_primitives/state_preparation/prepare_func.py
@@ -39,7 +39,7 @@ def prepare(qv, target_array, reversed: bool = False, method: str = "auto"):
 
         \sum_{i=0}^{N-1} b_i \ket{i}.
 
-    The ``target_array`` is normalized internally, i.e.
+    The ``target_array`` input vector is normalized internally, i.e.
 
     .. math::
 


src/qrisp/alg_primitives/state_preparation/prepare_func.py
```diff
@@ -17,91 +17,149 @@
...
+    Prepare a quantum state on ``qv`` from a target amplitude vector.
+
+    Given a vector :math:`b=(b_0,\dotsc,b_{N-1})` (corresponding to ``target_array``),
+    this routine prepares a state proportional to
+
+    .. math::
+
+        \sum_{i=0}^{N-1} b_i \ket{i}.
+
+    If ``target_array`` is a concrete array (non-Jasp mode), it is normalized internally, i.e.
```
renezander90: Why no normalization in Jasp mode?

src/qrisp/alg_primitives/state_preparation/qiskit_state_preparation.py
```diff
@@ -19,12 +19,15 @@
...
+
 def prepare_qiskit(qv, target_array, reversed=False):
-    
+
     try:
         target_array = np.array(target_array)
     except TracerArrayConversionError:
-        raise ValueError("Tried to initialize dynamic jax array using state preparation method qiskit")
+        raise ValueError(
+            "Tried to initialize dynamic jax array using state preparation method qiskit"
```
renezander90: "Tried to initialize **variable from** dynamic jax array using state preparation method qiskit" ?

tests/core_tests/test_state_preparation.py
```diff
@@ -118,6 +126,24 @@ def test_error_unrecognized_method(self):
...
+    def test_prepare_qswitch_with_endianness(self, big_endianness):
+        """Test state preparation via `prepare_qswitch` with different endianness settings."""
+
+        qv = QuantumVariable(3)
+        array = jnp.array([1, 1j, -1, -1j, 1, 1j, -1, -1j])
+        array /= jnp.linalg.norm(array)
+
+        prepare_qswitch(qv, array, big_endianness=big_endianness)
+
+        logical_sv = _compute_statevector_logical_qubits(
```
renezander90, line 138: Could we expose the big_endianness keyword also to the statevector_arrray method? (With False as default value.)

PietropaoloFrisoni: If you refer to the `statevector_array` method of `QuantumCircuit`, I would *really* like to change this convention (it is one of the few remaining places with opposite endianness convention), and possibly to expose a `reversed`/`big_endianness` parameter, as you just suggested.

 Unfortunately, it is not totally straightforward and it would make the computation of unitaries slightly more inefficient (or so I understood from discussions with Raphael). But I am not sure as I haven't looked into the details of the code yet. Definitely, changing this convention is on our radar (and Raphael agrees that should be changed).
 
 Still, at least the update(s) in this PR for such a method remove the outdated reference to `qiskit` and make this convention way more vocal for users

src/qrisp/alg_primitives/state_preparation/prepare_func.py
PietropaoloFrisoni: In `jasp` mode everything is harder to check as it is potentially a Tracer (including the size of the `QuantumVariable`): whether the input statevector has zero norm (in that case we cannot renormalize it), whether it has the expected size, and so on. So back in the days when I implemented this algorithm I removed all checks in `jasp` mode not to have a proliferation of `jax.lax.cond`, `jnp.where`, etc. along with the corresponding structure in concrete `NumPy`. It should not be a conceptual block, it is just that my focus was more on the algorithm implementation.

src/qrisp/alg_primitives/state_preparation/qiskit_state_preparation.py
PietropaoloFrisoni: I improved the error message. Thanks!

tests/core_tests/test_state_preparation.py
renezander90: One thing that puzzled me:

```
import numpy as np
from qrisp import *
from qrisp.operators import QubitOperator

# This is X \otimes Z
A = np.array([[ 0.+0.j,  0.+0.j,  1.+0.j,  0.+0.j],
       [ 0.+0.j,  0.+0.j,  0.+0.j, -1.+0.j],
       [ 1.+0.j,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j, -1.+0.j,  0.+0.j,  0.+0.j]])

b = np.array([0, 1, 0, 0]) # |1>
print(A @ b)
# [ 0.+0.j  0.+0.j  0.+0.j -1.+0.j]
# A acts as:
# |0> -> |2>; |1> -> -|3>; |2> -> |0>; 3 -> -|1>

# We added reverse_endianness to get a Pauli operator that is consistent with A @ b
O = QubitOperator.from_matrix(A, reverse_endianness=True).to_pauli()
print(O)
#1.0*Z(0)*X(1) This is inconsistent with X \otimes Z above (?)
print(O.to_array())
#[[ 0.+0.j  1.+0.j  0.+0.j  0.+0.j]
# [ 1.+0.j  0.+0.j  0.+0.j  0.+0.j]
# [ 0.+0.j  0.+0.j  0.+0.j -1.+0.j]
# [ 0.+0.j  0.+0.j -1.+0.j  0.+0.j]]
# This is Z \otimes X 

qf = QuantumFloat(2)
qf[:] = 1
# Apply operator O to QuantumFloat in state |1>
z(qf[0])
x(qf[1])
print(qf)
# {3: 1.0} This is consistent with A @ b
```

Hence, if we would change the endianness in ``to_array`` and ``from_matrix`` such that (by default) Z(0)X(1) <-> X \otimes Z (**big-endian**), then the QubitOperator would always be consistent with our QuantumFloat (**little-endian**) in the sense that:

A @ b = QubitOperator.from_matrix(A) acting on QuantumFloat in state |b>

I am was a bit confused about this big-endian vs little-endian situation, but the tensor product is somehow big-endian.

To summarize, the following **changes for QubitOperator** should be made (in a separate PR):
- Change ``from_matrix`` such that (by default): Z(0)X(1) <- X \otimes Z
- Change ``to_array`` such that (by default): Z(0)X(1) -> X \otimes Z

(If we additionally change the endianness of the ``statevector_array`` method, everything in Qrisp should be consistent and we wouldn't need any reversing of the endianness which we only introduced over time to deal with inconsistencies.) 

Would you agree?

src/qrisp/alg_primitives/state_preparation/prepare_func.py
renezander90: Yes, in general, we should keep the code simple and avoid excessive use of ``jax.lax.cond`` etc.
But, regarding the normalization: if the user provides an array of norm 0, wouldn't the computation be useless anyway? Thus, we could always renormalize.

```
@terminal_sampling
def main():

    qv = QuantumFloat(2)
    arr = jnp.array([0.,0.,0.,0.])
    prepare(qv, arr, reversed=True, method="qswitch")
    return qv

main()
# {0.0: nan, 1.0: nan, 2.0: nan, 3.0: nan}
```

I guess some angles for rotations get the value nan in this case? 

FYI: based on AI suggestions, I tried this:

```
import jax
import jax.numpy as jnp

def throw_error(val):
    raise ValueError(f"Invalid dynamic value: {val}")

@jax.jit
def compute(x):
    # Check if x is zero
    is_zero = jnp.all(x == 0)
    
    # If is_zero is True, execute the callback
    jax.lax.cond(
        is_zero,
        lambda _: jax.debug.callback(throw_error, x),
        lambda _: None,
        operand=None
    )
    return x + 1

# compute(jnp.array(0))  <-- This will raise the ValueError
```

But, inside jaspify it does not work:
```
@jaspify
def main():
    compute(jnp.array(0))
    return 0

main()
# array(0)
```

tests/core_tests/test_state_preparation.py
PietropaoloFrisoni: Thanks for the investigation! This will certainly save us time.

> Would you agree?

Yes, absolutely. I can look into it in another PR (since this is not related only to state preparation, but it also affects other parts of the code)

renezander90: Thank you. This would be appreciated. Yes, it would also help us to avoid some reversing of endianness and swapping qubits in the HHL tutorial https://qrisp.eu/general/tutorial/HHL.html, and in some other examples.

src/qrisp/alg_primitives/state_preparation/prepare_func.py
PietropaoloFrisoni: I was confused until I realized that now we are always renormalizing the statevector before the agorithm. This change was introduced in [this commit](https://github.com/eclipse-qrisp/Qrisp/commit/c05dc1f644d1d49534f30ccbe72699329bfddd7e). 

Therefore, the docstring is currently inconsistent with what the source code is doing.

I kept the current behaviour (always renormalize for every vector) and updated the docstring accordingly.

renezander90: I also forgot about that I made this change. I did this because I was used to provide non-normalized arrays to the ``prepare``function.

However, if we comment out this line:

```
import jax.numpy as jnp
from qrisp import *

@terminal_sampling
def main():

    qv = QuantumFloat(2)
    arr = jnp.array([0.,0.,0.,0.])
    prepare(qv, arr, reversed=True, method="qswitch")
    return qv

main()
# {3.0: 1.0} 
```

Hence, we avoid program errors caused by nan angles. It's unclear to me what would happen when attempting to run ``ry(nan, qb)`` on real hardware. Thus, I would now be in favor of either using ``jnp.where`` or not normalizing (I would still prefer the first option).

PietropaoloFrisoni: I am fine with pretty much every approach as long as it is consistent and properly documented : ). 

Feel free to push other commits to this PR if you want to change the normalization behaviour (and, if there are no other questions, feel free to merge it when you are done). 

This is not related to endianness, which was the reason why I felt the urgency to create this PR. I hope I'll find the time to fix `QubitOperator` as well before the next release 👍

PR #360, opened 2026-01-08
Computing the depth of a quantum circuit

PietropaoloFrisoni:
**Context**:  Before this PR, the `jasp` profiler was tightly coupled to a single metric: gate counting.
This made it difficult to extend the profiler to compute additional metrics on a `jaspr`.

This PR generalizes the profiler architecture so that multiple metrics can be computed in an extensible way, and introduces the first new metric built on top of this structure: circuit depth.

Implementing depth computation within the previous profiler structure would have resulted in significant duplication and further entanglement of logic. A minimal architectural separation was therefore introduced in this PR to make the profiler scalable to multiple metrics while keeping changes localized and backward-compatible.

This structure can be certainly improved and further refined in future work, but it hopefully provides a clean foundation for metric extensibility at the current stage.

Finally, a follow-up PR (with another metric + a few improvements to the architecture) is [currently open](https://github.com/PietropaoloFrisoni/Qrisp/pull/4)

**Description of the Change**: 

This PR has two main components:

- Profiler refactor

The profiler logic has been refactored to separate the profiling interpreter (common logic) and the metric implementation (metric-specific behavior). This is achieved by introducing a common `BaseMetric` abstraction and moving the previous `count_ops` logic into its own metric implementation.
The new structure allows additional metrics to be added without modifying the profiling interpreter.

- Depth metric implementation

Using the new structure, this PR implements a `DepthMetric` that computes the depth of the quantum circuit represented by a `jaspr`

**Benefits**:  The profiler can now compute multiple metrics in a structured way. The architecture is now extensible for future metrics. Finally, circuit depth can now be computed for any `jaspr`

**Possible Drawbacks:** The depth computation currently has several limitations.

- The simulation-based depth is not yet implemented
- Performance on very large/deep circuits has not yet been tested and/or optimized (follow-up work planned)

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
@@ -82,12 +82,18 @@ def _create_lookup_table(idx_start: ArrayLike, table_size: ArrayLike) -> ArrayLi
     return idx_start + jnp.arange(table_size, dtype=jnp.int64)
 
 
-# If this is triggered, the computation will fail anyway because of overflow.
-def _warn(_):
-    """Helper function to print a warning when the depth metric computation overflows the maximum number of qubits supported."""
+# The computation will fail anyway if this is triggered,
+# but this way we can print an informative message before the failure happens.
+def _warn(idx_end, max_qubits):
+    """Helper function to print a warning when the depth metric computation overflows."""
     jax.debug.print(
-        "ERROR: the depth metric computation overflowed the maximum number of qubits supported."
-        " Consider increasing the `max_qubits` parameter."
+        (
+            "ERROR: Depth metric computation overflowed: tried to create qubits with "
+            "global indices up to {idx_end}, but the maximum supported is {max_qubits}. "
+            "Consider increasing the `max_qubits` parameter for depth profiling."
+        ),
+        idx_end=idx_end,
+        max_qubits=max_qubits,
     )
 
 
@@ -153,7 +159,7 @@ class DepthMetric(BaseMetric):
 
         overflow = idx_end > jnp.int64(self.max_qubits)
         invalid = jnp.logical_or(invalid, overflow)
-        jax.lax.cond(invalid, _warn, lambda _: None, operand=None)
+        jax.lax.cond(invalid, _warn, lambda *_: None, idx_end, self.max_qubits)
 
         context_dic["_previous_qubit_array_end"] = idx_end
 


src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -107,9 +145,9 @@ def main(i):
             return measure(c)
 
         print(main(5))
-        # {'cx': 506, 'x': 22, 'h': 135, 'measure': 55, '2cx': 2, 's': 45, 't': 90, 't_dg': 90}
+        # {'s': 45, 'x': 22, 't_dg': 98, 'cx': 510, 't': 96, 'h': 139, 'measure': 55}
         print(main(5000))
-        # {'cx': 462552491, 'x': 20002, 'h': 112522500, 'measure': 37517500, '2cx': 2, 's': 37507500, 't': 75015000, 't_dg': 75015000}
+        # {'t': 751506, 'h': 1127254, 'x': 2002, 's': 375750, 't_dg': 751508, 'cx': 4629255, 'measure': 752500}
```
PietropaoloFrisoni, line 150: `count_ops` before was counting twice the measurement on 1-qubit gates. This has been fixed.

I am not sure about the remaining discrepancies. I tested this on the latest master branch, and I was getting the same result(s) as on this branch. I suspect there was something incorrect at the time this docstring was written. 

As far as I am concerned, the tests for `count_ops` are still passing, which should be the litmus test to verify that there are no errors with these types of changes/refactoring.

MatP1337:
amazing, cheers!

src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
```diff
@@ -0,0 +1,343 @@
...
+        depth_array, current_depth, invalid = metric_data
+
+        # Qubits have been already converted to their integer indices
+        # by the `jasp.get_qubit` primitive handler
+        qubit_ids = jnp.asarray(qubits, dtype=jnp.int64)
+        touched = jnp.take(depth_array, qubit_ids)
+        start = jnp.max(touched)
+
+        op = eqn.params["gate"]
+        duration = jnp.int64(_duration_for_op(op))
```
positr0nium: I'm now through with PR and it looks really great! The abstract Metric class significantly reduced the complexity of the ``count_ops`` interpreter and most likely also streamlined the ``depth`` interpreter by a lot.
There is one issue left with this very line. This line treats composite gates essentially as one big black box but in reality, the the decomposition can involve structure that skews the depth. Here is an example:

```python
from qrisp import *

qc = QuantumCircuit(2)
for i in range(10):
    qc.x(1)
    
gate = qc.to_gate()

def main():
    qv = QuantumVariable(2)
    
    for i in range(10):
        x(qv[0])
    
    qv.qs.append(gate, [qv[0], qv[1]])
    
    return qv
    
jitted_depth = depth(meas_behavior = "0")(main)()
static_depth = main().qs.transpile().depth()

print(jitted_depth)
# Yields 20
print(static_depth)
# Yields 10
```

PietropaoloFrisoni: Thanks, Raphael!

> This line treats composite gates essentially as one big black box

Yes, exactly. I tried to explain this in the ``Macro-gates and gate definitions`` section of the docstring.
Honestly, I was not even sure if it was correct to do something different here (that is, to track the depth separately on each qubit associated with a single multi-qubit gate).

Thanks for the example. I'll dive into it later today or tomorrow (I'll let you know if I have any questions) 👍

positr0nium: Yeah, as I mentioned the general concept of "depth" is really fuzzy because it depends on what you define as a basis gate and what is composite. It is much clearer when we are computing a particular depth (for instance t-depth in the Clifford + T gate set). By treating composite gates as "fundamental" we put in a behavior that makes the depth concept even more arbitrary, so I don't think that is a good idea.

PietropaoloFrisoni: Ok, I changed the implementation so that we transpile the gate definition and increment the depth separately on each qubit. 

P.S. I am doing some further improvements to the overall new profiler structure in the following PR (implementing the metric to track the qubit usage). I keep them separate otherwise it becomes a mess to review

MatP1337: I'm just commenting here as a note-to-self to be more mindful about separating the PRs and be nice to the person reviewing.

Apologies in advance about the block encoding class PR(s) and how they were done - lesson learned. Thanks @PietropaoloFrisoni for showcasing "best-practices" regarding this.

positr0nium (approved):
Amazing work! You got quite deep into it :)

PR #361, opened 2026-01-08
Stim conversion

positr0nium:
# Feature: Qrisp-Stim Integration & Noise Modeling Tools

## Summary
This PR introduces a comprehensive interface between Qrisp and **Stim**, a high-performance stabilizer circuit simulator. This new functionality allows users to export Qrisp code to Stim, model specific noise channels (Pauli errors, mismatched measurements), compute parity of measurement results, and extract Stim circuits directly from Jasp functions.

### 1. Circuit Conversion
- **`to_stim` / `qrisp_to_stim`**: Implemented the core logic to convert Qrisp `QuantumCircuit` objects into `stim.Circuit` objects.
- **Gate Support**: Includes mapping for standard Clifford gates and measurements. Classically controlled Pauli gates also work.
- **Measurement Tracking**: Returns maps (`return_measurement_map`, `return_detector_map`) to correlate Qrisp measurement objects/clbits with Stim's measurement record indices.

### 2. Traceable Stim Extraction (`@extract_stim`)
- **New Decorator**: Added `@extract_stim` to the Jasp module. This allows users to write quantum functions in Qrisp and immediately extract the corresponding Stim circuit without manual circuit management.
- **Return Value Handling**: Automatically separates Qrisp return values (like measurements or parity results) from the generated Stim circuit.

### 3. Noise Modeling Tools (`qrisp.misc.stim_tools`)
- **`StimNoiseGate`**: A new operation class to represent Stim-specific error channels within Qrisp circuits. These operations act as identity gates in standard Qrisp/Qiskit compilation but translate to noise instructions (e.g., `X_ERROR`) when converting to Stim.
- **`stim_noise`**: A convenience function to apply noise channels to variables/qubits.
- **Safety Restrictions**: Enforced strict limitations on `StimNoiseGate` (e.g., prohibiting `inverse`, `control` etc.) to prevent invalid noise configurations.

### 4. Parity Primitive
- **`parity` Function**: A new primitive for computing XOR of multiple measurement results, essential for syndrome calculations in error correction.
- **Nested Parity**: Supports parity of parity results with automatic symmetric difference (duplicate clbits cancel out for correct XOR semantics).
- **Stim Integration**: Parity results are converted to ``detector`` or ``observable`` in the Stim circuit.

### 5. Automatic Detector Discovery (`find_detectors`)
- **New Decorator**: `@find_detectors` automatically identifies stim `DETECTOR` instructions in quantum error correction circuits by leveraging the [tqecd](https://github.com/tqec/tqecd) library to analyze stabilizer flows.
- **Zero-effort QEC setup**: Users write standard Qrisp syndrome extraction circuits and the decorator discovers all valid parity checks, emitting the corresponding `parity()` calls automatically — no manual detector specification required.
- **Multi-round support**: Handles multi-round codes by splitting the stim circuit into per-round fragments (`_prepare_for_tqecd`) and detecting both intra-round and temporal (cross-round) detectors.
- **Static parameter support**: Non-`QuantumArray` arguments (integers, booleans, etc.) are captured by closure before JAX tracing and treated as compile-time constants inside the function body.
- **Debug mode**: `@find_detectors(return_circuits=True)` appends the raw stim circuit, the tqecd-prepared circuit, and the annotated circuit to the return value for inspection.
- **Noiseless expectation handling**: Automatically computes detector expectations via noiseless stim simulation, passing them to `parity()` so detectors work correctly even when the noiseless outcome is non-zero.
- **Filtering**: Detectors referencing measurements not present in the decorated function's return values are silently discarded, keeping only relevant detectors.
- **Constraints**: Requires `QuantumBool` arrays, Clifford-only gates, explicit resets between rounds, and the optional `tqecd` dependency.

### 6. Jaspr.to_qc() Array Manipulation
- **Measurement Arrays**: Full support for JAX array operations on measurement results during `Jaspr.to_qc()` conversion, including reshape, slice, concatenate, gather, transpose, and scatter.
- **ProcessedMeasurement**: Placeholder for measurement results that undergo classical post-processing (arithmetic, comparisons, bitwise ops) which cannot be represented in the circuit.
- **ParityHandle**: Placeholder for parity results. Semantically similar as ClBit but is not attached to the QuantumCircuit object,

## Documentation
- **New Module Documentation**: Created `reference/Utilities/stim_tools.rst` as the central hub for Stim documentation.
- **Guides & Examples**:
    - `extract_stim_doc`: How to use the extractor.
    - `stim_noise_doc`: How to inject noise (Bell pair examples).
    - `StimNoiseGate_doc`: API reference for the noise class.
    - `parity_doc`: Comprehensive examples for setting up syndromes (Repetition Code, Selective Noise).
    - `find_detectors_doc`: Usage guide and examples for the automatic detector discovery decorator.

## Testing
- **New Test Suites**:
    - `tests/interface_tests/test_stim_converter.py`: Validates the `to_stim` conversion logic, gate correctness, and map returns.
    - `tests/jax_tests/stim_extraction_tests.py`: Extensive testing of the JASP extraction, including parameter passing, return value structures, and parity logic.
    - `tests/jax_tests/test_qc_converter.py`: Tests for `Jaspr.to_qc()` including parity operations, array manipulation, and nested `@qache` with measurements.
    - `tests/stim_integration_tests/test_find_detectors.py`: Unit tests for the `@find_detectors` decorator covering single-round, multi-round, multi-array, partial return, static parameters, and debug-mode scenarios.
    - `tests/stim_integration_tests/test_find_detectors_vs_stim.py`: Validation tests that compare `@find_detectors` output against stim's built-in circuit generators to verify detector correctness.

### DIFF
--- a/src/qrisp/circuit/quantum_circuit.py
+++ b/src/qrisp/circuit/quantum_circuit.py
@@ -892,7 +892,7 @@ class QuantumCircuit:
             «      └─────────┘└────┘└─────────┘
 
         One can also transpile a specific composite gate in a QuantumCircuit, if desired. A Quantum
-        Phase Estimation circuit also contains a `QFT_dg` gate. 
+        Phase Estimation circuit also contains a `QFT_dg` gate.
 
         >>> from qrisp import p, QuantumVariable, QPE, multi_measurement, h
         >>> import numpy as np
@@ -903,11 +903,11 @@ class QuantumCircuit:
         >>>
         >>>     p(x*2*np.pi, qv[0])
         >>>     p(y*2*np.pi, qv[1])
-        >>> 
+        >>>
         >>> qv = QuantumVariable(2)
         >>>
         >>> h(qv)
-        >>> 
+        >>>
         >>> res = QPE(qv, U, precision = 3)
         >>>
         >>> print(qv.qs.compile())
@@ -921,11 +921,11 @@ class QuantumCircuit:
         >>>        return True
         >>>    else:
         >>>        return False
-        >>>    
+        >>>
         >>> transpiled_qc = test_circuit.transpile(transpile_predicate = transpile_predicate)
         >>>
         >>> print(transpiled_qc)
-        
+
 
         """
         from qrisp.circuit import transpile
@@ -1658,33 +1658,33 @@ class QuantumCircuit:
         return backend.run(self, shots)
 
     def statevector_array(self):
-        """
-        Performs a simulation of the statevector of self and returns a numpy array of
-        complex numbers.
+        r"""
+        Simulate the circuit statevector and return it as a NumPy array of complex
+        amplitudes.
 
         .. note::
 
-            Qrisps qubit ordering convention is reversed when compared to Qiskit,
-            because of simulation efficiency reasons.
-            As a rule of thumb you can remember:
+            The returned array uses **big-endian index ordering**. The array index
+            ``i`` maps to qubit values as
 
-            The statevector array of the following circuit has the amplitude 1 at the
-            index ``0010 = 2``
+            .. math::
 
-            .. code-block:: none
+                i = \sum_{k=0}^{n-1} q_k \, 2^{\,n-1-k},
 
-                qb.0: ─────
+            so :math:`q_0` is the most significant qubit. For two qubits this yields:
 
-                qb.1: ─────
-                      ┌───┐
-                qb.2: ┤ X ├
-                      └───┘
-                qb.3: ─────
+            - ``i = 0``  → :math:`|q_0=0, q_1=0\rangle`
+            - ``i = 1``  → :math:`|q_0=0, q_1=1\rangle`
+            - ``i = 2``  → :math:`|q_0=1, q_1=0\rangle`
+            - ``i = 3``  → :math:`|q_0=1, q_1=1\rangle`
+
+            This differs from Qrisp’s internal little-endian convention (only the
+            index-to-basis mapping changes).
 
         Returns
         -------
         numpy.ndarray
-            The statevector of this circuit.
+            The statevector of this circuit in big-endian order.
 
         Examples
         --------
@@ -1703,7 +1703,23 @@ class QuantumCircuit:
                 0.24999997+0.j, -0.24999997+0.j,  0.24999997+0.j, -0.24999997+0.j],
               dtype=complex64)
 
+        In this example, we create a :ref:`QuantumFloat` and prepare the normalized state
+        $\sum_{i=0}^3 \tilde b_i\ket{i}$ for $\tilde b=(0,1,2,3)/\sqrt{14}$.
 
+        >>> import numpy as np
+        >>> from qrisp import QuantumFloat
+        >>> b = np.array([0, 1, 2, 3], dtype=float)
+        >>> b /= np.linalg.norm(b)
+        >>> qf = QuantumFloat(2)
+        >>> qf.init_state(b)
+        >>> sv_array = qf.qs.statevector_array()
+        >>> print(f"b[1]: {b[1]:.6f} -> {sv_array[2]:.6f}")
+        b[1]: 0.267261 -> 0.267261-0.000000j
+        >>> print(f"b[2]: {b[2]:.6f} -> {sv_array[1]:.6f}")
+        b[2]: 0.534522 -> 0.534522-0.000000j
+
+        Here ``sv_array[2]`` corresponds to :math:`\ket{q_0=1, q_1=0}` and
+        ``sv_array[1]`` to :math:`\ket{q_0=0, q_1=1}`.
         """
         from qrisp.simulator import statevector_sim
 
@@ -2032,7 +2048,7 @@ class QuantumCircuit:
         from qrisp.interface import convert_to_cirq
 
         return convert_to_cirq(self)
-    
+
     # Several methods to apply the standard operation defined in standard_operations.py
     def measure(self, qubits, clbits=None):
         """
--- a/src/qrisp/simulator/simulator.py
+++ b/src/qrisp/simulator/simulator.py
@@ -212,7 +212,7 @@ def run(qc, shots, token="", iqs=None, insert_reset=True):
         return res
 
 
-@njit(cache = True)
+@njit(cache=True)
 def gen_res_dict(samples):
 
     hist = np.histogram(samples, np.arange(np.max(samples) + 2))[0]
--- a/src/qrisp/alg_primitives/state_preparation/prepare_func.py
+++ b/src/qrisp/alg_primitives/state_preparation/prepare_func.py
@@ -17,91 +17,149 @@
 """
 
 import numpy as np
-
 from jax.errors import TracerArrayConversionError
 
-from qrisp.alg_primitives.state_preparation.qswitch_state_preparation import prepare_qswitch
-from qrisp.alg_primitives.state_preparation.qiskit_state_preparation import prepare_qiskit
+from qrisp.alg_primitives.state_preparation.qiskit_state_preparation import (
+    prepare_qiskit,
+)
+from qrisp.alg_primitives.state_preparation.qswitch_state_preparation import (
+    prepare_qswitch,
+)
+from qrisp.jasp.tracing_logic import check_for_tracing_mode
 
-def prepare(qv, target_array, reversed=False, method = "auto"):
+
+def prepare(qv, target_array, reversed: bool = False, method: str = "auto"):
     r"""
-    This method performs quantum state preparation. Given a vector $b=(b_0,\dotsc,b_{N-1})$, the function acts as
+    Prepare a quantum state on ``qv`` from a target amplitude vector.
+
+    Given a vector :math:`b=(b_0,\dotsc,b_{N-1})` (corresponding to ``target_array``),
+    this routine prepares a state proportional to
 
     .. math::
 
-        \ket{0} \rightarrow \sum_{i=0}^{N-1}b_i\ket{i}
+        \sum_{i=0}^{N-1} b_i \ket{i}.
+
+    The ``target_array`` input vector is normalized internally, i.e.
+
+    .. math::
+
+        \tilde b_i = \frac{b_i}{\|b\|},
+        \qquad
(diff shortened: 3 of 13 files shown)


positr0nium:
Thanks! I think both suggestions are very doable and I will implement them soon.

In the past days I immersed myself with some more error correction theory so I can judge this also a bit better myself. One question that arose for me is the interface for detectors:
In Stim, the detector fires if the parity of the involved measurements deviates from the result of a noiseless simulation. On real hardware we however don't have the privilege of turning the noise off, so it is not possible to tell whether a detector is supposed to have even or odd parity. Since detectors are not just some simulation gimmick but most likely play a crucial role in real hardware executions, having this information is important for compiling error-corrected programs. I see two ways around it:

1. The program is stim-simulated once before the hardware execution and the intended parities are then included into the IR.
2. We include an optional keyword argument into the ``detector`` function, such that the user can indicate whether the parity is even or odd.

In reality I think option 1 already falls out: While Stim can in theory auto-generate this for standard static circuits, relying on this "auto-analyzer" forces us to always run a simulation before execution. This becomes problematic for very large-scale programs or dynamic workflows (like decoder feedback or magic state preparation via Repeat-Until-Success protocols), which can not be carried out by stim.

What do you think? Does this extra keyword argument make sense?
Also tagging @brianzi @jheinsoo @diehoq @JLenssen for input.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1890,6 +1890,116 @@ def to_pennylane(self):
...
+        -------
+        stim_circuit : stim.Circuit
+            The converted Stim circuit.
+        measurement_map : dict
+            (Optional) A dictionary mapping Qrisp Clbit objects to Stim measurement record indices. 
+            For example, ``{Clbit(cb_1): 2, Clbit(cb_0): 1}`` means ``Clbit("cb_1")``
+            corresponds to index 2 in Stim's measurement record.
+        detector_map : dict
+            (Optional) A dictionary mapping Qrisp Clbit objects to Stim detector indices.
+        detector_map : dict
```
renezander90: detector_map -> observable_map

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1890,6 +1890,116 @@ def to_pennylane(self):
...
+        which stim measurement index, we can use the ``return_measurement_map`` keyword
+        argument.
+        
+        >>> qc = QuantumCircuit(2, 2)
+        >>> qc.x(0)
+        >>> qc.cz(0, 1) 
+        >>> qc.measure(1, 1) # The first measurement of the circuit targets the second ClBit
+        >>> qc.measure(0, 0) # The second measurement of the circuit targets the first ClBit
+        >>> print(qc)
+              ┌───┐      ┌─┐
```
renezander90, line 1987: Even after restarting my kernel it looks like this:

Image

We should think again about the indices at some point (see https://github.com/eclipse-qrisp/Qrisp/issues/402).

src/qrisp/jasp/primitives/parity_primitive.py
```diff
@@ -0,0 +1,252 @@
...
+        When used within a function decorated with :func:`~qrisp.misc.stim_tools.extract_stim`, this function 
+        is translated into Stim's ``DETECTOR`` or ``OBSERVABLE_INCLUDE`` instructions in the generated circuit.
+        
+        * If ``observable=False`` (default), a ``DETECTOR`` instruction is created.
+        * If ``observable=True``, an ``OBSERVABLE_INCLUDE`` instruction is created.
+
+        If executed **without** ``extract_stim`` (i.e., in regular Qrisp simulation via :func:`~qrisp.jasp.jaspify`, for instance),
+        and ``observable=False``, the function verifies that the measured parity matches the ``expectation`` provided. 
+        Deviations from the parity expectation should solely stem from hardware noise,
+        i.e. a deviation in a **noiseless** simulation is an error *in the programm*.
```
renezander90: programm -> program

src/qrisp/jasp/primitives/parity_primitive.py
```diff
@@ -0,0 +1,252 @@
...
+            h(qv[0])
+            cx(qv[0], qv[1])
+            cx(qv[0], qv[2])
+            cx(qv[0], qv[3])
+            
+            a, b, c, d = tuple(measure(qv[i]) for i in range(4))
+            return parity(a, b, c, d)
+        
+        print(main())
+        # Yields 0
```
renezander90: Was the return type changed to bool (True/False)? I get this output ('False' merged with not-vanishing simulator bar):

Image

src/qrisp/jasp/primitives/parity_primitive.py
```diff
@@ -0,0 +1,252 @@
...
+            # Create arrays of measurements
+            a = jnp.array([True, False, True])
+            b = jnp.array([False, False, True])
+            
+            # Compute element-wise parity
+            result = parity(a, b)
+            return result
+        
+        print(array_parity())
+        # Yields [1, 0, 0]
```
renezander90: Executing this example yields: Exception: Parity expectation deviated from simulation result

src/qrisp/jasp/interpreter_tools/interpreters/qc_extraction_interpreter.py
```diff
@@ -14,181 +14,1029 @@
...
+        insert_outvalues(eqn, context_dic, outvalues)
+        
+    return qc_extraction_eqn_evaluator
+
 
+# =============================================================================
+# SECTION 5: Public API - jaspr_to_qc Function
+# =============================================================================
 
 def jaspr_to_qc(jaspr, *args):
```
renezander90, line 959: In the documentation only Jaspr.to_qc is shown (with the non-updated docstring).
And "Evaluation" table is there twice.

src/qrisp/misc/stim_tools/stim_backend.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Returns
+    -------
+    list[dict[str, int]]
+        List of result dictionaries, one per circuit.
+    """
+    return [run_on_stim(qc, shots) for qc, shots in batch]
+
+
+StimBackend = BatchedBackend(run_on_stim_batch)
```
renezander90: Would it make sense to show the StimBackend in the Backend Interface doc?

src/qrisp/simulator/simulator.py
```diff
@@ -18,6 +18,7 @@
 
 import threading
 import sys
+import shutil
```
renezander90, line 21: Image

In my vs code, the progress bar does not disappear after the simulation. I would assume that it has something to do with the changes in this file.

tests/interface_tests/repro_move_bug.py
```diff
@@ -0,0 +1,34 @@
+from qrisp import QuantumCircuit
+from qrisp.misc.stim_tools import StimNoiseGate
+import stim
+
+def test_repro():
```
renezander90: If it is only related to Stim, stim should appear in the filename.

src/qrisp/circuit/quantum_circuit.py
positr0nium: Is this about the naming of the qubits?

src/qrisp/circuit/quantum_circuit.py
positr0nium: Fixed https://github.com/eclipse-qrisp/Qrisp/pull/361/changes/8715341a6d51846e0fe2b108036d62928e31485e

src/qrisp/jasp/primitives/parity_primitive.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/4f48f9abe2722c0d3ce83b5f153d364ace831cb7

src/qrisp/jasp/primitives/parity_primitive.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/41e70441572345c36c3ee478bdc01212c5b5aa91

src/qrisp/jasp/primitives/parity_primitive.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/41e70441572345c36c3ee478bdc01212c5b5aa91

src/qrisp/jasp/interpreter_tools/interpreters/qc_extraction_interpreter.py
positr0nium: The docstring of jaspr_to_qc is for developers/AIs to quickly understand what the interface of the function is. The userfacing interface is the Jaspr method.
Removed the second evaluation keyword in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/3f11290ceef1718ed51906ca727685906027bf81

src/qrisp/misc/stim_tools/stim_backend.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/2f6e9a4039323d443ae43373527861a23b5b2109

src/qrisp/simulator/simulator.py
positr0nium: Attempted to fix in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/d249b4aa944618b089c91b9f1f51a930d44f748f
Please let me know if it worked.

tests/interface_tests/repro_move_bug.py
positr0nium: Removed the file in https://github.com/eclipse-qrisp/Qrisp/pull/361/commits/b785cf5366b082bf6164f967db3ca7ea87fc722e

src/qrisp/simulator/simulator.py
renezander90: Yes it works :)

src/qrisp/circuit/quantum_circuit.py
renezander90: Yes, but we have this discussion in the issues already. 
I guess, It could be confusing to some user, but I also see the practicality issues with other naming convention. 
This was more of an observation than an issue.

PR #362, opened 2026-01-08
deprecated the garbage_collection keyword of make_jaspr

positr0nium:
The ``garbage_collection`` keyword argument was used to automatically reset and delete local QuantumVariables when exiting a function scope. Since such variables are essentially no longer accessible to the user, they block qubit resources but can in no way contribute further to the computation. In particular it is not possible to measure them. Reseting these qubits or leaving them alive until the end of the computation there makes no difference to any observable quantity so the reasoning was we might as well reset them.

The reseting however introduces a "surprising" quantum operation (i.e. an operation that the user didn't really instruct but might cause extra costs or problems if the hardware doesn't support resets). According to the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) we should therefore remove the feature.

Furthermore the garbage collection feature required tracking in a dynamic way which QuantumVariables are "alive" and which are not, which was a continuous source of headache.

As a direct consequence the ``RUS`` decorator now behaves a bit differently, which is also reflected in the updated docstring:

```
    .. warning::
        
        The ``RUS`` decorator will automatically reset and deallocate any 
        QuantumVariables returned by the trial function, if the termination
        bool is not ``True``.
        It is however left up to the user to deallocate all LOCAL
        QuantumVariables/QuantumArrays within the trial function that are not returned.
        Any non-deallocated variables will accumulate over every iteration and
        block their corresponding qubits indefinitely.
```

### DIFF
--- a/tests/primitives_tests/arithmetic_tests/test_remaud_adder.py
+++ b/tests/primitives_tests/arithmetic_tests/test_remaud_adder.py
@@ -29,7 +29,7 @@ def test_L1_ladder():
         with invert():
             for i in range(N - 1):
                 cx(circ1[i], circ1[i + 1])
-        jsp = make_jaspr(main, garbage_collection="none")(1)
+        jsp = make_jaspr(main)(1)
         circ2 = jsp.to_qc(N)
 
         assert circ2.compare_unitary(circ1.qs)


documentation/source/general/tutorial/HHL.ipynb
renezander90, line 1: While I resonate with avoiding "surprising" quantum operations, this looks much more like low-level programming.

Could this be addressed similarly as the prevention of redundant allocations where we rewrite the Jaspr?

Basically, when using the ``@RUS``decorator, we would reset and delete all QuantumVariables, that are not returned,  at the **end** (not necessarily directly after measurement).

PR #366, opened 2026-01-11
Implemented jasp version of jax.lax.switch

renezander90:
To be consolidated with ``qswitch`` (https://github.com/eclipse-qrisp/Qrisp/pull/352)


renezander90:
Part of https://github.com/eclipse-qrisp/Qrisp/pull/374

PR #371, opened 2026-01-13
BlockEncoding class

renezander90:
This PR introduces a new BlockEncoding class. 

The current way of representing a block-encoding as a tuple has several design flaws:
-  if does not keep track of the normalization factor which is required for many applications; 
-  it allows for only one ancillary variable (only "num_qubits" is tracked), whereas block-encodings with multiple ancillary variables exist - for example via multiplication of block-encodings, or nested block-encodings
- a tuple is a rigid structure which does not allow for evolution without causing breaking changes

The current implementation of the class aims to provide a very basic structure that will likely evolve over time - as block-encodings are an active field of research. It can also be used to bundle various methods, e.g. for creating specialized block-encodings for different types of inputs (circular matrices, sparse matrices, etc.)

I would appreciate your comments @MatP1337 @positr0nium

### DIFF
--- a/src/qrisp/block_encodings/block_encoding.py
+++ b/src/qrisp/block_encodings/block_encoding.py
@@ -19,20 +19,23 @@
 from __future__ import annotations
 import inspect
 from dataclasses import dataclass
-from jax import tree_util
 from jax.tree_util import register_pytree_node_class
 import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


renezander90:
This PR will be part of GQSP PR.

PR #389, opened 2026-01-27
Refactor Cuccaro Adder (Attempt 2)

purva-thakre:
Continues work from #350

### DIFF
--- a/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
+++ b/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
@@ -15,7 +15,11 @@
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 ********************************************************************************
 """
-from qrisp import *
+from qrisp.core import QuantumVariable, x, cx, mcx
+from qrisp.qtypes import QuantumFloat
+from qrisp.environments import conjugate, custom_control
+from qrisp.misc import int_encoder
+from qrisp.jasp import jrange, jlen
 import jax.numpy as jnp
 
 @custom_control
@@ -73,7 +77,7 @@ def cuccaro_adder(a, b, c_out=None, ctrl = None):
         # create a QuantumFloat of the same size as the other quantum input
         q_a = b.duplicate()
 
-        with conjugate(q_a.encode)(a):
+        with conjugate(int_encoder)(q_a, a):
             # begin with q_a in the state |a>
             if c_out is not None:
                 cuccaro_adder(q_a, b, c_out = c_out)
@@ -114,7 +118,6 @@ def cuccaro_adder(a, b, c_out=None, ctrl = None):
     dim_b = jlen(b)
 
     ancilla = QuantumFloat(max_size)
-    ancilla[:] = 0
 
     if c_out is not None:
         ancilla2 = c_out


src/qrisp/alg_primitives/arithmetic/adders/ripple_carry_adder.py
```diff
@@ -19,297 +19,6 @@
...
-
-# Wrapper for the Cuccaro procedure
-# This function mainly serves to determine the input qubits of the Cucarro procedure
-# depending on the shape of the input QuantumFloats
-# The effect of this function is
-# qv1 += qv2
-
-
-@gate_wrap(is_qfree=True, permeability=[1])
-def inpl_add(
```
purva-thakre, line 115: removing this function leads to following unit test failures elsewhere:

- [ ] tests/algorithms_tests/test_QMCI.py::test_QMCI - ImportError: cannot import name 'inpl_add' from 'qrisp' (/home/runner/work/Qrisp/Qrisp/src/qrisp/__init__.py)
- [ ] FAILED tests/jax_tests/test_HHL_demo.py::test_HHL_demo - ImportError: cannot import name 'inpl_add' from 'qrisp.alg_primitives.arithmetic' (/home/runner/work/Qrisp/Qrisp/src/qrisp/alg_primitives/arithmetic/__init__.py)
- [ ] FAILED tests/jax_tests/test_arithmetic_tools.py::test_min_max - ImportError: cannot import name 'inpl_add' from 'qrisp' (/home/runner/work/Qrisp/Qrisp/src/qrisp/__init__.py)
- [ ] FAILED tests/primitives_tests/arithmetic_tests/test_comparisons.py::test_quantum_float_comparison - ImportError: cannot import name 'inpl_add' from 'qrisp' (/home/runner/work/Qrisp/Qrisp/src/qrisp/__init__.py)
- [ ] FAILED tests/primitives_tests/arithmetic_tests/test_comparisons.py::test_quantum_modulus_comparison - ImportError: cannot import name 'inpl_add' from 'qrisp' (/home/runner/work/Qrisp/Qrisp/src/qrisp/__init__.py)
- [ ] FAILED tests/primitives_tests/arithmetic_tests/test_quantum_arithmetic.py::test_quantum_arithmetic - ImportError: cannot import name 'inpl_add' from 'qrisp.alg_primitives.arithmetic' (/home/runner/work/Qrisp/Qrisp/src/qrisp/alg_primitives/arithmetic/__init__.py)
- [ ] FAILED tests/primitives_tests/arithmetic_tests/test_quantum_division.py::test_quantum_divison - ImportError: cannot import name 'inpl_add' from 'qrisp.alg_primitives.arithmetic' (/home/runner/work/Qrisp/Qrisp/src/qrisp/alg_primitives/arithmetic/__init__.py)

src/qrisp/alg_primitives/arithmetic/adders/ripple_carry_adder.py
```diff
@@ -125,9 +40,8 @@ def inpl_add(
 
         qf1 += qf2
 
-    There are two different algorithms available:
+    There is two one algorithm available:
```
purva-thakre:
```suggestion
    There is one algorithm available:
```

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -0,0 +1,198 @@
...
+        (
+            isinstance(a, QuantumFloat) ^ isinstance(b, QuantumFloat)
+        )
+        or
+        # verify both inputs are quantum
+        (
+            isinstance(a, QuantumFloat) and isinstance(b, QuantumFloat)
+        )
+    ):
+        raise ValueError("Attempted to call the CDKPM adder on invalid inputs.")
```
positr0nium: I think the other adders allow addition on quantum types that are not QuantumFloat. To me restricting it to QuantumFloat is an annecessary restriction but I also understand the other point. I guess it is a design decision. What do others think @renezander90 @diehoq @PietropaoloFrisoni ?

tests/primitives_tests/arithmetic_tests/test_cdkpm_adder.py
```diff
@@ -0,0 +1,191 @@
...
+        ctrl_qbl = QuantumBool()
+        if ctrl_qbl_value:
+            x(ctrl_qbl[0])
+        cuccaro_adder(a, b, ctrl=ctrl_qbl)
+        return measure(a), measure(b)
+    
+    jaspr = make_jaspr(run_jasp_adder_with_control)(2, 3)
+    result_a, result_b = jaspr(i, j)
+    assert result_a == expected_result[0]
+    assert result_b == expected_result[1]
```
positr0nium: A neat little trick that you can use to make your arithmetic implementation super safe for many edge cases is using ``boolean_simulation``. Using this feature you can very rapidly test thousands of input constellations. Example:

```python
from qrisp import *

@boolean_simulation
def main(N, L, j, k):
    
    A = QuantumFloat(N)
    B = QuantumFloat(L)
    A[:] = j
    B[:] = k

    cuccaro_adder(j, B)
    return measure(A), measure(B)
    
for N in range(2, 5):
    for L in range(2, 5):
        for j in range(2**N):
            for k in range(2**L):
                A, B = main(N, L, j, k)
                assert A == j
                assert B == (k+j)%(2**L)
```

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -0,0 +1,198 @@
...
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+from qrisp import *
+import jax.numpy as jnp
+
+@custom_control
```
positr0nium, line 25: I think there is a problem with the uncomputation in the custom controlled quantum-quantum case. The following code gives uncomputation error message:

```python
from qrisp import *

@boolean_simulation
def main(N, L, j, k):
    
    A = QuantumFloat(N)
    B = QuantumFloat(L)
    A[:] = j
    B[:] = k
    qbl = QuantumBool()
    qbl.flip()

    with control(qbl):
        cuccaro_adder(A, B)
    return measure(A), measure(B)
    
for N in range(2, 5):
    for L in range(2, 5):
        for j in range(2**N):
            for k in range(2**L):
                A, B = main(N, L, j, k)
                assert A == j
                assert B == (k+j)%(2**L)
```

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
purva-thakre: > I think the other adders allow addition on quantum types that are not QuantumFloat.

Yes, they do but the encoding methods for the other types were iffy. I could not get them to work. So, I decided to restrict to just the QuantumFloat. But maybe the others have some suggestions on this....

At least from what I understand, we can't encode an int/float in a QuantumVariable (it has to be a QuantumFloat type) but still they are mentioned in the other adder docstrings.

positr0nium: We can always use the ``int_encoder`` function. In fact I think using this function instead of ``QuantumFloat.encode`` would be more consistent with the other adders: They are "raw" adders, i.e. they don't care about non-trivial exponents. With the current implementation there is indeed exponent related behavior:
```python
from qrisp import cuccaro_adder, gidney_adder

qf = QuantumFloat(3, -2)

gidney_adder(1, qf)

print(qf)
# {0.25: 1.0}

qf = QuantumFloat(3, -2)

cuccaro_adder(1, qf)

print(qf)
# {1: 1.0}
```
In my opinion it is best to separate the encoding related logic from the pure addition logic i.e. the encoding logic is available in the QuantumFloat methods:
```python
from qrisp import *

qf = QuantumFloat(3, -2)

qf += 1

print(qf)
# {1: 1.0}
```

positr0nium: I know I recommended you `` QuantumFloat.encode`` (sorry for that) 😅

purva-thakre: > We can always use the int_encoder function. 

Technically, this is also used by the `QuantumVariable` encode method as a whole. 

But it doesn't work unless I change to the `QuantumFloat` type (I know it's in the definition too in terms of which value can be encoded). I have also tried `example_qv[:]=1`. The [docstring encode example for this class](https://qrisp.eu/reference/Core/generated/qrisp.QuantumVariable.encode.html#qrisp-quantumvariable-encode) only focuses on `QuantumFloat`. As a whole, I am not too sure if we should mention the generic `QuantumVariable` in the adder docstrings when an adder might only work on a QuantumFloat. 

https://github.com/eclipse-qrisp/Qrisp/blob/51f7e0dbaaaeccfcc0d86a9af0e12c29e04c6a5f/src/qrisp/core/quantum_variable.py#L730

Screenshot from 2026-03-05 14-27-23

One of the types mentioned in the remaud adder docstring is a generic `Quantumvariable`. But the unit tests for this adder only focus on a `QuantumFloat`. This made me think the non-quantum float mentions were leftovers from an older implementation or some other function's docstring. 

https://github.com/eclipse-qrisp/Qrisp/blob/51f7e0dbaaaeccfcc0d86a9af0e12c29e04c6a5f/src/qrisp/alg_primitives/arithmetic/adders/remaud.py#L249-L252

https://github.com/eclipse-qrisp/Qrisp/blob/51f7e0dbaaaeccfcc0d86a9af0e12c29e04c6a5f/tests/primitives_tests/arithmetic_tests/test_remaud_adder.py#L22

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
purva-thakre: Looks like the issue is with deleting `extension_anc_b` safely because the in-place addition is performed on `b`. The way I understand it, I do not think this variable can be safely deleted without uncomputing the in-place addition on b when `a` and `b` are of unequal size. 

You can see the same warning is not an issue when both `A` and `B` are of the same size as `extension_anc_b` is not required at all in this case.

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # reduce the size of a to the size of b if a is larger than b
+    effective_size_a = jnp.minimum(dim_a, dim_b)
+    a = a[:effective_size_a]
+
+    # create an extension ancilla to change the size of a when it is smaller than b
+    extension_size = jnp.maximum(0, dim_b - dim_a)
+    extension_anc_a = QuantumVariable(extension_size)
+    extended_a = a[:] + extension_anc_a[:]
+    a = extended_a
```
purva-thakre, line 115: As discussed earlier today, the adder now truncates the size first input if it is larger than the second input. 

This allows us to avoid creating an extension of the second input to be the same size as the first input. For the in-place addition, the second input and it's extension would have been used i.e. it is then difficult to delete the extension for the second input at the end. See https://github.com/eclipse-qrisp/Qrisp/pull/389#discussion_r2903699238

I have added a line in the docstring specifying this:

> Note that when the first input is larger than the second input, the function will perform
    modulo addition (relative to the size of the second input) after the first input is truncated to be the same size as the second input.

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -73,7 +77,7 @@ def cuccaro_adder(a, b, c_out=None, ctrl = None):
         # create a QuantumFloat of the same size as the other quantum input
         q_a = b.duplicate()
 
-        with conjugate(q_a.encode)(a):
+        with conjugate(int_encoder)(q_a, a):
```
purva-thakre, line 80: @positr0nium I would prefer to use the `encode` method as `int_encoder` does not catch issues with signed `QuantumFloat`.

positr0nium: Similar to the other adders, this function should be a "raw" adder, i.e. it just views its inputs as arrays of (qu)bits and performs the addition truth table. No typing semantics here. Using ``encode`` also prevents having other types than ``QuantumFloat`` as inputs.

purva-thakre: I think there is some misunderstanding here. 

On the line before `with conjugate`, we are creating a duplicate of `b` to make certain `q_a` is of the same type as `b`. In my previous comment, I used `QuantumFloat` as an example, but I think `with conjugate(q_a.encode)(a)` will ensure the encoding method of the relevant quantum type is used. This way, we do not have to worry about `int_encoder` encoding a bad value.

positr0nium: Yes, but what if ``b`` is a ``QuantumChar``? Then the function will throw an error because the typing of ``b`` prevents encoding a number. This is what I mean with "raw adder" - the typing semantics play no role, it just executes the adder and that's it

purva-thakre: > Yes, but what if b is a QuantumChar? Then the function will throw an error because the typing of b prevents encoding a number. 

Then, isn't the function working as expected here? Suppose my first input is `5` and my second input is of type `QuantumChar`. As expected, the cuccaro adder will fail when we use `encode`. This is where the extra checks in the main adder become useful. 

I would imagine the main in-place adder function
-  checks if the first input's value is of the same type as the value being encoded in the second input 
- if the first input is incompatible, convert it to the compatible type as `QuantumChar`  such that the cuccaro adder can then act on both inputs

PR #406, opened 2026-02-06
Metric to track qubit usage

PietropaoloFrisoni:
**Context**:  This PR introduced a new metric to keep track of how many qubits have been (de)allocated throughout the program in `jasp` mode. See the docstrings and examples for details.

**Description of the Change**:  As above.

**Benefits**:  The profiler can now compute one more metric.

**Possible Drawbacks:** None that I can think of.

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
@@ -237,6 +237,16 @@ class BaseMetric(ABC):
         - Outvars: (QuantumCircuit)
         """
 
+    @abstractmethod
+    def handle_parity(
+        self, invalues: Sequence, eqn: JaxprEqn, context_dic: ContextDict
+    ) -> Sequence:
+        """
+        Handle the `jasp.parity` primitive.
+
+        TODO: add semantics here.
+        """
+
     def handle_create_quantum_kernel(self, *_args, **_kwargs):
         """Handle the `jasp.create_quantum_kernel` primitive."""
 
@@ -383,56 +393,70 @@ def make_profiling_eqn_evaluator(metric: BaseMetric) -> Callable:
             scan_body = eval_jaxpr(
                 eqn.params["jaxpr"], eqn_evaluator=profiling_eqn_evaluator
             )
-            
+
             # Extract scan parameters
             num_consts = eqn.params["num_consts"]
             num_carry = eqn.params["num_carry"]
             length = eqn.params["length"]
             reverse = eqn.params.get("reverse", False)
             unroll = eqn.params.get("unroll", 1)
-            
+
             # Separate inputs
             consts = invalues[:num_consts]
-            init = invalues[num_consts:num_consts + num_carry]
-            xs = invalues[num_consts + num_carry:]
-            
+            init = invalues[num_consts : num_consts + num_carry]
+            xs = invalues[num_consts + num_carry :]
+
             # Create a wrapper function that includes constants
             if num_consts > 0:
+
                 def wrapped_body(carry, x):
-                    args = consts + list(carry) + list(x) if isinstance(x, tuple) else consts + list(carry) + [x]
+                    args = (
+                        consts + list(carry) + list(x)
+                        if isinstance(x, tuple)
+                        else consts + list(carry) + [x]
+                    )
                     result = scan_body(*args)
                     if not isinstance(result, tuple):
                         result = (result,)
                     return result[:num_carry], result[num_carry:]
+
             else:
+
                 def wrapped_body(carry, x):
                     args = list(carry) + (list(x) if isinstance(x, tuple) else [x])
                     result = scan_body(*args)
                     if not isinstance(result, tuple):
                         result = (result,)
                     return result[:num_carry], result[num_carry:]
-            
+
             # Call JAX scan with the reinterpreted body
             if len(xs) == 1:
                 xs_arg = xs[0]
             else:
                 xs_arg = tuple(xs)
-            
+
             if len(init) == 1:
                 init_arg = init[0]
             else:
                 init_arg = tuple(init)
-            
-            final_carry, ys = jax.lax.scan(wrapped_body, init_arg, xs_arg, length=length, reverse=reverse, unroll=unroll)
-            
+
+            final_carry, ys = jax.lax.scan(
+                wrapped_body,
+                init_arg,
+                xs_arg,
+                length=length,
+                reverse=reverse,
+                unroll=unroll,
+            )
+
             # Prepare output
             if not isinstance(final_carry, tuple):
                 final_carry = (final_carry,)
             if not isinstance(ys, tuple):
                 ys = (ys,)
-            
+
             outvalues = final_carry + ys
-            
+
             insert_outvalues(eqn, context_dic, outvalues)
 
         elif eqn.primitive.name == "jit":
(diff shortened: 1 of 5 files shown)


src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -401,6 +409,119 @@ def depth_counter(*args):
...
+    - increased whenever qubits are allocated (e.g., via ``QuantumVariable`` creation),
+    - decreased whenever qubits are explicitly deleted (e.g., via ``qv.delete()``),
+
+    and the decorated function returns the final counter value at program end.
+
+    This metric is useful for detecting allocations that are never deallocated
+    (e.g., missing ``delete()`` calls), and for understanding how many qubits
+    remain live after a circuit finishes.
+
+    Note that this is not the total number of qubits allocated over time,
```
MatP1337: Would it be possible to return both the "surviving number of qubits", as well as the max number of qubits allocated at a certain time during the alg? Having both would complete .resources() with all the information one desires when analyzing the costs of BlockEncoding.

Otherwise amazing addition! Kudos and thanks!

PietropaoloFrisoni: Thank you Mat!

Yes, there is potentially more useful information that we could return. For example, I was thinking that maybe we can return a dict like:

```
{allocated_qubits: ... , deallocated_qubits: ....}
```

where these refer to the 'cumulative' information throughout the algorithm.
Or, as you suggest, some other entry that represents the max number of qubits allocated at a certain time during the algorithm (so not just the cumulative information).

It should not be hard, but at this point, it is potentially safest to implement this after the release (I believe all the code should be merged by the end of this week), so I also have the time to ask around and see what people would prefer this metric to return, in addition to this information. Please let me know what you think : )

PietropaoloFrisoni: P.S. Ideally, using the information returned by this metric, you could plot the number of allocations/deallocations vs. time. Definitely something fun to work on 🚀

MatP1337: Sounds good! I agree that it's better not to rush, but this is a "must-have" feature the users of the BlockEncoding class would really appreciate (also from some feedback obtained today in a 4h exchange).

The reason for it being that some custom block encoding have significantly lower depth that they compensate with a (massive) ancilla overhead. Having these metrics side by side would allow for a thorough and fair/objective comparison across different approaches without hiding the details in fine print.

@renezander90, what do you think?

Hope to see asap (no pressure, tho) in 8.1 with an incognito version bump.

Thanks so much for the contributions!

PietropaoloFrisoni: P.P.S. I forgot to mention that this PR also fixes a bug in `depth` (although it should be a rare edge case)

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -95,78 +111,131 @@ def _measurement_body_fun(
...
     ) -> Sequence:
-        """Handle the `jasp.create_qubits` primitive."""
+        """
+        Handle the `jasp.create_qubits` primitive.
+
+        The `create_qubits_p` primitive has the following semantics:
+
+        - Invars: (size, QuantumCircuit)
+
+        - Outvars: (QubitArray, QuantumCircuit)
```
MatP1337, line 125: I'm really grasping at straws here, trying to find any typos, or something wrong with the PR, but I'm not familiar with the Invars/Outvars terminology. I'll assume that this is a common term and "best practice". If that's the case, ignore this. If that is not the case, feel free to quickly provide a bit of additional info

src/qrisp/jasp/interpreter_tools/abstract_interpreter.py
```diff
@@ -27,10 +27,74 @@
...
+        * ``Qubit``           ->  None
+        ...
+
+    - In ``depth``:
+        * ``QuantumCircuit``  ->  (depth_array, current_depth, invalid_flag)
+        * ``QubitArray``      ->  (lookup_table, logical_size)
+        * ``Qubit``           ->  integer qubit id
+        ...
+
+    and so on for other metrics.
```
MatP1337, line 64: Adding another example for the qubit amount could be useful. Simply from the above two I cannot really "see" how this looks for qubit count, and I don't immediately get how to then adapt this for max num of qubits at a certain time.

If you think this is redundant, then feel free to disregard this suggestion.

tests/jax_tests/test_num_qubits.py
MatP1337, line 1: really rigorous testing, thanks again for exposing me for best-practices! Learned a lot while reviewing this!

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
PietropaoloFrisoni: Thank you @MatP1337 , that is way too kind of you.

Yes, `invars` and `outvars` are the same terms used in the [JAX documentation](https://docs.jax.dev/en/latest/_autosummary/jax.extend.core.JaxprEqn.html#jax-extend-core-jaxpreqn) and appearing as attributes in the [JaxprEqn](https://github.com/jax-ml/jax/blob/979067ea0b9284b83a3432f06aed6c068f542a5b/jax/_src/core.py#L455) (which is the class that defines the jaxpr)

MatP1337: gotcha, consider this resolved then.

src/qrisp/jasp/interpreter_tools/abstract_interpreter.py
PietropaoloFrisoni: >  Simply from the above two I cannot really "see" how this looks for qubit count, and I don't immediately get how to then adapt this for max num of qubits at a certain time

I think you are right, but my impression is that it depends on the fact that, right now, this metric is not too useful.

What the metric currently does is the result of a series of historical coincidences/accidents. For example, the initial decision to return a single integer and choosing the same name as the static `num_qubits` method, or the fact that the `delete_qubits` primitive is no longer automatically called on every `jaspr` by default (this was changed recently in a separate PR). 

As mentioned before, I'd like to return to this metric as soon as possible and make it more useful/interesting.
Perhaps the most instructive thing is to return a dict with all the allocations/deallocations during the circuit, such as:

```
{"alloc_1": ..., "alloc_2": ..., , "dealloc_1": ...,} 
```

or some other data structure with the exact sequence of (de)allocations. So that you can really know how the qubits are (de)allocated in your circuit. We could also provide a `plot=False` argument that plots the (de)allocation as a function of time.

Sorry if this metric is not exactly what you wanted/needed 😅 
But I promise it will be more useful very soon 🚀

MatP1337: Yeah, thanks for clearly explaining. I think I get the idea now.

Don't be sorry. It was the first step in the right direction and, while it doesn't yet do what I'd like it to do, it is still laying the foundations for it. Who knows, perhaps just focusing on that the idea for the ``plot`` argument wouldn't arise

I really like that idea a lot! could be really useful for the users and to understand the nature of the algorithm, especially when these get too big to dissect through manually.

Looking forward to you exploring this direction and I have no doubt that also the ``max_qubits`` will be included real quick.

PR #410, opened 2026-02-09
Add concurrency option to CI

purva-thakre:
https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

This PR allows the option to cancel WIP jobs in a PR when there are multiple jobs running per commit. The `cancel-in-progress: true` line will only favor running a job for the latest commit.

### DIFF
--- a/.github/workflows/qrisp_test.yml
+++ b/.github/workflows/qrisp_test.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


purva-thakre:
This PR is ready to be merged.

purva-thakre:
It's not to improve the run time for the pipeline. This only stops running consecutive jobs for the same PR. 

Suppose I have a commit named `fix 1`, this triggers the pipeline. While the pipeline is running, suppose I forgot to add a code example to the docstring. I add the example, name this commit `fix 2` and push it. If I push `fix 2` while the pipeline for `fix 1` is running, there will be 2 different jobs that will be running. With this PR, the `fix 1` jobs are canceled in favor of `fix 2`.

Issue #412, opened 2026-02-09
Bug: controlled/inverted state preparation with method='qswitch'


renezander90:
```
from qrisp import *

@terminal_sampling
def main():

    coeffs = jnp.array([1.,2.,1.,3.])
    qb = QuantumBool()
    h(qb)
    qv = QuantumFloat(2)

    with control(qb):
        prepare(qv, coeffs, method="qswitch")

    return qv

main()
```
Yields: ValueError: Tried to evaluate jaxpr with insufficient arguments: expected 7 (including 3 consts), got 10.

and

```
from qrisp import *

@terminal_sampling
def main():

    coeffs = jnp.array([1.,2.,1.,3.])
    qb = QuantumBool()
    h(qb)
    qv = QuantumFloat(2)

    with invert():
        prepare(qv, coeffs, method="qswitch")

    return qv

main()
```
Yields: ValueError: Tried to evaluate jaxpr with insufficient arguments: expected 6 (including 3 consts), got 9.

Most likely, the ``prepare_qswitch``method should be wrapped in ``@custom_control`` and ``@custom_inversion`` @PietropaoloFrisoni @positr0nium

PietropaoloFrisoni:
Thanks @renezander90 . 

For context, I was not aware of `custom_control` and `custom_inversion` when I wrote the algorithm used in `prepare_qswitch`. My bad (since `prepare` was supposed to dispatch to `prepare_qswitch`, I thought that any wrapping logic applied to `prepare` would have automatically worked). I can look into this as soon as I have time.

In the meantime (that is, before the next release), I can open a PR and try to raise an informative `NotImplementedError` if you try to call `prepare_qswitch` as you show in your example (within the indentation level of `with invert()` or ` with control(qb):`). Please let me know if you have any objection to this.

Of course, if you think it is a very easy fix and/or if you want to add more context to the discussion, please let me know. Thanks!

PR #415, opened 2026-02-10
Custom inv for qswitch state prep

renezander90:
```
from qrisp import *

@terminal_sampling
def main():

    coeffs = jnp.array([1.,2.,1.,3.,1.,2.,1.,3.])
    qb = QuantumBool()
    h(qb)
    qv = QuantumFloat(3)

    prepare(qv, coeffs, method="qswitch")
    with invert():
        prepare(qv, coeffs, method="qswitch")

    return qv

main()
```
Yields: {0.0: 1.0} as expected

#412


renezander90:
Custom in/ctrl should not be necessary. Instead the bug should be fixed.

PR #421, opened 2026-02-12
Ensure a value can be correctly encoded in `QuantumFloat`

purva-thakre:
The changes in this PR verify a value can be correctly encoded in the provided `QuantumFloat`. An error is raised in static mode and the value is snapped to 0 in dynamic mode. 

- First code block raises `ValueError: Not enough qubits to encode value 10 in unsigned QuantumFloat of 4 qubits and exponent -2.`
~- Second code block returns the encoded value as `0` and prints the statement `Warning: Value cannot be safely encoded in the provided QuantumFloat.`~

```py
from qrisp import QuantumFloat
a = QuantumFloat(4, -2, signed = False)
a[:] = 10
```

### DIFF
--- a/src/qrisp/qtypes/quantum_float.py
+++ b/src/qrisp/qtypes/quantum_float.py
@@ -371,6 +371,7 @@ class QuantumFloat(QuantumVariable):
 
     # Define outcome_labels
     def decoder(self, i):
+        """Convert measurement outcome (integer) back to human-readable value."""
 
         if self.signed:
             res = _signed_int_iso_inv(i, self.msize) * jnp.float64(2) ** self.exponent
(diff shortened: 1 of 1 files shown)


purva-thakre:
I implemented a temporary workaround in [0b7e548](https://github.com/eclipse-qrisp/Qrisp/pull/421/commits/0b7e548fc35d7c100a982ab6cab400099957f753). 

```py
from qrisp import QuantumFloat
a = QuantumFloat(4, -1, signed = True)
a[:] = 3000
print(a)
```
raises `ValueError: Not enough qubits to encode value 3000 in signed QuantumFloat of 4 qubits and exponent -1.`

```py
from qrisp import QuantumFloat
a = QuantumFloat(4, -1, signed = False)
a[:] = 3000
print(a)
```
raises `ValueError: Not enough qubits to encode value 3000 in unsigned QuantumFloat of 4 qubits and exponent -1.`

purva-thakre:
One thing I have been struggling with is how to stop the computation in dynamic mode when I am trying to encode a large value in a tiny `QuantumFloat` . 

I found `checkify.check` in [JAX docs](https://docs.jax.dev/en/latest/debugging/index.html#functional-error-checks-with-jax-experimental-checkify) but this only works when the function is returning a value. Since the `encode` method for `QuantumFloat`does not return anything, the computation proceeds by encoding a bad value in the defined `QuantumFloat`. 

I could use `debug.print` in the case when the computation still proceeds to notify a user that the computation has proceeded with a 'bad encoding'.  @positr0nium @renezander90 If you have any thoughts on this, it would be helpful to understand why we don't use  `debug.print` for such cases. I remember Raphael has said in the past that the computation should be verified in standard Python mode before using it in dynamic mode.

purva-thakre:
As discussed with @positr0nium, this PR only raises an error for static mode. There is no additional check for the computation to fail in dynamic mode. The standard workflow should be to verify the code works as expected in static mode before executing it in dynamic mode.

Screenshot from 2026-03-03 11-01-34

src/qrisp/qtypes/quantum_float.py
```diff
@@ -400,6 +400,8 @@ def encoder(self, i):
         else:
             return res.astype(int)
 
+    # not used anwhere in this file, worth removing? Ask raphael.
+    # function is also not used anywhere in the codebase
     def sb_poly(self, m=0):
         """
         Returns the semi-boolean polynomial of this `QuantumFloat` where `m` specifies
```
purva-thakre: is it worth removing this function? It is not used anywhere in Qrisp?

src/qrisp/qtypes/quantum_float.py
```diff
@@ -442,6 +444,34 @@ def sb_poly(self, m=0):
         return 2**self.exponent * poly
 
     def encode(self, encoding_number, rounding=False, permit_dirtyness=False):
+        from jax import lax
```
purva-thakre:
```suggestion

```

purva-thakre:
Closed due to rebasing issues.

PR #425, opened 2026-02-13
Incorporate code + doc review from René

carlotta-k:


src/qrisp/algorithms/cold/problems/QUBO.py
renezander90: When running this with the same settings, I get '000000': [0.12024759504809902, np.float64(0.0)] as first result and the optimal solution with very low probability. 

When increasing N_steps to 30, I get 101101': [0.8577728445431091, np.float64(-3.4)] which is a good sign that the methods works properly. :)
But, did it really produce the result you show for N_steps=6? Or is this a mismatch between run_args and results?

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -54,7 +56,71 @@ class DCQOProblem:
...
+
+        # Create AGP
+        A_lam = sum([Y(i) for i in range(N)]) # uniform
+
+        # Function for uniform AGP coefficients
+        def alpha(lam):
+            A = lam * h 
+            B = 1 - lam
+            nom = np.sum(A + 4*B*h)
+            denom = 2 * (np.sum(A**2) + N * (B**2)) + 4 * (lam**2) * np.sum(np.tril(J, -1).sum(axis=1))
```
renezander90, line 101: line-break

src/qrisp/algorithms/cold/dcqo_problem.py
renezander90: What do the abbreviations COLD, LCD stand for?

renezander90, line 29: Link to the relevant paper(s) for this implementation would be very helpful.

src/qrisp/algorithms/cold/problems/QUBO.py
renezander90: Update: I isolated the problem:

- On c713621d2568269924a081caca6a079771fd019e, I get '000000': [0.12024759504809902, np.float64(0.0)] for N=6 , but solution quality decreases with increased step size: '000000': [0.6136461364613647, np.float64(0.0)] (for N=50), basically maximizing rather than minimizing.

- On 9b9479f4a8662a4e1766962944680ea8439adcc1 ("Minor change to H_init" - you flipped the sign), I get '000000': [0.12024759504809902, np.float64(0.0)] for N=6, but for N_steps=50: '101101': [0.96367, np.float64(-3.4)], so we are converging to the optimal solution :)

The latter (new) is of course the plausible behavior **which we should keep**, but previously we randomly got a good solution for N=6. I would suggest that you adjust to, e.g., N_steps=25 in the example and tutorial.

We also observe that the probability of the optimal solution first decreases to almost 0, then increases slowly, and then increases rapidly:

N=5: 0.022420448408968178
N=10: 0.011540115401154013
N=20: 0.002049979500204998 (decrease)
N=22: 0.007711969152123391
N=24: 0.07739184521630957
N=26: 0.21189423788475772
N=28: 0.4763390535621425
**N=30: 0.8577728445431091**

Do you have an idea why? Could the schedule be further optimized? Is the step size for <20 steps to large for the (counter)adiabatic evolution to work?

carlotta-k: Yes exactly, the sign in H_init was the crucial point. It seems to me that the example working so well before for N=6 steps was rather a coincidence. So I will update both the tutorial and the example to proper step sizes!

About the question why it increases to rapidly: It could be that this is the case because N=25 is at T/2. So this would mean that we need at least t = 0.5 or smaller in the trotterization. I will check this by decreasing/increasing the evolution time T and looking whether the behaviour stays the same.

PR #426, opened 2026-02-13
Strong Convolution Performance Upgrade 🥳

positr0nium:
Convolution_Comparison

### DIFF
--- a/src/qrisp/algorithms/gqsp/convolution.py
+++ b/src/qrisp/algorithms/gqsp/convolution.py
@@ -20,12 +20,9 @@ import numpy as np
 from qrisp import (
     QuantumVariable,
     QuantumBool,
-    conjugate,
-    p,
 )
-from qrisp.alg_primitives import QFT
+from qrisp.alg_primitives import gidney_adder
 from qrisp.algorithms.gqsp.gqsp import GQSP
-from qrisp.jasp import qache, jrange
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


renezander90:
The analysis might be a bit more subtile:

- For **small filters**, the **gidner_adder** variant clearly outperforms.
- For **very large filters**, the **QFT** variant could outperform at some point (length of the filter).

This point depends on the number of qubits . 

Here is a comparison for an L=100 filter for 30 qubits:

- gidney: {'rz': 101, 'h': 8613, 's': 2871, 'rx': 100, 'x': 98, 't': 5742, 'cx': 25740, 't_dg': 5742, 'measure': 2872}
- QFT: {'h': 60, 'cx': 7770, 'rz': 101, 'x': 98, 'rx': 100, 'p': 11520, 'measure': 1}

On a fault-tolerant architecture (the relevant regime for a degree 100 filter), the gidney variant will still outperform because of the overhead of approximating p gates with Clifford + T. 

**gidney** should be the default, but it might make sense to keep the fourier option (keyword). What's your take on that?

A similar trade-off exists in classical signal processing:
Bildschirmfoto 2026-02-14 um 11 00 56

positr0nium:
You already discovered the tradeoffs :) The core advantages of the Fourier adder are:

* Very CX-cheap controlled and uncontrolled version if a lot of additions can be chained onto the same variable.
* Can be implement in three lines of code (QFT conjugation + Phase gate loop).

The disadvantages:

* $\mathcal{O}(N^2)$ overhead for getting in and out of Fourier-space.
* Heavily relies on arbitrary angle gates.

Advantages of the Gidney-adder:

* Only 4 T gates per addition bit.

Disadvantages:

* Requires $\mathcal{O}(N)$ ancilla qubits.

The ancillae overhead is remedied with the Cuccaro adder but that one requires 8 T gates.

Tagging also @purva-thakre :)

PR #428, opened 2026-02-15
Frontpage upgrade

positr0nium:
Several visual upgrades for the front page

### DIFF
--- a/documentation/source/index.rst
+++ b/documentation/source/index.rst
@@ -76,7 +76,7 @@
 
             <a href="./reference/Core/generated/qrisp.QuantumVariable.html">
             <div class="example-img-plot-overlay">
-            <p class="sd-card-text">Build on NumPy-style, non-unitary linear algebra through block encodings and Chebyshev-based quantum signal processing</p>
+            <p class="sd-card-text">Explore ground states of lattice models using quantum signal processing</p>
             </div>
             </a>
 


documentation/source/_static/eclipse_font.png
MatP1337, line 1: While the color is same in this image, upon compiling html it shows as grey on my side.

Image

documentation/source/_static/LinearAlgebra.png
MatP1337, line 1: I have trouble seeing how the plot is showcasing the concepts of linear algebra. I strongly believe the following thumbnail showing the simple syntax would work much better:

Image

documentation/source/_static/plot_linear_subspaces.py
MatP1337, line 1: as above, I am not a huge fan of this thumbnail - seems rather random and disconnected from the content it's supposed to "invite to". See thumbnail above as an alternative.

documentation/source/index.rst
```diff
@@ -195,46 +71,50 @@
...
         :shadow: none
-        
+    
         .. raw:: html
 
-            <a href="./reference/Algorithms/QuantumBacktrackingTree.html">
+            <a href="./reference/Core/generated/qrisp.QuantumVariable.html">
             <div class="example-img-plot-overlay">
-            <p class="sd-card-text">Solve backtracking problems by leveraging quantum walks</p>
+            <p class="sd-card-text">Build on NumPy-style, non-unitary linear algebra through block encodings and generalised quantum signal processing</p>
```
MatP1337: I think this text could be improved. 

Better alternative: "Use block encodings as programming abstractions to perform Quantum Linear Algebra using the Numpy-like interface of our BlockEncoding class."

Reads better, more inviting, and more descriptive.

documentation/source/index.rst
```diff
@@ -436,10 +341,11 @@ Who is behind Qrisp
 
 .. raw:: html
 
+    <hr class="section-divider">
     <div class="code-example-text">
     
 
-Qrisp is an Eclipse open-source project developed mainly at `Fraunhofer FOKUS <https://www.fokus.fraunhofer.de/en/>`_, an industrial research facility based in Berlin. It is publicly funded by the `German ministry of economic affairs <https://www.digitale-technologien.de/DT/Navigation/DE/ProgrammeProjekte/AktuelleTechnologieprogramme/Quanten_Computing/Projekte/Qompiler/qompiler.html>`_ and the European Union with the aim to enable commercial use of quantum computation. To achieve this, we aim to open this field of research to a broad audience of developers. Furthermore we are proud to announce that Qrisp will become a part of the `Eclipse foundation <https://www.eclipse.org/>`_!
+Qrisp is an Eclipse open-source project developed mainly at `Fraunhofer FOKUS <https://www.fokus.fraunhofer.de/en/>`_ and `IQM <https://meetiqm.com/>`_. We aim to open quantum computation to a broad audience of developers.
```
MatP1337: I think it would be best to remove this text altogether and let the logos do the talking.

The argument being the true open-source nature of the project. Mentioning FOKUS and IQM might discourage others from looking into and contributing to it.

documentation/source/index.rst
MatP1337, line 1: I think the order of the thumbnails should be rearranged. While backtracking and Shor are a staple of Qrisp, I think that the more novel features should be highlighted. Jasp/native MLIR integration, Stim, BlockEncoding class, etc.

Is there a reason for the order of the thumbnails? I think that the hamiltonian simulation could be more inviting to the user than the backtracking tutorial.

MatP1337 (changes requested):

documentation/source/_static/LinearAlgebra.png
positr0nium: I understand the appeal of the syntax thumbnail, and I agree that concise syntax is one of your works strongest differentiators.
My concern is mainly frontpage audience fit: this tile needs to communicate value in 1–2 seconds to people with very different backgrounds (including non-experts), and this is where the [curse-of-knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) effect can become a problem.

The syntax snippet is powerful once someone has understanding of block encodings, but on a landing page it can read as generic Python with unexplained symbols.
That’s why I prefer a visually aligned motive for the homepage simply signalling "linear algebra" and then let the tutorial page deliver the syntax “wow effect” when the context exists. The statement you want to deliver (“syntax is simple”) can barely be expressed visually; it becomes persuasive only when the domain context is already established. This is why it is better delivered inside the tutorial rather than the frontpage.

So my position is not against the syntax - it’s about sequencing: concept first on the frontpage, syntax depth one click later.
I’m happy to keep iterating the visual to make the linear-algebra connection clearer.

documentation/source/index.rst
positr0nium: My two cents: starting out with something that people know and love (NumPy, linear algebra) and connect it towards the less well-known concepts is more inviting than the other way around. "You should meet the students where they are". 

Having said this, I'm also fine with your formulation. Just wanted to give the reasoning here. 😇

documentation/source/index.rst
renezander90: But the backtracking paper was just accepted :)
Jasp is listed under "Key Features".

documentation/source/_static/LinearAlgebra.png
renezander90: Let me also point out that it is actually not just about linear algebra, but also **signal processing**, a well-established concept from analysis. Hence, we could explore building upon the thumbnail form the filtering tutorial.

documentation/source/index.rst
renezander90: Related to the the above discussion:

It looks to me like we highlight relatable **applications** of quantum computing in the first part (Backtracking - solves constrained satisfaction problems, Shor - breaks RSA,...) and **features** below.

Hence, we could highlight the signal processing aspect (solves ground state preparation, a highly relevant application) in the first part, and the Block-encoding abstraction for Quantum Linear Algebra in the second part alongside Jasp.

documentation/source/index.rst
```diff
@@ -260,6 +140,27 @@ Key Features
...
+            :align: center
+
+        .. raw:: html
+        
+            <div class="key-features-text">
+        
+
+    
+        **Jasp integration** |br|
+        Exert hybrid real-time algorithm control with Catalyst and Jasp. Qrisp's JAX integration enables seamless hybrid quantum-classical workflows with automatic differentiation. `Learn more → <./general/tutorial/Jasp.html>`_
```
renezander90: Not sure about highlighting "automatic differentiation" here. This does not apply to the quantum part.

documentation/source/_static/LinearAlgebra.png
positr0nium: What about these?
Proposal_0
proposal_1

renezander90:
I don‘t quite understand what they represent.

Gesendet von Outlook für iOS<https://aka.ms/o0ukef>
________________________________
Von: Raphael Seidel ***@***.***>
Gesendet: Monday, February 16, 2026 10:09:11 AM
An: eclipse-qrisp/Qrisp ***@***.***>
Cc: Zander, René ***@***.***>; Assign ***@***.***>
Betreff: Re: [eclipse-qrisp/Qrisp] Frontpage upgrade (PR #428)


@positr0nium commented on this pull request.

________________________________

On documentation/source/_static/LinearAlgebra.png<https://github.com/eclipse-qrisp/Qrisp/pull/428#discussion_r2811354725>:

What about these?
Proposal_0.png (view on web)<https://github.com/user-attachments/assets/1cdf57c6-12a6-4f6b-91a3-82b47475ef2a>
proposal_1.png (view on web)<https://github.com/user-attachments/assets/b331a2f4-0345-4e35-a8d7-8529d0a8b09f>

documentation/source/_static/LinearAlgebra.png
MatP1337: What are these? They look cool but I don't see what it is supposed to be. That's why the equation/syntax works well imo because while yes, it's syntax, the person on the landing page can make sense of it rather fast and could be more eager to learn more.

While these are aesthetic, I struggle to see the relevance.

positr0nium: This a stylized block matrix. The very same can be said about the other panels, because they are supposed to be a *visualisation*. Syntax is not suitable to be visualized.

renezander90: Since my comments landed below I repeat them here: On first look I was confused what exactly they represent. I assume, it is a blend of the Heisenberg lattice and block-encoding? It looks nice! Maybe we can refine it still a bit to be more expressive.

positr0nium: btw: The visualisation of linear algebra is inspired by how [Wikipedia does it](https://en.wikipedia.org/wiki/Linear_algebra).

MatP1337: If the consensus is to have the visualization than sure, I'm down, but then further iterations should be made. While the first one is based on Wikipedia, it adds nothing particularly "quantum" to it.

The latest suggestions look nicer, but still lack that clarity of seeing the connection to the content it's supposed to represent. Perhaps we should brainstorm ideas here and then try to come with with a visualization of it.

I also agree with @renezander90 that a filtering feature could be also showcaseable as an application of the "feature" that is the GQSP-based ``BlockEncding`` class.

documentation/source/_static/eclipse_font.png
MatP1337: As mentioned in the meeting - perhaps using the official Eclipse white logo is the way to go here to stay compliant. Maybe having the white qrisp logo here would also work and the text in white. Bc of the background the contrast will be enough.

documentation/source/_static/LinearAlgebra.png
positr0nium: As discussed in yesterdays meeting, here are two more proposals:
One representing a grid of atoms in a "ferromagnetic state" and one representing Chebychev expansion.

image_0
image_1

MatP1337: While the first one is perhaps a tad nicer aesthetically, I greatly prefer the second one, since it contains more substance about the topic, as well as the tutorial.

I would, however move the plot a bit lower, if possible so that it fills more of the screen.

Awesome!

documentation/source/_static/eclipse_font.png
MatP1337: image

PR #431, opened 2026-02-17
rewrote the introduction page

positr0nium:
Tagging for review @renezander90 @MatP1337 @diehoq @PietropaoloFrisoni

### DIFF
--- a/documentation/source/reference/index.rst
+++ b/documentation/source/reference/index.rst
@@ -11,7 +11,7 @@ Qrisp is a high-level quantum programming framework, written in pure Python, tha
 Qrisp targets three goals simultaneously:
 
 * **Expressiveness**: A high-level programming interface with typed quantum variables, automatic uncomputation, and environment-based control flow that maps naturally to the way algorithms are described on paper.
-* **Performance**: Compilation to efficient circuits through qubit recycling, gate optimization, and  lowering via Jasp to LLVM-backed infrastructure that can handle thousands of qubits and trillions of gates.
+* **Performance**: Compilation to efficient circuits through qubit recycling, gate optimization, and lowering via Jasp to LLVM-backed infrastructure that can compile function calls with thousands of qubits and trillions of gates.
 * **Portability**: Output as standard circuit objects that run on physical backends from IBM, IQM, AQT, Rigetti, and others, as well as a wide range of simulators.
 
 Since Qrisp programs are Python, developers have direct access to the entire scientific Python ecosystem - NumPy, SciPy, JAX, and beyond - without any language boundary.


documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
-programs is also written in Python, giving developers direct
-access to the vast ecosystem of scientific and industrial libraries
-that Python has to offer. The framework’s output is circuit objects,
-which means that Qrisp programs can be run on many of today’s
-physical quantum backends and simulators.
+* **Expressiveness**: A high-level programming interface with typed quantum variables, automatic uncomputation, and environment-based control flow that maps naturally to the way algorithms are described on paper.
+* **Performance**: Compilation to efficient circuits through qubit recycling, gate optimization, and - via Jasp - lowering to LLVM-backed infrastructure that can handle thousands of qubits.
+* **Portability**: Output as standard circuit objects that run on physical backends from IBM, IQM, AQT, Rigetti, and others, as well as a wide range of simulators.
+
+Since Qrisp programs are Python, developers have direct access to the entire scientific Python ecosystem - NumPy, SciPy, Jax, and beyond - without any language boundary.
```
PietropaoloFrisoni:
```suggestion
Since Qrisp programs are Python, developers have direct access to the entire scientific Python ecosystem - NumPy, SciPy, JAX, and beyond - without any language barrier.
```

Suggested by Grammarly (optional suggestion) + spelling for JAX. I am not going to repeat the same suggestions for the remaining `JAX` entries in the page.

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+   * - :doc:`QiskitBackend <Backend Interface/index>`
+     - Any Qiskit-compatible backend (Aer, IBM hardware)
+   * - :doc:`AQTBackend <Backend Interface/index>`
+     - AQT Arnica cloud
+   * - :doc:`QiskitRuntimeBackend <Backend Interface/index>`
+     - IBM Qiskit Runtime sessions
+   * - :doc:`VirtualBackend <Backend Interface/index>`
+     - Run arbitrary dispatch code adhering to the Qrisp interface
+
+Qrisp also supports direct export to Qiskit circuit and other circuit formats for seamless interoperability.
```
PietropaoloFrisoni:
```suggestion
Qrisp also supports direct export to Qiskit circuits and other circuit formats for seamless interoperability.
```

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+* **MLIR** / **Catalyst MLIR** - intermediate stages for advanced optimization passes.
+
+
+Key Jasp Features
+^^^^^^^^^^^^^^^^^
+
+* **qache** - A caching decorator that traces quantum functions only once per calling signature, then reuses the cached Jaspr. Functions called multiple times incur tracing cost only on the first invocation.
+* **jrange** - A traceable loop construct that replaces Python's ``range``, enabling dynamic iteration counts in compiled code.
+* **Classical control flow** - Measurement results are native integers (not opaque classical bits), so they can drive ``if``/``else`` branching, arithmetic, or any Jax operation in real time.
+* **Repeat-until-success (@RUS)** - A decorator for trial-based quantum protocols that loop until a classical measurement condition is met.
```
PietropaoloFrisoni:
```suggestion
* **Repeat-until-success (@RUS)** - A decorator for trial-based quantum protocols that loops until a classical measurement condition is met.
```

PietropaoloFrisoni:
Ok, I fetched all remotes and I can build the documentation locally for this branch as well. Thanks @MatP1337 .

> But it is a slow proccess within Eclipse to first suggest you, open the voting, and then after a week and with a majority of votes, you are committer.

I would appreciate if you could start the suggestion/voting/decision process as soon as possible, both for me and @purva-thakre. Raphael already mentioned this to us, and of course we wanted to respect your time and your internal processes.

Taking into account the projects we are planning to accomplish and the fact that we have been contributing for 4 and 3 months, respectively, it seems to me that the time is mature enough to start the voting process.

Please let me know what you think. Thanks!

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+    import datetime
+    from qrisp import QuantumVariable
+
+    class QuantumDate(QuantumVariable):
+        def __init__(self, size, starting_date):
+            self.starting_date = starting_date
+            QuantumVariable.__init__(self, size)
+
+        def decoder(self, i):
+            return self.starting_date + datetime.timedelta(i)
```
PietropaoloFrisoni:
```suggestion
import datetime
from qrisp import QuantumVariable

class QuantumDate(QuantumVariable):
    """A quantum variable that represents a date."""
    
    def __init__(self, size, starting_date):
        self.starting_date = starting_date
        QuantumVariable.__init__(self, size)

    def decoder(self, i):
        return self.starting_date + datetime.timedelta(i)
```

This suggestions will surely mess up the documentation.

I think it would be good to have at least a docstring for the class : )

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
-* An accessible and flexible framework for creating and executing quantum algorithms on physical backends.
-   
-Qrisp is written in pure Python and the source code for Qrisp
-programs is also written in Python, giving developers direct
-access to the vast ecosystem of scientific and industrial libraries
-that Python has to offer. The framework’s output is circuit objects,
-which means that Qrisp programs can be run on many of today’s
-physical quantum backends and simulators.
+* **Expressiveness**: A high-level programming interface with typed quantum variables, automatic uncomputation, and environment-based control flow that maps naturally to the way algorithms are described on paper.
+* **Performance**: Compilation to efficient circuits through qubit recycling, gate optimization, and - via Jasp - lowering to LLVM-backed infrastructure that can handle thousands of qubits.
```
purva-thakre:
```suggestion
* **Performance**: Compilation to efficient circuits through qubit recycling, gate optimization, and  lowering via Jasp to LLVM-backed infrastructure that can handle thousands of qubits.
```

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+   * - :doc:`InversionEnvironment <Quantum Environments/InversionEnvironment>`
+     - Reverses (daggers) the enclosed operations
+   * - :doc:`ConjugationEnvironment <Quantum Environments/ConjugationEnvironment>`
+     - Applies :math:`U \cdot \mathrm{body} \cdot U^\dagger` patterns
+
+Environments compose naturally. A controlled operation inside another ControlEnvironment is handled by Qrisp's compiler intelligently - it does not naively synthesize a double-controlled version of every gate but instead leverages the structure for more efficient circuits.
+
+::
+
+    from qrisp import QuantumFloat, control, QFT, h
```
PietropaoloFrisoni, line 99:
```suggestion
    from qrisp import QFT, QuantumFloat, control, h
```

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+            return self.starting_date + datetime.timedelta(i)
+
+Quantum typing integrates with Python's type hints, enabling self-documenting function signatures and type-based dispatch in the Jasp compilation pipeline.
+
+
+QuantumArrays
+=============
+
+Multiple QuantumVariable instances of the same type can be organized into a :doc:`Core/QuantumArray` - an N-dimensional array that inherits from NumPy's ``ndarray``. This brings familiar operations like slicing, reshaping, transposing, and concatenation to quantum data, all operating as zero-cost views (no extra qubits allocated). For arrays of type QuantumFloat, matrix multiplication and tensor contraction are available.
+
```
purva-thakre, line 75: I think there is a language barrier here. What do you mean by `zero-cost views`?

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+   * - :doc:`ConjugationEnvironment <Quantum Environments/ConjugationEnvironment>`
+     - Applies :math:`U \cdot \mathrm{body} \cdot U^\dagger` patterns
+
+Environments compose naturally. A controlled operation inside another ControlEnvironment is handled by Qrisp's compiler intelligently - it does not naively synthesize a double-controlled version of every gate but instead leverages the structure for more efficient circuits.
+
+::
+
+    from qrisp import QuantumFloat, control, QFT, h
+
+    def QPE(psi, U, precision):
```
PietropaoloFrisoni, line 101: Should we say a few words (potentially just a docstring is enough) on what this function does?

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+* :doc:`Quantum Types/QuantumChar` / :doc:`Quantum Types/QuantumString` - Character and string types for symbolic computation.
+
+Creating a custom quantum type is straightforward - inherit from QuantumVariable and override the ``decoder`` method:
+
+::
+
+    import datetime
+    from qrisp import QuantumVariable
+
+    class QuantumDate(QuantumVariable):
```
PietropaoloFrisoni, line 59:
```suggestion
    class QuantumDate(QuantumVariable):
    """A quantum variable that represents a date."""
```

I think we should insert at least a docstring and/or something else that mention what this class is supposed to do

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
 
-QuantumVariables of the same type can be managed in a class called :doc:`Core/QuantumArray`. This class is an inheritor of the numpy ndarray class and thus provides many convenient and established features like slicing or reshaping.
+.. math::
 
-One of the unique features of Qrisp is the possibility to conveniently load classical data into the quantum computer using logic synthesis. This can be achieved using the :doc:`Core/QuantumDictionary` class. This class can be treated like a regular Python dictionary up until the point where a QuantumVariable is used as a key. This will then return another QuantumVariable where the values are entangled with the values of the key QuantumVariable as dictated by the content of the dictionary.
+    \begin{pmatrix} A/\alpha & \cdot \\ \cdot & \cdot \end{pmatrix}
 
-Using the concept of :doc:`Quantum Environments/index` it is possible to programm using many of the established paradigms from classical computing such as conditional execution of blocks of code. This is described in :doc:`Quantum Environments/ConditionEnvironment`.
+Three besides the basic signature, three additional constructors cover different use cases:
```
purva-thakre: this sentence might be missing a word....

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+
+Standard Python-based circuit construction hits a wall at scale: a 35-bit modular multiplication already takes ~20 seconds to compile, and typical RSA key sizes reach 2000 bits. Furthermore, many quantum algorithms require *real-time classical computation* - classical logic that executes during the quantum program, faster than decoherence - which an interpreted language fundamentally cannot provide.
+
+Jasp solves both problems by making Qrisp code traceable through `Jax <https://jax.readthedocs.io/>`_, Google's framework for high-performance numerical computing. Instead of executing a function with actual values, Jax sends *tracer* objects through it, recording operations into a functional intermediate representation (Jaxpr). Jasp extends this with quantum primitives, producing a ``Jaspr`` - a Jax-compatible IR that captures both quantum operations (gate applications, measurements) and classical real-time computations (arithmetic, control flow, even neural network inference) in a single representation.
+
+
+Using Jasp
+^^^^^^^^^^
+
+For many use cases, adopting Jasp is a single decorator:
```
purva-thakre, line 232: missing word/phrase here, should be `adopting Jasp as a single decorator`

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+Quantum computing promises strong speedups for problems across optimization, chemistry, and machine learning - but actually *programming* a quantum computer remains far harder than it should be. Most quantum frameworks still operate at the level of individual qubits and gates: the quantum equivalent of writing assembly. The result is code that is tedious to write, difficult to maintain, and nearly impossible to scale.
```
PietropaoloFrisoni, line 7: This is good rhetoric, but isn't it a bit too strong? Many frameworks (Qiskit, Cirq) are not purely assembly-level anymore (this might sound a little dismissive)

documentation/source/reference/index.rst
PietropaoloFrisoni: I am not a compiler engineer, but from the technical perspective LLVM doesn’t handle qubits, but rather handles classical control and optimized kernels.

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+        def decoder(self, i):
+            return self.starting_date + datetime.timedelta(i)
+
+Quantum typing integrates with Python's type hints, enabling self-documenting function signatures and type-based dispatch in the Jasp compilation pipeline.
+
+
+QuantumArrays
+=============
+
+Multiple QuantumVariable instances of the same type can be organized into a :doc:`Core/QuantumArray` - an N-dimensional array that inherits from NumPy's ``ndarray``. This brings familiar operations like slicing, reshaping, transposing, and concatenation to quantum data, all operating as zero-cost views (no extra qubits allocated). For arrays of type QuantumFloat, matrix multiplication and tensor contraction are available.
```
PietropaoloFrisoni: > This brings familiar operations like slicing, reshaping, transposing, and concatenation to quantum data, all operating as zero-cost views (no extra qubits allocated)

This is a strong claim. If always true, great! 

If not, we can potentially replace with `typically implemented as zero-cost views` or something similar

documentation/source/reference/index.rst
renezander90: thousands of qubits and trillions of gates.

documentation/source/reference/index.rst
```diff
@@ -2,77 +2,336 @@ Reference
...
+    from qrisp import QuantumVariable, h, cx
+
+    qv = QuantumVariable(3)
+    h(qv[0])
+    cx(qv[0], qv[1])
+    cx(qv[1], qv[2])
+    print(qv)
+    # Outcome: {'000': 0.5, '111': 0.5}
+
+Under the hood, each QuantumVariable is registered in a :doc:`Core/QuantumSession` that manages its lifecycle - qubit allocation, entanglement tracking, compilation, and deallocation. A system of automatic :doc:`Core/Session Merging` ensures that variables from different sessions are transparently unified when they interact, so in most cases the user never has to think about sessions at all.
```
renezander90: that manages its lifecycle: qubit allocation, gate application, compilation, and deallocation.

documentation/source/reference/index.rst
positr0nium: This means that slicing/reshaping/transposing/concatenation doesn't allocate additional qubits but instead it just "organizes" the access.

documentation/source/reference/index.rst
positr0nium: Why would it not be true? The only time you allocate additional qubits is when you call the default QuantumArray constructor.

documentation/source/reference/index.rst
positr0nium: LLVM backed infrastructure means MLIR specifically, for which qubit handling passes exist (at least theoretically).

documentation/source/reference/index.rst
positr0nium: Qiskit and Cirq allow us to build circuits and handle individual gates. Technically speaking, this is even below assembly. Assembly contains instructions like add or move, which is already one level higher than instructions like AND or XOR.

documentation/source/reference/index.rst
purva-thakre: It might be better to describe it in this manner then.

documentation/source/reference/index.rst
positr0nium: Your phrasing would imply that "jasp" is a "single decorator". For my taste the original formulation is more on point.

documentation/source/reference/index.rst
MatP1337: Perhaps might be useful to also comment on this in a brief sentence (as you've done in your response @positr0nium)

positr0nium: Made it more precise

documentation/source/reference/index.rst
PietropaoloFrisoni: Ok, I probably misunderstood what you mean here. Feel free to ignore this (we can chat another time) 👍

PR #433, opened 2026-02-18
Better type hints and fusion handling for the `depth` metric

PietropaoloFrisoni:
This PR does a few things:

- Fixes [this bug](https://github.com/eclipse-qrisp/Qrisp/issues/439)

- Adds handling for single qubits in the `fuse` primitive for the `@depth` decorator

- Raises a few warnings for tricky and nasty edge cases involving the computation of the depth of a quantum circuit (such as negative slice values). This could be implemented, but I am not confident in jumping on this before the upcoming release.

- Removes `initial_metric` for the metric base class (it is just confusing to keep it here)

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
@@ -123,13 +123,14 @@ def _warn_slice(idx_slice1, idx_slice2):
 
 # The computation currently does not fail if the delete/reset operation is called,
 # but this way we can print an informative message.
-def _warn_delete_reset():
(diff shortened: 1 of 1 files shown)


src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
```diff
@@ -347,6 +357,8 @@ def handle_delete_qubits(self, invalues, eqn, context_dic):
 
         _, metric_data = invalues
 
+        _warn_not_implemented("delete_qubits")
+
```
PietropaoloFrisoni, line 361: I will probably just replace this with a docstring since this can easily blow-up the output.

Please let me know what you think

PR #442, opened 2026-02-18
Fix qq_montgomery_multiply_modulus to handle arbitrary Montgomery shifts

### LINKED ISSUE #418: Quantum-Quantum modular multiplication between QuantumModuli
The following snippet:
```
@jaspify
def main():
    p=7

    a_0 = QuantumModulus(p)
    a_0[:]= 2
    a_1 = QuantumModulus(p)
    a_1[:]= 6
    
    b = QuantumModulus(p)
    b = a_0*a_1

    return measure(b)
print(main())
```

yields:
`IndexError: list index out of range`

This behaviour is intended (as per the docstring of the qq_montgomery_multiply_modulus method) and it is due to the factors not being in Montgomery form; however it is not consistent with the static case, in which multiplication in standard form does not cause any issue.

diehoq:
This PR fixes #418.

### DIFF
--- a/tests/jax_tests/test_jasp_modulus.py
+++ b/tests/jax_tests/test_jasp_modulus.py
@@ -194,3 +194,34 @@ def test_modulus_qq_multiply():
         return measure(a), measure(b), measure(c)
 
     assert test() == (7, 12, (7*12) % 13)
+
+
+def test_modulus_qq_multiply_standard_form():
+    """Regression test: qq_montgomery_multiply_modulus must work when both
+    inputs are in standard form (m=0).  Before the fix, the function used x.m
+    directly as the reduction shift, which was 0 for standard-form inputs and
+    produced wrong results."""
+    from qrisp import QuantumModulus, jaspify, measure
+
+    @jaspify
+    def test():
+        a = QuantumModulus(13)
+        a[:] = 7
+        b = QuantumModulus(13)
+        b[:] = 12
+        c = a * b
+        return measure(a), measure(b), measure(c)
+
+    assert test() == (7, 12, (7 * 12) % 13)
+
+    # Also test with a different modulus to cover more bit-width cases.
+    @jaspify
+    def test2():
+        a = QuantumModulus(31)
+        a[:] = 17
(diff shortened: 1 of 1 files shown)


src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
```diff
@@ -303,15 +309,24 @@ def qq_montgomery_multiply_modulus(x: QuantumModulus, y: QuantumModulus):
...
-    # res = qq_montgomery_multiply(x, y, N, m, inpl_adder)
-    # return res
+    # Compute the reduction shift from the modulus, exactly like montgomery_mod_mul:
+    #   m = ceil(log2((N-1)^2 + 1)) - n
+    # This is the number of extra bits needed so that the full integer product
+    # a*b fits in (n + m) bits, independent of the inputs' Montgomery shifts.
+    # NOTE: We compute n from N (not from x.size) to keep it a plain Python int,
+    # since x.size can be a JAX tracer during tracing.
+    import numpy as np
+    n = int(np.ceil(np.log2(int(N))))
```
renezander90, line 351: In order for this to work always properly, we should make the modulus N strictly static.

At the moment you can initialize a QuantumModulus with a dynamic modulus:
```
from qrisp import *

@jaspify
def main():

    a = QuantumFloat(3)
    a[:] = 5
    N = jnp.int32(measure(a))
    print(N)

    b = QuantumModulus(N)
    return measure(b)

main()
```
Yields: TypeError: while_loop body function carry input and carry output must have equal types, but they differ:

Note that the error occurs in the `measure`.

However, BigInteger can also be initialized with a traced **BigInteger**.

Why is it necessary to have static m, n here?

positr0nium: We should also account for the fact that the modulus can be a BigInteger. This would crash ``np.log2``. I don't think we need an implementation of log2 on BigInteger though. For this usecase it should be sufficient to loop over the bit-level representation.

PR #451, opened 2026-02-22
Deactivated warnings for depth profiler

renezander90:


positr0nium (approved):

positr0nium:
> > I would not know, how to describe this issue in a way that is helpful to the user. 
> 
> I would suggest: _the `depth` metric can currently handle the `slice` operation correctly only when the lower bound of the slice is strictly smaller than the upper bound_ (or something along those lines)
> 
> As for the warning about `delete`, @positr0nium already summarized effectively. In the description of the `max_qubits` argument, we can simply add that deleting qubits does not affect this number (namely, even if we delete qubits, they still contribute to increasing the total number of qubits we can handle up to `max_qubits`). I hope this is clear enough

I think the problem is the -1 syntactic sugar. This should not be a big issue. I think I can fix this later.

renezander90:
I added some warnings :). @positr0nium I assume you will take care of the bump?

(It's probably better to do a proper analysis in the coming week, than fixing on a Sunday evening before the release :).)

positr0nium:
The fix for the slicing was just 3 lines :) @renezander90 please check. I think it should be fine now (at least regarding the slicing).

src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -387,6 +387,16 @@ def circuit(n):
...
         :ref:`kernelized <quantum_kernel>` function.
 
+    .. warning::
+
+        The depth metric an experimental feature and may not behave as expected in certain edge cases.
+
+        -   The memory management operations ``reset`` and ``delete`` are currently ignored.
+            Qubits freed by these calls still count toward the ``max_qubits`` limit.
+        
+        -   This metric can currently handle the slice operation correctly only when
```
renezander90, line 397: Can we then remove this?

positr0nium (approved):

PR #452, opened 2026-02-24
Post-release improvements to the metric profiler

PietropaoloFrisoni:
**Context:** This PR implements some improvements to the profiler infrastructure that I didn't have time to implement before the release of `qrisp 0.8`.

**Description of the Change:** Here is a list of changes:

- `profiling_dic` is no longer required as an input to the `BaseMetric` abstract class, as this is only required by `count_ops`. The only required input for every metric is `meas_behavior`, which specifies how to handle different measurement values.

- We introduce an abstract property in the `BaseMetric` abstract class for the initial value of the metric for every class that inherits from it (before this was created outside the class). Crucially, this is only used once for the initial value of every metric, as after this step the QuantumState goes through the JIT pipeline.

- Two new abstract `cache_key` and `from_cache_key` methods (the second one is a `classmethod`) have been introduced. This improves how we currently re-construct the metric within the `get_compiled_profiler` function ( now every metric knows how to re-construct itself, instead of handling everything separately from outside the class)

- We add more tests for edge cases regarding the `slice` primitive for the `depth` metric computation.  

- Minor improvements (improved format, type annotations, etc.)

**Benefits:** Slightly more efficient and elegant.

**Possible Drawbacks:** None that I can think of.

**Related GitHub Issues:** None.

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/abstract_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/abstract_interpreter.py
@@ -20,8 +20,7 @@ from typing import Callable, Sequence
 
 import jax.numpy as jnp
 from jax import make_jaxpr
-from jax._src.core import JaxprEqn
-from jax.extend.core import ClosedJaxpr, Literal
+from jax.extend.core import ClosedJaxpr, Jaxpr, JaxprEqn, Literal
 
 from qrisp.jasp import check_for_tracing_mode
 
@@ -108,8 +107,19 @@ class ContextDict(dict):
         return res
 
 
-def exec_eqn(eqn, context_dic):
-    """Evaluate a single equation within the given context dictionary."""
+def exec_eqn(eqn: JaxprEqn, context_dic: ContextDict) -> None:
+    """
+    Evaluate a single equation within the given context dictionary.
+
+    Parameters
+    ----------
+    eqn : JaxprEqn
+        The equation to evaluate.
+
+    context_dic : ContextDict
+        The context dictionary mapping Jaxpr variables to their current values.
+
+    """
 
     invalues = extract_invalues(eqn, context_dic)
     res = eqn.primitive.bind(*invalues, **eqn.params)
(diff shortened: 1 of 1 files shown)


src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -393,9 +393,10 @@ def circuit(n):
 
         -   The memory management operations ``reset`` and ``delete`` are currently ignored.
             Qubits freed by these calls still count toward the ``max_qubits`` limit.
-        
-        -   This metric can currently handle the slice operation correctly only when 
-            the lower bound of the slice is strictly smaller than the upper bound.
+
+        -   Slice bounds follow Python semantics for common cases (negative indices,
+            out-of-bounds clamping, empty slices, etc.). However, non-unit steps are not supported.
```
PietropaoloFrisoni, line 398: If you try to use a non-unit step, [an informative error is raised before](https://github.com/eclipse-qrisp/Qrisp/pull/325) reaching this part of the code : )

tests/jax_tests/test_depth.py
```diff
@@ -657,6 +657,106 @@ def main(num_qubits1, num_qubits2):
 
         assert main(4, 4) == 3
 
+    def test_slice_negative_stop(self):
```
PietropaoloFrisoni, line 689: The sentences written as comments along each gate in each test have been confirmed in debugging mode. That is, with: 

```
jax.config.update("jax_disable_jit", True)
```

src/qrisp/jasp/interpreter_tools/interpreters/count_ops_metric.py
```diff
@@ -99,7 +134,7 @@ def handle_create_qubits(self, invalues, eqn, context_dic):
     def handle_get_qubit(self, invalues, eqn, context_dic):
 
         # Trivial behavior since we don't need qubit address information
-        return None
+        return invalues
```
positr0nium: Why does this no longer return None? Passing around SSA values that are not needed can cause a performance overhead.

src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
```diff
@@ -143,32 +114,67 @@ class DepthMetric(BaseMetric):
...
 
+    @classmethod
+    def from_cache_key(cls, cache_key) -> "DepthMetric":
+        meas_behavior, max_qubits = cache_key
+        return cls(meas_behavior, max_qubits)
+
+    def cache_key(self) -> Tuple[Callable, int]:
+        return (self.meas_behavior, self.max_qubits)
+
+    @property
```
positr0nium: What is the reasoning behind the ``property``? It is not a performance issue right now but could become one because it is unusual that a property call **allocates** resources.

src/qrisp/jasp/interpreter_tools/interpreters/count_ops_metric.py
PietropaoloFrisoni: I returned the original invalues to be compatible with the type hint in the abstract base class, but I agree that increasing that size of the context dic with something that is not needed can only cause issues (we should update the type hint or ignore pylance entirely). I'll revert this. Thanks!

src/qrisp/jasp/interpreter_tools/interpreters/depth_metric.py
PietropaoloFrisoni: Yes, the allocated resources here depend on `max_qubits`, which is not guaranteed to be small. I agree, this should not be a property (same for other metrics). If you are implicitly asking why this is part of the class, it is just because it is conceptually related to the class, but we only use it once (before jitting the evaluator, and after that everything is handled through the contextdic). 

I think the perfect design for this would be a `staticmethod`, but this is not possible because it depends on `self.max_qubit` which is an instance attribute.

PR #459, opened 2026-02-27
Catch unallowed values in `BigInteger`

purva-thakre:
As discussed on Discord. 

```py
from qrisp import BigInteger

# Create a BigInteger
# The size is the amount of 32-bit integers, size=4 yields a 128-bit integer
x = BigInteger.create_static(-128, size=4)

print(x())
```
The above code block fails with `ValueError: Input must be non-negative, got -128`. 

~The above code block fails with `JaxRuntimeError: Input must be non-negative. (check failed)`. Same if I replace `create` with `create_dynamic`~

### DIFF
--- a/src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
+++ b/src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
@@ -153,10 +153,6 @@ class BigInteger:
         Constructs a fixed-width BigInteger with exactly `size` limbs, interpreting
         the input modulo 2^(32*size). JIT-friendly.
 
-        .. warning::
-
-            When called with a negative value, the function will treat it as zero.
-
         Parameters
         ----------
         n : int or float or jnp.integer or jnp.floating


src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
```diff
@@ -136,6 +136,9 @@ def create_static(n, size):
         BigInteger
             Fixed-width representation of `n` modulo 2^(32*size).
         """
+        if n < 0:
+            raise ValueError(f"Input must be non-negative, got {n}")
```
purva-thakre:
```suggestion
            raise ValueError(f"Input must be non-negative, got {n}.")
```

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
```diff
@@ -169,6 +172,9 @@ def create(n, size):
         JAX's host integer range (typically up to 64 bits). For arbitrarily large
         Python integers, prefer `create_static`.
         """
+        from jax.experimental import checkify
+        
+        checkify.check(n >= 0, "Input must be non-negative.")
```
purva-thakre: using `checkify.check` has unintended consequences elsewhere. :sweat_smile:

The failures are related to using `checkify` in JIT compiled context : ValueError: Cannot abstractly evaluate a checkify.check which was not functionalized. This probably means you tried to stage (jit/scan/pmap/...) a `check` without functionalizing it through `checkify.checkify`.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
```diff
@@ -169,6 +176,9 @@ def create(n, size):
         JAX's host integer range (typically up to 64 bits). For arbitrarily large
         Python integers, prefer `create_static`.
         """
+        # Ensure input is non-negative
+        # if negative, treat as zero to avoid unintended large positive wraparound in the loop below.
+        n = jnp.maximum(n, 0)
```
purva-thakre: @positr0nium What are your thoughts to forcing a negative value to be `0 ` when in dynamic mode? I cannot think of any cons to this approach. I can't interrupt the computation with something akin to raising an error in static mode. With a `0` value, the end user will at least know something has gone wrong in their code. 

This way, there is no chance of it turning `-128` into `340282366920938463463374607431768211328` which is still a valid `BigInteger` value. 

image

purva-thakre:
As discussed in https://github.com/eclipse-qrisp/Qrisp/pull/421#issuecomment-3989940157, the standard workflow is to use static mode to verify the code works as expected and no errors are raised before executing in dynamic mode.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
positr0nium: This would cause silent errors, which can be very hard to debug. The current version is much better.

PR #461, opened 2026-03-03
Refactoring the core components of the codebase (example PR)

PietropaoloFrisoni:
**Context:** This PR introduces several improvements to the `QuantumCircuit` class. It is intended to be the first in a series of PRs aimed at gradually improving the robustness, scalability, and maintainability of the codebase.

The work originally started while addressing a `pylance` warning in the `extend` method of `QuantumCircuit`, which revealed a few additional issues and opportunities for improvement.

**Description of the Change:** Here is a list of changes:

- The first docstring example of `QuantumCircuit` has been fixed, updated, and tested. The same was done for the docstring example of the `extend` method.

- A test file for the `QuantumCircuit` class has been created. Although `QuantumCircuit` is one of the most user-facing core components of Qrisp, it previously did not have a dedicated test file.
This PR introduces a minimal test structure covering the code touched here, but additional tests will be added in future PRs.

- *Breaking change*: the default value of the `translation_dic` parameter in the `extend` method is now `None` instead of `"id"`. Using `"id"` as a default value suggested that a string was an intended argument type, while the docstring clearly describes a dictionary-based interface. In practice, the previous implementation only handled `"id"` and failed for any other string. The new behavior makes the expected type explicit and avoids misleading defaults.

- Minor improvements, including formatting cleanup and additional type annotations.

**Benefits:** 

- Improved correctness and consistency

- Better test coverage for a core component

- Clearer typing and improved IDE/static analysis support

- More maintainable code going forward

**Possible Drawbacks:** None that I can think of (except for the breaking change above)

**Related GitHub Issues:** None.

### DIFF
--- a/src/qrisp/circuit/quantum_circuit.py
+++ b/src/qrisp/circuit/quantum_circuit.py
@@ -16,10 +16,13 @@
 ********************************************************************************
 """
 
+from hashlib import sha256
 from typing import Dict, List, Set
 
 import numpy as np
 import sympy
+from qiskit import QuantumCircuit as QiskitQuantumCircuit
+from qiskit.visualization import circuit_drawer
 
 import qrisp.circuit.standard_operations as ops
 from qrisp.circuit import Clbit, Instruction, Operation, Qubit
@@ -511,12 +514,9 @@ class QuantumCircuit:
         self.data = temp_data
         return res
 
-    # TO-DO write qiskit independent printer
+    # TODO write qiskit independent printer
 
-    # Printing method
     def __str__(self):
-
-        from qiskit.visualization.circuit_visualization import circuit_drawer
 
         from qrisp.interface import convert_to_qiskit
 
@@ -1055,8 +1055,6 @@ class QuantumCircuit:
         from qrisp.interface import convert_to_qiskit
 
         qiskit_qc = convert_to_qiskit(self, transpile=False)
-
-        from qiskit.visualization import circuit_drawer
 
         return circuit_drawer(qiskit_qc, output="latex_source", **kwargs)
 
@@ -1710,7 +1708,6 @@ class QuantumCircuit:
         return statevector_sim(self)
 
     def __hash__(self):
-        from hashlib import sha256
 
         res = 0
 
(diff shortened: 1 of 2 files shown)


src/qrisp/circuit/quantum_circuit.py
```diff
@@ -455,32 +451,22 @@ def extend(self, other, translation_dic="id"):
...
-            clbits = []
-
-            for cb in instruction_other.clbits:
-                clbits.append(translation_dic[cb.identifier])
-
+        if isinstance(translation_dic, str):
+            if translation_dic != "id":
+                raise ValueError(
+                    "Invalid translation_dic string. Expected 'id' or a dict mapping qubits of other to qubits of self."
+                )
```
PietropaoloFrisoni: This was done to prevent a breaking change, but I think that here `None` as default value would make much, much more sense. By using `"id"` as a default value, we are suggesting that a string is the suggested type, but actually the previous version was only able to handle `"id"` and failed with any other string.

Edit: I introduced a breaking change for the reason explained above.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -205,34 +201,39 @@ class QuantumCircuit:
 
     """
 
-    qubit_index_counter = np.zeros(1, dtype=int)
-    clbit_index_counter = np.zeros(1, dtype=int)
-    xla_mode = 0
+    qubit_index_counter: np.ndarray = np.zeros(1, dtype=int)
+    clbit_index_counter: np.ndarray = np.zeros(1, dtype=int)
+    xla_mode: int = 0
```
PietropaoloFrisoni, line 206: I am not a big fan of using a mutable data structure (such as a `NumPy` array) to label qubits, as this results in random label values (it depends on how many time this code is compiled when importing the packages). However, this kind of changes is out of scope for the current PR

positr0nium: The problem was/is that it is not straight forward to update class attributes:

```python
class Foo:
    counter = 0

    def __init__(self):
        self.counter = self.counter + 1

    def __str__(self):
        return str(self.counter)
    
a = Foo()
b = Foo()
c = Foo()
d = Foo()
print(d)
# Yields 1
```
This however works fine and we can adopt it:

```python
class Foo:
    counter = 0

    def __init__(self):
        Foo.counter = self.counter + 1

    def __str__(self):
        return str(self.counter)
    
a = Foo()
b = Foo()
c = Foo()
d = Foo()
print(d)
# Yields 4
```

PR #465, opened 2026-03-04
Add xDSL dialect definition for the JASP quantum dialect

bruno-ah-um:
- This PR introduces `src/qrisp/jasp/mlir/xdsl_dialect.py`, which registers the JASP quantum dialect with xDSL. This covers all three types (`QuantumState`, `Qubit`, `QubitArray`) and all twelve operations (`create_qubits`, `get_qubit`, `get_size`, `slice`, `fuse`, `reset`, `measure`, `delete_qubits`, `create_quantum_kernel`, `consume_quantum_kernel`, `quantum_gate`, `parity`) as typed `IRDLOperation` subclasses.

- It loads `JaspDialect` into the xDSL Context in `jaxpr_lowering.py` so the dialect is active whenever Qrisp emits MLIR.

- It expands `tests/jax_tests/test_mlir.py` with tests covering basic dialect operations.

Before this PR, JASP types and op names were printed with surrounding quotes (e.g. !"jasp.QuantumState", "jasp.create_qubits"(...)), meaning xDSL treated them as unregistered. After this PR, they print in standard MLIR assembly format (!jasp.QuantumState, jasp.create_qubits ...), enabling typed pattern matching in future xDSL passes.

### DIFF
--- a/src/qrisp/jasp/mlir/xdsl_dialect.py
+++ b/src/qrisp/jasp/mlir/xdsl_dialect.py
@@ -145,7 +145,7 @@ class CreateQubitsOp(IRDLOperation):
     assembly_format = (
         "$amount attr-dict `,` $qst_in"
         " `:` type($qst_in) `,` type($amount)"
-        " `->` type($result)"
+        " `->` type($result) `,` type($qst_out)"
     )
 
 
--- a/src/qrisp/jasp/mlir/dialect_definition/JaspOps.td
+++ b/src/qrisp/jasp/mlir/dialect_definition/JaspOps.td
@@ -16,7 +16,7 @@ def CreateQubitsOp : JaspOp<"create_qubits"> {
     let arguments = (ins 0DTensorOf<[I64]>:$amount, QuantumStateType:$qst_in);
     let results = (outs QubitArrayType:$result, QuantumStateType:$qst_out);
     
-    let assemblyFormat = "$amount attr-dict `,` $qst_in  `:` type($qst_in) `,` type($amount) `->` type($result)";
+    let assemblyFormat = "$amount attr-dict `,` $qst_in  `:` type($qst_in) `,` type($amount) `->` type($result) `,` type($qst_out)";
 }
 
 def GetQubitOp : JaspOp<"get_qubit"> {


src/qrisp/jasp/mlir/xdsl_dialect.py
```diff
@@ -0,0 +1,422 @@
...
+    amount = operand_def(_TensorI64)
+    qst_in = operand_def(QuantumStateType)
+
+    result = result_def(QubitArrayType)
+    qst_out = result_def(QuantumStateType)
+
+    assembly_format = (
+        "$amount attr-dict `,` $qst_in"
+        " `:` type($qst_in) `,` type($amount)"
+        " `->` type($result)"
```
positr0nium: This should also return the QuantumState, no?

src/qrisp/jasp/mlir/xdsl_dialect.py
```diff
@@ -0,0 +1,422 @@
...
+    meas_q = operand_def(_QubitOrArray)
+    in_qst = operand_def(QuantumStateType)
+
+    meas_res = result_def(_TensorInt)
+    out_qst = result_def(QuantumStateType)
+
+    assembly_format = (
+        "$meas_q `,` $in_qst attr-dict"
+        " `:` type($meas_q) `,` type($in_qst)"
+        " `->` type($meas_res) `,` type($out_qst)"
```
positr0nium, line 265: I think xDSL provides a typing verification mechanism. For this you have to overload ``Operation.verify_`` (I'm not sure though. Maybe @superlopuh can give some pointers here). This verification could be used to ensure that measurement operations that receive a QubitArray indeed return ``Tensor<int64>`` and measurement operations that receive a Qubit indeed return ``Tensor<int1>``. 
A similar verification might also be helpful for QuantumGateOp. Here the verification should ensure that the order of arguments is indeed *qubits, *gate_parameters, *quantum_state.

positr0nium (changes requested):

src/qrisp/jasp/mlir/xdsl_dialect.py
```diff
@@ -0,0 +1,422 @@
...
+
+    QubitArrays enable expression of dynamically-sized programs. They are
+    semantically equivalent to immutable arrays of integers.
+    """
+
+    name = "jasp.QubitArray"
+
+
+# Shorthand constraint for operands that accept either a Qubit or a QubitArray.
+_QubitOrArray = AnyOf([base(QubitType), base(QubitArrayType)])
```
superlopuh: I'd expect this to work, I find this a little prettier but it's a matter of preference, it should give exactly the same constraint in the end (and my version is a little slower). I say this because in most cases you can just pass `QubitType | QubitArrayType` directly wherever you would pass `_QubitOrArray` instead, which would remove one step of indirection.
```suggestion
_QubitOrArray = irdl_to_attr_constraint(QubitType | QubitArrayType)
```

src/qrisp/jasp/mlir/xdsl_dialect.py
superlopuh: Yep, I would go for a `verify_` method to do what you describe.

src/qrisp/jasp/mlir/xdsl_dialect.py
bruno-ah-um: I implemented this as in JaspOps.td. I can make both of them return a QuantumState.

https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/jasp/mlir/dialect_definition/JaspOps.td#L12-L20

src/qrisp/jasp/mlir/xdsl_dialect.py
bruno-ah-um: ok, I added verification, see https://github.com/eclipse-qrisp/Qrisp/pull/465/changes/8b7aa5ff8f57f680edafb3f117d2f2f643f0ce8f

src/qrisp/jasp/mlir/xdsl_dialect.py
bruno-ah-um: done in https://github.com/eclipse-qrisp/Qrisp/pull/465/changes/55c22c568ffa6b91d6bd682040467b9d5982695c

src/qrisp/jasp/mlir/xdsl_dialect.py
```diff
@@ -0,0 +1,454 @@
...
+    (parametric) gates.
+    """
+
+    name = "jasp.quantum_gate"
+
+    gate_type = attr_def(StringAttr)
+    # All operands are collected here: zero or more gate parameters/qubits
+    # followed by the quantum state as the last element.  Use the `gate_params`
+    # and `in_qst` properties to access them by role.
+    gate_operands = var_operand_def()
```
superlopuh: why not two sets of operands?

bruno-ah-um: Yes, if makes also verify simpler, see my commit: https://github.com/eclipse-qrisp/Qrisp/pull/465/changes/b95b40bffbfa0a164fd2050a36099f5971dd69c3

src/qrisp/jasp/mlir/xdsl_dialect.py
positr0nium: Ah yes, this is a bug in the dialect definition. Feel free to adjust it such that it also returns the QuantumState in the assembly. 🙏

bruno-ah-um: Yes, it should return a QuantumState, I fixed it: https://github.com/eclipse-qrisp/Qrisp/pull/465/changes/38cca526af641fcedf168c50e1fc9714f38abf8a

PR #473, opened 2026-03-10
Gidney Adder Refactor

### LINKED ISSUE #393: Multiple variations of `gidney` adder
There are multiple definitions/variations of the the Gidney adder. To avoid confusion, we should condense these into a single function and get rid of the extras.

- `jasp_gidney_adder` in `src/qrisp/examples/jasp_gidney_adder.py`
- Main 'gateway' function `gidney_adder` in `src/qrisp/alg_primitives/arithmetic/adders/gidney/gidney_adder.py`. This function relies on 4 different subfunctions: `jasp_qq_gidney_adder, jasp_cq_gidney_adder, qq_gidney_adder, cq_gidney_adder`. 

Based on a conversation with @positr0nium, the Gidney adder might have a better performance in static mode. We should consider refactoring the existing code (in addition to removing the subfunctions) by comparing the performance of the Gidney adder in static and dynamic mode.

purva-thakre:
Fixes #393


tests/primitives_tests/arithmetic_tests/test_qcla.py
```diff
@@ -343,13 +343,13 @@ def test_cq_qcla_adder():
...
     qc = b.qs.compile(workspace = 10, gate_speed = gate_speed, compile_mcm = True)
-    assert qc.t_depth() < 6
+    assert qc.t_depth() <= 6
     
     a = QuantumFloat(40)
     b = QuantumFloat(40)
     qcla(a, b)
     qc = b.qs.compile(workspace = 50, gate_speed = gate_speed, compile_mcm = True)
-    assert qc.t_depth() < 23
+    assert qc.t_depth() <= 26
```
purva-thakre, line 355: To do: Undo these changes to the QCLA adder unit tests. 

The current version of the Gidney adder works as expected if the comparison is between the `t_depth` of Cuccaro and Gidney adders. For some reason, the `t_depth` of the QCLA adder goes up due to changes to the Gidney adder. These unit tests were changed to account for this. 

Both inputs are of the same size in the plot below. 

Image

src/qrisp/alg_primitives/arithmetic/adders/gidney/qq_gidney_adder.py
```diff
@@ -1,127 +0,0 @@
...
-
-from qrisp.core import QuantumVariable
-from qrisp.qtypes import QuantumBool
-from qrisp.environments import invert, custom_control
-from qrisp.core.gate_application_functions import mcx, cx
-
-
-# This function performs the Gidney adder from https://arxiv.org/pdf/1709.06648.pdf
-@custom_control
-def qq_gidney_adder(a, b, c_in=None, c_out=None, ctrl=None):
```
purva-thakre, line 27: Question: is it worth adding deprecation warnings for the different Gidney adder variants that are being removed in this PR? 

As far as I can tell, only the main function `gidney_adder` is public facing.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Based on https://arxiv.org/abs/1709.06648. Works in both standard
+    Python and jasp tracing modes.
+    """
+
+
+    if not isinstance(a, (QuantumVariable, DynamicQubitArray, list)):
+
+        if isinstance(a, str):
+            a = int(a[::-1], 2) if a else 0
```
purva-thakre, line 27: had to add this check due to the following: 

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/src/qrisp/alg_primitives/arithmetic/adders/qcla/classical_quantum/cq_sum_path.py#L34-L37

`cq_sum_path` uses the Gidney adder on L58 and L60 with string inputs. `cq_gidney_adder` allowed string input. 

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/src/qrisp/alg_primitives/arithmetic/adders/gidney/cq_gidney_adder.py#L65

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if not check_for_tracing_mode() and isinstance(a, (int, np.integer)):
+            a = int(a) % (1 << len(b))
```
purva-thakre: this is a temporary addition due to the following line in `inpl_q_int_mult`

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/src/qrisp/alg_primitives/arithmetic/ripple_mult.py#L73

This can lead to `Not enough qubits to encode integer N` error from `int_encoder` in the following unit test. It might be better to change the adder rather than adding this check here (which should go into the adder class). 

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/tests/primitives_tests/arithmetic_tests/test_quantum_arithmetic.py#L190

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,143 @@
...
+from qrisp.jasp import jrange, jlen, DynamicQubitArray, check_for_tracing_mode
+
+
+def _session_of(b):
+    """Return b's QuantumSession in static mode, or None under tracing."""
+    if check_for_tracing_mode():
+        return None
+    # b may be a QuantumVariable, a list of Qubits, or a DQA.
+    first = b[0]
+    return first.qs()
```
purva-thakre: this helper function is defined as a temporary way to fix unit test failures associated with the `qcla_adder`. Everything should happen in the same `QuantumEnvironment` to avoid `KeyError: 'gidney_anc.0'`

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/src/qrisp/alg_primitives/arithmetic/adders/qcla/quantum_quantum/qq_qcla_adder.py#L64-L68

https://github.com/eclipse-qrisp/Qrisp/blob/1aada76a958637eb448a769aaeb4da24fec98bd8/src/qrisp/alg_primitives/arithmetic/adders/qcla/classical_quantum/cq_qcla_adder.py#L70-L75

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _qv(size, name):
+        if qs is not None:
+            return QuantumVariable(size, name=name, qs=qs)
+        return QuantumVariable(size, name=name)
+
+    def _qbool(name):
+        if qs is not None:
+            return QuantumBool(name=name, qs=qs)
+        return QuantumBool(name=name)
```
purva-thakre: same as https://github.com/eclipse-qrisp/Qrisp/pull/473/changes#r3085333378

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
purva-thakre: The helper functions were removed in https://github.com/eclipse-qrisp/Qrisp/pull/473/commits/3ef61c32179dc3c7bfd69859b7f2d0f17101fe09.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
purva-thakre: Removed this line in https://github.com/eclipse-qrisp/Qrisp/pull/473/commits/3ef61c32179dc3c7bfd69859b7f2d0f17101fe09. Led to the following failures (as expected).

```
FAILED jax_tests/test_montgomery.py::test_montgomery_not_jasp_qq - Exception: Not enough qubits to encode integer 15
FAILED jax_tests/test_montgomery.py::test_montgomery_find_order - Exception: Not enough qubits to encode integer 55
FAILED primitives_tests/arithmetic_tests/test_modular_arithmetic.py::test_modular_arithmetic - Exception: Not enough qubits to encode integer 40
```

purva-thakre:
@positr0nium This PR is pretty much ready except for the QCLA adder T depth/count regression issues. I do not know a lot about how the QCLA adder works. Any pointers here would be helpful if you can see the fault.

The T-depth and T count between the two versions of the Gidney adder used by the QCLA adder varies. The plot below only shows the data for QQ inputs; I see the same behavior with CQ input. 
Screenshot from 2026-04-27 15-04-04

The T-depth and T count between the old and new Gidney adder functions is the same. As expected, these values are much lower than that of the Cuccaro adder. 
Screenshot from 2026-04-27 15-03-34

I have tried refactoring the function with help of Opus as much as I could to look for the regression. One possible culprit pointed out by Opus is how c_in and c_out are handled between the two versions of the Gidney adder. I am going to take a look at the metrics related to these after a short break from this function 😅. 

Edit: Forgot to add that the T depth of the Gidney adder works identically to Fig 1 of the Gidney paper. There is a unit test for it too. For this reason, I think the fault is not in my new adder function, some optimization in the QCLA adder is being affected with the new function. 

https://github.com/purva-thakre/Qrisp/blob/d50831b74b35c5c2c9b2dd1af7d9ffad66ae81ee/tests/primitives_tests/arithmetic_tests/test_gidney_adder.py#L374

Screenshot from 2026-04-27 15-24-49

Edit: This issue was resolved. Raphael had a trick for reducing the T gate/depth when a carry in qubit was provided. Previous version was applying the mcx gate which increased the count.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,493 @@
...
+            # truncate classical a to the width of b so higher bits don't produce spurious
+            # gates (addition is mod 2^n)
+            if is_concrete_int:
+                a = int(a) % (1 << jlen(b))
+
+            # in JAX tracing mode, wrap the integer in a JAX array so
+            # subsequent bit-shift operations stay within the JAX graph
+            if check_for_tracing_mode():
+                if isinstance(a, int):
+                    a = jnp.array(a, dtype=jnp.int64)
```
purva-thakre: it might be better to move this check to the main `Adder` class. 

When the first input is not quantum and can represent a higher value than the second input, we can either end the in-place addition with overflow error or resort to modulo addition (which is what this code block ensures in the Gidney function)

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+                mcx([ctrl, b_qbs[0]], gidney_anc[0], method="gidney")
+            else:
+                cx(b_qbs[0], gidney_anc[0])
+    elif a_bits[0]:
+        if ctrl is not None:
+            mcx([ctrl, b_qbs[0]], gidney_anc[0], method="gidney")
+        else:
+            cx(b_qbs[0], gidney_anc[0])
+
+    if check_for_tracing_mode():
```
positr0nium: Why distinguish the logic between static and dynamic mode here? In my smoke tests the dynamic case works indeed for both cases.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+                    x(gidney_anc[i - 1])
+
+            mcx([gidney_anc[i - 1], b_qbs[i]], gidney_anc[i], method="gidney_inv")
+
+            with control(bit_i):
+                if ctrl is not None:
+                    cx(ctrl, gidney_anc[i - 1])
+                else:
+                    x(gidney_anc[i - 1])
+    else:
```
positr0nium: Same here

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+                else:
+                    x(gidney_anc[i - 1])
+
+    if check_for_tracing_mode():
+        with control(a_bit0):
+            if ctrl is not None:
+                mcx([ctrl, b_qbs[0]], gidney_anc[0], method="gidney_inv")
+            else:
+                cx(b_qbs[0], gidney_anc[0])
+    elif a_bits[0]:
```
positr0nium: Here there seems to be a problem because the control environment can not process ``None`` but I think this could also be resolved.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+    for j in jrange(n - 2):
+        i = j + 1
+        # Move carry dependency forward one position in the ripple.
+        cx(gidney_anc[i - 1], b_qbs[i])
+        cx(gidney_anc[i - 1], a_qbs[i])
+        mcx([a_qbs[i], b_qbs[i]], gidney_anc[i], method="gidney")
+        cx(gidney_anc[i - 1], gidney_anc[i])
+
+    if ctrl is not None:
+        mcx([ctrl, gidney_anc[n - 2]], b_qbs[n - 1])
```
positr0nium: I think one of these ``mcx`` could be ``method=gidney``. The following script shows that there is a growing number of non-gidney ``mcx`` with increasing ``n``. This can not be observed on the main branch.
```python
from qrisp import *

def c_adder(n):

    a = QuantumFloat(n)
    b = QuantumFloat(n)
    qbl = QuantumBool()

    with control(qbl):
        gidney_adder(a, b)

    return b

print(c_adder(10).qs.transpile(1).count_ops())
# {'uncompiled_gidney_mcx': 9, 'cx': 48, '2cx': 11, 'uncompiled_gidney_mcx_inv': 9}
```python

File: src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py (hunk 8 above)
Comment (purva-thakre) [reply src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py]: Mostly because the QCLA adder is not compatible in dynamic mode and I was having trouble reducing the T depth initially. I think you might be right that I do not need this.

File: src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py (hunk 11 above)
Comment (purva-thakre) [reply src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py]: Yeah, sorry. This should uniformly be gidney mcx. Looks like there were merge issues when I was cherry picking commits locally.

File: src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py [hunk 12]
```diff
@@ -0,0 +1,588 @@
...
+        Precomputed static bits of ``a_int`` in non-tracing mode.
+    a_int_is_bigint : bool
+        Indicates whether ``a_int`` must be read via ``get_bit``.
+    ctrl : Qubit or None
+        Optional outer control qubit.
+    """
+    if check_for_tracing_mode():
+        for i in jrange(n):
+            if c_in_qb is not None and i == 0:
+                continue
```
positr0nium: ``continue`` doesn't work with jrange. We should update the docs about this.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+    a_bits : list[int] or None
+        Precomputed static bits of ``a_int`` in non-tracing mode.
+    a_int_is_bigint : bool
+        Indicates whether ``a_int`` must be read via ``get_bit``.
+    ctrl : Qubit or None
+        Optional outer control qubit.
+    """
+    if check_for_tracing_mode():
+        for i in jrange(n):
+            if c_in_qb is not None and i == 0:
```
positr0nium: The ``i==0`` comparison should give a tracer, which makes me think that something here is dead code.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+
+    if c_in_qb is not None:
+        cx(c_in_qb, gidney_anc[0])
+        mcx([a_qbs[0], b_qbs[0]], gidney_anc[0], method="gidney_inv")
+
+        if ctrl is not None:
+            lsb_ctrl_anc = (
+                QuantumBool(name="gidney_lsb_ctrl_anc*")
+                if check_for_tracing_mode()
+                else QuantumBool(name="gidney_lsb_ctrl_anc*", qs=qs)
```
positr0nium: Setting the ``qs = qs`` keyword is an artifact from the time before we had jasp. The reasoning behind it was that we would accelerate the compiler speed by skipping certain checks and automations. By explicitely giving the QuantumSession, we could skip the QuantumSession merging procedure (this is realized through the ``fast_append`` context manager). Since we now have Jasp, compilation speed in static mode is no longer an issue because scalable compilation is supposed to happen in Jasp. Subsequently, we can drop these.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
```diff
@@ -0,0 +1,588 @@
...
+            + " Received invalid list for b; expected a non-empty list of qubits."
+        )
+
+    if isinstance(a, list) and not a_is_qubit_list:
+        raise ValueError(
+            invalid_input_pair_msg
+            + " Received invalid list for a; expected a list of qubits."
+        )
+
+    b_is_quantum = b_is_quantum_obj or b_is_qubit_list
```
positr0nium: Alot of this "bookkeeping" is most likely adder agnostic and could probably be moved to ``amend_inpl_adder`` (@bruno-ah-um 👀). Let's do it in a different PR though.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
purva-thakre: > Since we now have Jasp, compilation speed in static mode is no longer an issue because scalable compilation is supposed to happen in Jasp. Subsequently, we can drop these.

We cannot as the QCLA adder relies on this. A lot of the changes to the Gidney adder function are to make the QCLA adder happy. For example, 

https://github.com/eclipse-qrisp/Qrisp/blob/8fed21134f626a2090fe22f975bddb202005de3e/src/qrisp/alg_primitives/arithmetic/adders/qcla/quantum_quantum/qq_qcla_adder.py#L59

Do you mind if I make changes to the QCLA adder itself? many of the things you pointed out in your review could be easily fixed afterward.

tests/primitives_tests/arithmetic_tests/test_gidney_adder.py
```diff
@@ -0,0 +1,618 @@
+"""
```
positr0nium, line 1: Copilot says the following scenarios are not tested:

| Scenario | Covered? |
|---|---|
| Classical `a` + quantum `b` (static) | ✅ |
| Quantum `a` + quantum `b` (static) | ✅ |
| Classical `a` + `c_in` + `c_out` + `ctrl` (all 4, static) | ✅ |
| Quantum `a` + `c_in` + `c_out` + `ctrl` (all 4, static) | ✅ |
| **Classical `a` + `c_in` in dynamic/tracing mode** | ❌ **Missing** — triggers the `jrange`+`continue` bug |
| Classical `a` + `c_out` in dynamic mode | ❌ Missing |
| Quantum `a` + `c_out` in dynamic mode | ❌ Missing |
| String `a` in dynamic mode | ❌ Missing |
| `DynamicQubitArray` as `b` input | ❌ Missing (code supports it but not tested) |
| Negative / large `a` values in dynamic mode | ❌ Missing |
| `c_in` + `c_out` + `ctrl` all together, dynamic mode | ❌ Missing |
| Unequal-size `a` and `b` (static) | ✅ |
| Unequal-size `a` and `b` (dynamic) | ✅ |
| Ancilla cleanup verification (static) | ✅ |
| Ancilla cleanup verification (dynamic) | ❌ Missing |
| Raw `Qubit` carry wires (not `QuantumBool`) | ✅ |
| Invalid input validation (static) | ✅ |
| Invalid input validation (dynamic) | ✅ |
| Invalid binary string raises `ValueError` | ✅ |
| `BigInteger` classical input (dynamic) | ✅ |
| T-depth vs Cuccaro adder | ✅ |
| T-count formula with optional inputs | ✅ |
| Controlled addition on/off (static, both paths) | ✅ |
| Controlled addition on/off (dynamic, quantum `a` only) | ✅ |
| Single-qubit register boundary | ✅ |
| `numpy` integer compatibility | ✅ |
| Negative classical input wraps modulo | ✅ |
| List-of-qubits input | ✅ |

positr0nium (changes requested):

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
positr0nium: Yes, please go ahead :)

tests/primitives_tests/arithmetic_tests/test_gidney_adder.py
purva-thakre: > Ancilla cleanup verification (static) 	✅
> Ancilla cleanup verification (dynamic) 	❌ Missing

The static mode unit test for ancilla cleanup verification is not that well written, to be honest. Since the ancillas are defined inside the function and deleted before exiting the function, the only way I could think of was tracking the qubits in a QuantumSession.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
purva-thakre: Thanks! This was indeed dead code. 

I went ahead and added a page to the `Reference` section of the docs. See #598 

I still need to add some information about how to work with lists etc. in dyamic mode in addition to how indexing in dynamic mode works. Let me know if there is something worth adding to this page and I have not thought of it yet.

tests/primitives_tests/arithmetic_tests/test_gidney_adder.py
purva-thakre: I do not think I will be able to add some dynamic unit tests due to a possible bug as discussed in #608. 

Until this bug is fixed, the Gidney adder will fail to work in dynamic mode when `c_in` or `c_out` are provided as my code relies on concatenating these into the target register. If you can find the source of the bug, that's great; otherwise I will get to fixing it after the adder PRs. 

Edit: went ahead and added the unit tests but marked them as xfail.

src/qrisp/alg_primitives/arithmetic/adders/gidney_adder.py
purva-thakre: Yes, as discussed previously, that is the plan. For now, I am going to add a temporary function to validate the inputs to the Gidney which will then get replaced by the method of the `Adder` class.

positr0nium (approved):
Nice work, Purva!

PR #479, opened 2026-03-12
GQSVT

renezander90:
Implements Generalized Quantum Singular Value Transform (https://arxiv.org/pdf/2312.00723), algorithms based on (G)QSVT, and additional constructors for BlockEncodings.

**Algorithms**
- ``gqsp.GQSVT``
- ``gqsp.inversion`` now provides the option to choose the method (QET (only for Hermitian matrices), GQSVT).
- ``dalzell_inversion``Dalzell Quantum Linear System Solver: https://arxiv.org/pdf/2406.12086
- ``pseudo_inverse``Threshold pseudo inverse via GQSVT. 

**Constructors**
- ``from_projector``: Constructs a BlockEncoding for a projector, e.g., |1><3|.
- ``from_eye``: Constructs a BlockEncoding for ``np.eye``with support for parameter ``k``specifying the diagonal.

**Tests**

### 1. test_gqsvt.py
- **`test_gqsvt`** (parametrized with 4 cases)
  - Tests the GQSVT (Generalized Quantum Singular Value Transformation) algorithm on a small 4×4 matrix
  - Verifies polynomial transformations with different parities (odd/even)
  - Parameters tested: `[1., 1.]` (odd), `[1., 2., 1.]` (even), `[0.,1.,0.,1.]` (odd), `[2.,1.,0.,1.,2.,3.]` (odd)

- **`test_gqsvt_inversion`**
  - Tests GQSVT-based matrix inversion on a 4×4 matrix
  - Verifies that the quantum result matches the classical matrix inverse

### 2. test_pseudo_inversion.py
- **`test_pseudo_inversion`**
  - Tests GQSVT-based pseudo-inversion algorithm on a 4×4 matrix
  - Tests threshold-based inversion with specified epsilon (0.01) and theta (0.4) parameters
  - Includes post-selection on ancilla qubits being in |0⟩ state
  - Validates against numerically computed pseudo-inverse

### 3. test_dalzell_inversion.py
- **`test_dalzell_inversion`** (parametrized with 1 case)
  - Tests Dalzell's inversion algorithm on a 4×4 Hermitian matrix
  - Validates that success probability exceeds 50% threshold
  - Compares quantum results to classical matrix inverse

**Summary**: 4 test functions testing quantum matrix inversion algorithms (GQSVT, pseudo-inversion, and Dalzell's method) across the new QSP (Quantum Singular Value Transformation) module functionality.

### 4. test_block_encoding.py (modified)

**`test_block_encoding_from_eye`**
- Tests the `BlockEncoding.from_eye(k)` method
- Creates identity operators on different diagonals (k=0 for main diagonal, k=-4 for shifted diagonal)
- Tests tensor product of two eye block encodings
- Verifies correct quantum measurement outcomes

**`test_block_encoding_from_projector`**
- Tests the `BlockEncoding.from_projector(left_state, right_prep)` method  
- Creates projector P = |1⟩⟨+| where |+⟩ is created via Hadamard
- Tests that applying the projector to superposition state yields |1⟩ with probability 1.0

**`test_block_encoding_from_projector_value_error`**
- Tests error handling for `BlockEncoding.from_projector`
- Verifies that a `ValueError` is raised when left and right dimensions don't match
- Checks specific error message about size mismatch


MatP1337:
@renezander90  I think this PR is close to being ready for review and then merge,  would you agree?

documentation/source/reference/Algorithms/GQSP.rst
MatP1337, line 1: Should the non G and generalized approaches be separated? Also posed this question in https://github.com/eclipse-qrisp/Qrisp/pull/607

documentation/source/reference/Block Encodings/methods/from_eye.rst
MatP1337, line 1: Should we do the reformatting of moving the constructors in a separate folder for v0.9, or wait until v0.9.X? It wouldn't be a breaking change afaik

documentation/source/reference/Block Encodings/methods/from_projector.rst
MatP1337, line 1: same question as for ``.from_eye``

src/qrisp/algorithms/gqsp/dalzell_inversion.py
```diff
@@ -0,0 +1,264 @@
...
+import numpy.typing as npt
+from numpy.polynomial import Chebyshev
+from qrisp.algorithms.gqsp.gqsvt import GQSVT
+from qrisp.block_encodings import BlockEncoding
+from typing import Callable
+
+
+def dalzell_inversion(A: BlockEncoding, prep_b: Callable, t: float, eps: float, kappa: float) -> BlockEncoding:
+    r"""
+    Performs the `Dalzell quantum algorithm <https://arxiv.org/pdf/2406.12086>`_
```
MatP1337: I'd also include the seminar next to the paper since it shares the idea more clearly (at least for me). Link here: https://www.youtube.com/watch?v=OwqhdCioj4Y

src/qrisp/algorithms/gqsp/dalzell_inversion.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Parameters
+    ----------
+    A : BlockEncoding
+        The block-encoded matrix to be inverted. It is assumed that
+        the eigenvalues of $A/\alpha$ lie within $D_{\kappa} = [-1, -1/\kappa] \cup [1/\kappa, 1]$.
+    prep_b : Callable
+        A function ``prep_b(*operands)`` preparing the right hand side $\ket{b}$.
+    t : float
+        An estimate $t$ for the norm $\|x\|_2=\|(A/\alpha)^{-1}b\|_2$ for normalized $\|b\|_2=1$.
```
MatP1337, line 87: We should comment what happens when such an estimate is not available, and the resulting "loss of efficiency" of the approach.

src/qrisp/algorithms/gqsp/dalzell_inversion.py
```diff
@@ -0,0 +1,264 @@
...
+        An upper bound for the condition number $\kappa$ of $A$.
+        This value defines the "gap" around zero such that all sigular values (eigenvalues) of $A/\alpha$
+        lie in the domain $D_{\kappa}$.
+
+    Returns
+    -------
+    BlockEncoding
+        A new BlockEncoding instance representing the kernel reflection operator.
+
+    Notes
```
MatP1337, line 102: Maybe talk about comment on line R86 within the Notes and not in Parameters

src/qrisp/algorithms/gqsp/dalzell_inversion.py
```diff
@@ -0,0 +1,264 @@
...
+
+        @terminal_sampling
+        def main():
+            operand = operand_prep()
+            ancillas = BA_inv.apply(operand)
+            return operand, *ancillas
+
+        res_dict = main()
+
+        # Post-selection on ancillas being in |0> state
```
MatP1337, line 158: How about adding also the ``.apply_rus`` example without post-selection? I doubt it'd be long since we just reuse everything and just redefine ``main`` as ``main_rus``

src/qrisp/algorithms/gqsp/dalzell_inversion.py
```diff
@@ -0,0 +1,264 @@
...
+        lie in the domain $D_{\kappa}$.
+
+    Returns
+    -------
+    BlockEncoding
+        A new BlockEncoding instance representing the kernel reflection operator.
+
+    Notes
+    -----
+    - **Complexity**: The query complexitiy of the algorithm is determined by the polynomial degree and scales as $\mathcal O(\kappa\log(1/\epsilon))$.
```
MatP1337, line 104: Some further insights could be found in the following recent paper: https://arxiv.org/abs/2604.22185

I think we should link it, while inviting the users to try getting the QRE via ``.resources``

src/qrisp/algorithms/gqsp/dalzell_inversion.py
MatP1337, line 1: Really thorough walkthrough via the comments within the code. I appreciate that a lot!

src/qrisp/algorithms/gqsp/gqsvt.py
```diff
@@ -0,0 +1,196 @@
...
+from qrisp.algorithms.gqsp.gqsp import GQSP
+from qrisp.algorithms.gqsp.gqsp_angles import gqsp_angles
+from qrisp.algorithms.gqsp.helper_functions import poly2cheb, _rescale_poly
+from qrisp.block_encodings import BlockEncoding
+from typing import Literal, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from jax.typing import ArrayLike
+
+
```
MatP1337, line 31: remove line break

src/qrisp/algorithms/gqsp/gqsvt.py
```diff
@@ -0,0 +1,196 @@
...
+        # Reconstruct transformed matrix
+        A_poly = (U @ np.diag(S_poly) @ Vh).conj().T
+
+        res = A_poly @ b / np.linalg.norm(A_poly @ b)
+        print(res)
+        # [0.85184734, 0.47324852, 0.07098728, 0.21296184]
+
+    .. warning:: 
+
+        For non-Hermitian matrices performing Singular Value Transform
```
MatP1337, line 154: Would be insightful to also provide the argument why.

src/qrisp/algorithms/gqsp/helper_functions.py
```diff
@@ -174,3 +178,65 @@ def _rescale_poly(
...
+
+        tail_start = int(N * 0.8)
+        tail_sum = np.sum(np.abs(coeffs[tail_start:]))
+
+        if tail_sum < eps:
+            break
+
+        N *= 2
+
+    if N > max_N:
```
MatP1337, line 229: Just out of curiosity: how should the user address this? To put it more clearly: how should I proceed if I receive this error?

src/qrisp/algorithms/gqsp/inversion.py
```diff
@@ -17,31 +17,51 @@
...
 
 
-def inversion(A: BlockEncoding, eps: float, kappa: float) -> BlockEncoding:
+def inversion(A: BlockEncoding, eps: float, kappa: float, method: Literal["QET", "GQSVT"] = "QET") -> BlockEncoding:
     r"""
     Quantum Linear System Solver via Quantum Eigenvalue Transformation (QET).
     Returns a BlockEncoding approximating the matrix inversion of the operator.
 
-    For a block-encoded **Hermitian** matrix $A$ with normalization factor $\alpha$, this function returns a BlockEncoding of an
+    For a block-encoded matrix $A$ with normalization factor $\alpha$, this function returns a BlockEncoding of an
```
MatP1337: should we instead of removing "Hermitian" emphasize "not necessarily Hermitian", similarly to how we do it in the paper and somewhere else within this PR.

src/qrisp/algorithms/gqsp/inversion.py
```diff
@@ -119,20 +141,59 @@ def main():
...
+        raise ValueError(
+            f"Invalid method specified: '{method}'. "
+            f"Allowed methods are: {', '.join(ALLOWED_METHODS)}"
+        )
 
-    # Set _rescale=False to apply p(A/α) instead of p(A).
-    A_inv = QET(A, p, kind="Chebyshev", rescale=False)
+    p = _inversion_cheb(1.0 / kappa, eps)
+
+    if method == "GQSVT":
```
MatP1337, line 153: with QSVT in mind, what should be the order of if, elif, else here? Based on the problem instance the resources will differ among them. Perhaps we should make another if based on the Hermitianness of the system to apply A, B, or C.

src/qrisp/algorithms/gqsp/pseudo_inversion.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def pseudo_inversion(
+    A: BlockEncoding,
+    eps: float,
+    theta: float,
+    delta: float = None,
+) -> BlockEncoding:
+    r"""
+    Returns a BlockEncoding approximating the threshold matrix pseudo inversion of the operator.
```
MatP1337: would be useful to explain what a "pseudo inversion" is/means here.

src/qrisp/block_encodings/block_encoding.py
MatP1337, line 1: In this change, a lot of line breaks should be removed (:

MatP1337: From line 585 until line 644 and 738 until 801

tests/algorithms_tests/qsp_tests/test_dalzell_inversion.py
MatP1337, line 1: Thorough tests, nice!

MatP1337 (changes requested):
Some minor changes, especially when it comes to documentation. Code looks clean crispy.

Super excited seeing this in main! Dalzell linear solver intensifies 🔥

MatP1337:
@renezander90 it could be me not being to locate the test_qet.py file, but I am not able to locate the test_qet.py file 🥲🫠

Sharing just in case (:

src/qrisp/algorithms/gqsp/gqsvt.py
```diff
@@ -0,0 +1,196 @@
...
+
+    angles, new_alpha = gqsp_angles(p)
+
+    def new_unitary(*args):
+        if parity == "even":
+            x(args[1])
+        GQSP(args[0], *args[1:], unitary=BE_herm.unitary, angles=angles)
+        x(args[1]) # Ensure measuring ancilla in |0> yields correct result
+
+    new_anc_templates = [QuantumBool().template()] + BE_herm._anc_templates
```
positr0nium, line 189: ~~This creates stale ``QuantumBool`` instances, that are not cleaned up. We should maybe think about having a class method version of ``.template``~~

Edit: I'm stupid. I already implemented a much better solution myself to this some time ago so it should not be an issue.

src/qrisp/block_encodings/block_encoding.py
```diff
@@ -524,6 +524,282 @@ def unitary(*args):
             is_hermitian=is_hermitian,
         )
 
+    @classmethod
+    def from_eye(
+        cls: "BlockEncoding",
+        k: int = 0,
```
positr0nium: In the [numpy version](https://numpy.org/doc/stable/reference/generated/numpy.eye.html), the first two arguments of the function are used to specify the shaped of the matrix and the ``k`` argument only comes later. Obviously this doesn't apply here but it can be confusing when reading code which has just one argument. To resolve this, I think it would be better to not call it ``k`` but something more descriptive such as ``off_diagonal``.

positr0nium (approved):
The contents of this PR stretch far beyond my expertise - a review of the mathematical soundness would most likely require weeks of deep study. The UI is however minimal and well designed. If any major problems show up, I think we should be able to fix them without any breaking changes.

Massive Kudos to you, @renezander90. You most likely know it yourself but this is cutting edge quantum programming.

documentation/source/reference/Block Encodings/methods/from_eye.rst
renezander90: This will be done in a separate PR.

documentation/source/reference/Algorithms/GQSP.rst
renezander90: I would keep the whole QSP zoo in one place.

MatP1337: Agreed, but then let'S rename to qsp from gqsp?

src/qrisp/algorithms/gqsp/helper_functions.py
renezander90: Increase `max_N`:)

src/qrisp/algorithms/gqsp/inversion.py
renezander90: Yes, we set the keyword argument to `None` by default. In this case, we user `QET` if `is_hermitian=True` and `QSVT` if False. If a method is provided by the user, then we use this on. We can adjust this in the QSVT PR.

src/qrisp/algorithms/gqsp/pseudo_inversion.py
renezander90: Added a link https://en.wikipedia.org/wiki/Moore–Penrose_inverse

src/qrisp/algorithms/gqsp/gqsvt.py
renezander90: You're not stupid. You implemented the cleanup :)

MatP1337 (approved):
Most of the review comments were addressed. I think this PR is ready to merge to main.

Awesome, lfg!

PR #489, opened 2026-03-19
Refactoring + test implementations for the `QuantumCircuit` class (Part3)

PietropaoloFrisoni:
**Context** 

This is a *pure refactoring PR* for `QuantumCircuit` (the third one in the series). No new quantum features are introduced. The goals are better type safety, consistent internal state, cleaner import structure, and comprehensive test coverage.

---

**Bug Fixes**

- `abstract_params` was inconsistently typed

In `bind_parameters`, the result circuit's `abstract_params` was assigned `{}` (an empty `dict`) instead of `set()`. Everywhere else in the codebase it is a `set`. This was a latent bug: if abstract-parameter operations were ever appended to a circuit returned by `bind_parameters`, `dict.update()` would be called with a `set` of SymPy symbols and raise a `TypeError`. Now `abstract_params` is consistently `set()` throughout the entire lifecycle of a `QuantumCircuit`.

- `to_qasm2` silently discarded the transpiled circuit

In the fallback branch of `to_qasm2` (triggered when the circuit contained gates not representable in QASM 2.0), the code called `transpile(qiskit_qc, ...)` but then passed the *original, untranspiled* `qiskit_qc` to `dumps()`. The transpiled result was computed and immediately discarded. The method now correctly exports the transpiled circuit.

- `parity` accepted only `list` as multi-clbit input

The guard `not isinstance(clbits, list)` was narrow: passing a `tuple` or any other sequence would silently wrap it in a list of one element. The check is now `isinstance(clbits, Clbit)`, correctly handling all sequence types.

---

**Breaking changes**

- `measure`: single-qubit detection logic changed

This seems more a bug fix rather than a breaking change, but I flag this explicitly as such for visibility.
The old guard was `isinstance(qubits, (list, QuantumVariable))` to decide whether to allocate multiple classical bits. Any type not in that list (e.g. a `tuple` or `range`) was treated as a single-qubit input and would receive only one classical bit. The new guard checks `isinstance(qubits, (Qubit, int))` for the single-qubit path, so `tuple`, `range`, and other iterables are now correctly treated as multi-qubit inputs. Code that passed a `tuple` of qubits and expected a single `Clbit` to be allocated will now receive one `Clbit` per qubit.

---

**Other changes**

- **Imports**: lazy per-method imports of `qiskit.qasm2.dumps`, `qiskit.qasm3.dumps`, `qiskit_transpile`, and misc utilities (`cnot_count`, `get_depth_dic`, `t_depth_indicator`, `cnot_depth_indicator`) moved to module level. `TYPE_CHECKING`-guarded type aliases `QubitLike` and `ClbitLike` added.
- **`add_clbit`**: duplicate-identifier check is now skipped in XLA/Jasp accelerated mode (`xla_mode < 2`), consistent with how `add_qubit` behaves.
- **`count_ops`**: `try/except KeyError` replaced with `dict.get()`.
- **`compose`**: inplace path now has an explicit `return None`; return type annotated as `QuantumCircuit | None`.
- **`depth`**: return value wrapped in `int()` for a guaranteed `int` return type.
- **`__hash__`**: rewritten using comprehensions with inline comments explaining the algorithm; semantics unchanged.
- **Docstrings**: expanded with runnable examples for `get_depth_dic`, `cnot_count`, `cnot_depth`, `depth`, `t_depth`, `num_qubits`, `control`, `compose`, `run`, `measure`, `parity`, `to_qasm2`, `to_qasm3`, `statevector_array`, `transpile`.
- **Tests**: ~100 new test methods covering all refactored methods.

P.S. I left explicit comments in the PR : )


positr0nium (approved):

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -308,13 +327,14 @@ def add_clbit(self, clbit: Clbit | None = None) -> Clbit:
...
         if clbit is None:
-            clbit = Clbit("cb_" + str(self.clbit_index_counter[0]))
+            clbit = Clbit(f"cb_{self.clbit_index_counter[0]}")
 
         if not isinstance(clbit, Clbit):
             raise TypeError(f"Tried to add type {type(clbit)} as a classical bit")
 
-        if any(cb.identifier == clbit.identifier for cb in self.clbits):
-            raise ValueError(f"Clbit name {clbit.identifier} already exists")
+        if self.xla_mode < 2:
```
PietropaoloFrisoni, line 340: I am not entirely sure what is the purpose of this `self.xla_mode` check. I included it here because it appeared in the `add_qubit` method as well. If there is a reason why this should not be present in the `add_clbit` method, please let me know

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -789,27 +814,86 @@ def get_unitary(self, decimals: int | None = None) -> np.ndarray:
...
+        this QuantumCircuit.
+
+        The circuit is fully transpiled before counting, so that any composite gate
+        containing CX/CY/CZ gates is decomposed first.
 
         Returns
         -------
         int
-            The amount of CNOT gates.
+            The total number of CX, CY, and CZ gates after transpilation.
```
PietropaoloFrisoni, line 872: If it is not intentional that this method also counts `cy` and `cz` on top of `cx`, then this is a bug and we need to change the implementation of the `cnot_count` function

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -789,27 +814,86 @@ def get_unitary(self, decimals: int | None = None) -> np.ndarray:
...
+        QuantumCircuit.count_ops : Returns a full breakdown of every gate type
+            in the circuit.
+
+        """
 
         return cnot_count(self)
 
-    def transpile(self, transpilation_level=np.inf, **qiskit_kwargs):
+    def transpile(
+        self, transpilation_level: int | float = np.inf, **qiskit_kwargs
```
PietropaoloFrisoni, line 900: I am not enthusiast about using an interface-dependent value as default value of an optional argument (`numpy` in this case), but we might come back to this in the future (seems reasonable to avoid such changes for the time being).

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1595,47 +1834,71 @@ def append(self, operation_or_instruction, qubits=[], clbits=[]):
 
         self.data.append(Instruction(operation, qubits, clbits))
 
-    def run(self, shots=None, backend=None):
+    # TODO: Update after PR #331 is merged
+    def run(
```
PietropaoloFrisoni, line 1838: As mentioned in the docstring, this should be updated after the PR for the Backend refactoring is merged

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1696,57 +1961,82 @@ def statevector_array(self):
...
+            )
 
-        res *= len(self.qubits) ** 2
+            # Weight by (i+1)² so that swapping two instructions changes
+            # the total, making the hash order-sensitive.
+            total += hash((index_hash, param_hash, op_hash)) * (i + 1) ** 2
 
-        return hash(res)
+        # Scale by n² so that circuits with different qubit counts are
+        # unlikely to collide even when their instruction sequences match.
```
PietropaoloFrisoni, line 2035: Here I tried to explain what the code was already doing according to my understanding

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1025,137 +1166,214 @@ def bind_parameters(self, subs_dic):
 
             subs_circ.data.append(Instruction(op, ins.qubits, ins.clbits))
 
-        subs_circ.abstract_params = {}
+        subs_circ.abstract_params = set()
```
PietropaoloFrisoni, line 1169: I changed it to a `set` as the `abstract_params` attribute was initially created as such (so we do not change the type). Looking at the code, there should be no usage of this attribute that requires a `dict` instead of a `set`.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1025,137 +1166,214 @@ def bind_parameters(self, subs_dic):
...
-        except:
-            from qiskit.qasm2 import QASM2ExportError, dumps
-
-            try:
-                return dumps(qiskit_qc)
-            except (QASM2ExportError, TypeError):
-                from qiskit import transpile
-                from qiskit.qasm3 import dumps
-
-                transpiled_qiskit_qc = transpile(
```
PietropaoloFrisoni, line 1088: `transpiled_qiskit_qc` was dead code. I assumed that the intent was to use its output as argument of `dumps` below

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1025,137 +1166,214 @@ def bind_parameters(self, subs_dic):
...
+                    "s_dg",
+                    "t_dg",
+                    "cx",
+                    "cz",
+                    "rz",
+                ],
+            )
+            qasm_str = dumps_qasm2(transpiled_qiskit_qc)
+
+        if filename is not None:
```
PietropaoloFrisoni: I think this part was missing. If not, we can remove it (it is only used when `filename` is not `None`, which is the default

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1025,137 +1166,214 @@ def bind_parameters(self, subs_dic):
...
+        h qb_75[0];
+        cx qb_75[0], qb_76[0];
 
         """
+
         qiskit_qc = self.to_qiskit()
-        from qiskit.qasm3 import dumps
+        qasm_str = dumps_qasm3(qiskit_qc)
+
+        if filename is not None:
```
PietropaoloFrisoni, line 1318: Same as above

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1,32 +1,56 @@
...
+# * This Source Code may also be made available under the following Secondary
+# * Licenses when the conditions for such availability set forth in the Eclipse
+# * Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+# * with the GNU Classpath Exception which is
+# * available at https://www.gnu.org/software/classpath/license.html.
+# *
+# * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+# ********************************************************************************
+
+"""This module contains the QuantumCircuit class, which is the main class to describe quantum circuits in Qrisp."""
```
PietropaoloFrisoni, line 17: Here I changed by converting the huge string into a comment and adding a short string to explain the purpose of the file, as this seems to be [common in software development](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/missing-module-docstring.html), although if this is currently inconsistent with the rest of the ecosystem. We can also postpone this change for now.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -2017,65 +2323,120 @@ def to_cirq(self):
...
+        """
         if clbits is None:
-            from qrisp import QuantumVariable
 
-            if isinstance(qubits, (list, QuantumVariable)):
-                clbits = []
-                for i in range(len(qubits)):
-                    clbits.append(self.add_clbit())
-            else:
+            if isinstance(qubits, (Qubit, int)):
```
PietropaoloFrisoni, line 2386: With the previous version, any type which is not a `list` or `QuantumVariable` (e.g. a `tuple` or `range`) was treated as a single-qubit input and would receive only one classical bit. With the new guard, `tuple`, `range`, and other iterables are now correctly treated as multi-qubit inputs. Code that passed a tuple of qubits and expected a single Clbit to be allocated will now receive one Clbit per qubit.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -51,8 +75,6 @@ class QuantumCircuit:
         The amount of qubits this QuantumCircuit is initialized with. The default is 0.
     num_clbits : int, optional
         The amount of classical bits. The default is 0.
-    name : str, optional
```
PietropaoloFrisoni, line 54: This parameter was not used and removed from the signature in a previous PR. I forgot to update the docstring as well

src/qrisp/circuit/quantum_circuit.py
positr0nium: Let's postpone this for now. The problem that I see is that when we update the license docstring to "(c) 2027" via global notepad++ replacement, it will not change here.

src/qrisp/circuit/quantum_circuit.py
positr0nium: The point of the ``xla_mode`` feature is to accelerate the construction of very large circuits. I guess the reason why it was lacking is that during profiling the construction ``add_clbit`` did not show up as a bottleneck. But you are right, to be consistent the check should be there.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -789,27 +819,86 @@ def get_unitary(self, decimals: int | None = None) -> np.ndarray:
...
+        --------
+        QuantumCircuit.count_ops : Returns a full breakdown of every gate type
+            in the circuit.
+
+        """
 
         return cnot_count(self)
 
-    def transpile(self, transpilation_level=np.inf, **qiskit_kwargs):
+    def transpile(
```
positr0nium, line 901: At some point in the future I would like to deprecate the ``transpilation_level`` and ``transpilation_predicate`` keywords and have them under a new method called ``QuantumCircuit.decompose``.

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -1696,57 +1961,82 @@ def statevector_array(self):
...
+        n = len(self.qubits)
+        total = 0
+
+        for i, instr in enumerate(self.data):
+            # For each qubit the instruction acts on (in operand order),
+            # we record its position within self.qubits. Using
+            # circuit-global indices rather than Qubit object identities
+            # makes two structurally identical circuits hash the same even
+            # when their Qubit objects differ.
+            qubit_indices = tuple(self.qubits.index(qb) for qb in instr.qubits)
```
positr0nium: It is probably much more efficient to build up a dictionary of type ``{Qubit: int}`` before the loop instead of calling the ``list.index`` method, which performs linear search everytime. Could you quickly implement this change @PietropaoloFrisoni?

src/qrisp/circuit/quantum_circuit.py
```diff
@@ -2017,65 +2318,120 @@ def to_cirq(self):
...
+    # TODO: Extend to accept integer indices for clbits as well
+    def parity(
+        self,
+        clbits: Clbit | Sequence[Clbit],
+        expectation: int = 0,
+        observable: bool = False,
+    ) -> ParityHandle:
         """
-        Instructs a parity operation on a set of classical bits.
+        Append a parity (XOR) check over classical bits to the circuit.
```
positr0nium, line 2403: I noticed that this docstring was (slightly) inprecise even before your change.
The current behavior is that parity returns 
``p = b_0 ⊕ b_1 ⊕ … ⊕ b_{n-1} ⊕ expectation ``
such that it always returns ``0`` if the input value matches the expected parity (this is similar how it works in Stim).

positr0nium (changes requested):
Thanks for these efforts, @PietropaoloFrisoni :)

src/qrisp/circuit/quantum_circuit.py
PietropaoloFrisoni: This would make a lot of sense

src/qrisp/circuit/quantum_circuit.py
PietropaoloFrisoni: Ok, I will leave the description as part of the string which includes the ad

src/qrisp/circuit/quantum_circuit.py
PietropaoloFrisoni: Ok, I will leave it then

src/qrisp/circuit/quantum_circuit.py
PietropaoloFrisoni: Thanks, I added the `⊕ expectation ` part

positr0nium (approved):
Nice work!

PR #494, opened 2026-03-20
Fix `VirtualBackend` documentation for `shots` default.

tgehr:
The documentation currently states that a default is required for the `shots` parameter, but actually `None` will be passed, sometimes explicitly. This makes it clear that the `run_func` should accept a default of `None` and needs to handle it correctly.


positr0nium:
Danchä

Issue #312, opened 2025-11-12
Multi-control in `control` enviroment in jasp-mode


cjn10:
In the _non-jasp mode_, it is possible to use many qubits in the `control` enviroment:
```
from qrisp import *

def test():
    qf = QuantumFloat(3)
    qf[:] = 1
    with control(qf[0:2]):
        x(qf[2])
    return multi_measurement([qf])

test()
>>> {(1,): 1.0}
```
However, using `jaspify` this fails:
```
from qrisp import *
@jaspify
def test():
    qf = QuantumFloat(3)
    qf[:] = 1
    with control(qf[0:2]):
        x(qf[2])
    return measure(qf)

test()
>>> AttributeError: 'DynamicQubitArray' object has no attribute 'aval'
```

Issue #301, opened 2025-10-30
``count_ops`` gets stuck when QuantumModulus.decoder is called


positr0nium:
The following snippet reproduces the issue:

```python
from qrisp import *

@count_ops(meas_behavior = "1")
def main():
    a = BigInteger.create_static(13, size=4)
    b = BigInteger.create_static(1, size=4)
    qm = QuantumModulus(a)
    # This works
    return b

    # This gets stuck
    return qm.decoder(b)

print(main())
```
The actual jax execution concludes but it becomes stuck in some C function call when trying to import the results into Python. I think this is due to some component within the decoder.

ACE07-Sev:
They have to comment first before they can be assigned.

cjn10:
Hi, sorry for the late reply. I am already working on this but I don't think the error has anything to do with the decoder or the quantum modulus. I will investigate further.

positr0nium:
Yes, I think it is probably some "exotic" Jax operation inside one of the BigInteger arithmetic functions

PR #495, opened 2026-03-20
QuantumModulus and BigInteger tweaks

### LINKED ISSUE #328: Bug: initialization of QuantumModulus with BigInteger in static mode
The following snippet:

```
from qrisp import *

p = BigInteger.create(13, size = 1)
qm = QuantumModulus(p)
```
yields:

`TypeError: loop of ufunc does not support argument 0 of type BigInteger which has no callable log2 method
`

### LINKED ISSUE #418: Quantum-Quantum modular multiplication between QuantumModuli

### LINKED ISSUE #469: qq_montgomery_multiply_modulus does not work with BigInteger moduli
## Description

`qq_montgomery_multiply_modulus` currently does not support `BigInteger` moduli for quantum-quantum multiplication of `QuantumModulus` values.

The function computes the reduction shift `m` using:
```python
n = int(np.ceil(np.log2(int(N))))
m = int(np.ceil(np.log2((int(N) - 1) ** 2) + 1)) - n
```

This breaks when `N` is a `BigInteger` because `int(N)` and `np.log2` do not handle the `BigInteger` type directly. Additionally, the partial-product accumulation loop and Montgomery reduction would need to operate on `BigInteger`-width registers.

## Current behavior

Attempting to multiply two `QuantumModulus` instances backed by `BigInteger` moduli raises an error or produces incorrect results in JASP mode.

## Expected behavior

`qq_montgomery_multiply_modulus` should detect `BigInteger` moduli and dispatch to an appropriate BigInteger-aware multiplication path, similar to how `cq_montgomery_multiply_inplace` already handles BigInteger inputs.

## Context

This was identified during the fix in PR #403 (`fix-qq-montgomery-multiply`), which resolved a related issue with arbitrary Montgomery shifts for standard (non-BigInteger) moduli. The BigInteger case was left as a known limitation.

### LINKED ISSUE #492: Bug: out-of-range basis states in QuantumModulus cause NaN measurements and simulator crash
## Bug description

`QuantumModulus(p)` allocates `p.bit_length()` qubits, supporting `2^n` basis states, but only values `0..p-1` are valid. When a gate like `h()` creates a superposition over **all** basis states, the out-of-range states (`p..2^n-1`) cause two problems:

1. **NaN in measurement results** — the decoder cannot map these values, so they appear as `nan`
2. **Simulator crash for larger primes** — when kaliski or other algorithms operate on a superposition containing invalid states, the simulator hits a `SystemError` in the numba sparse-matrix kernel

### Minimal reproducer

```python
from qrisp import *

v = QuantumModulus(5)
h(v)
print(v)
# {0: 0.125, 1: 0.125, 2: 0.125, 3: 0.125, 4: 0.125, nan: 0.125}
```

For `p=11` with kaliski, this leads to a hard crash:
```
SystemError: <function _numba_unpickle at 0x...> returned a result with an exception set
```

### Expected behaviour

`QuantumModulus(p)` should either:
- Prevent operations that create out-of-range states, or
- Correctly handle / display all `2^n` basis states (e.g., reduce mod p or show raw values)
- Adjust the probability of having a nan.

diehoq:
This PR started as a correctness fix for qq_montgomery_multiply_modulus to handle arbitrary Montgomery shifts, but it also pulls in the follow-up work needed to make some part of QuantumModulus arithmetic robust under tracing, cached operands, decoded measurements, and BigInteger-based modular arithmetic. In practice, the branch now covers three themes: Montgomery-shift correctness, QuantumModulus runtime robustness, and BigInteger compatibility fixes.

## Commit Breakdown

- `d727974b` fixes `qq_montgomery_multiply_modulus` so it works with arbitrary Montgomery shifts instead of assuming both operands share the same shift. It also makes the result shift consistent with the static Montgomery multiplication semantics.
- `a6300212` fixes the return type in `cq_montgomery_multiply`, so the Montgomery multiplication path returns a `QuantumModulus` instead of the wrong quantum type.
- `26ee5498` adds a regression test for `qq_montgomery_multiply_modulus` with standard-form inputs, covering the failure mode that originally motivated the branch.

- `e4420bd8` fixes `QuantumModulus.jdecoder` for BigInteger moduli and for the `m == 0` case, avoiding broken decoding paths when the modulus is represented as a `BigInteger`.

- `97c8e186` fixes decoded measurement aggregation so probabilities are accumulated when multiple raw basis states decode to the same classical value, instead of later states overwriting earlier ones. This changes the reported probability outcomes and is especially relevant for `QuantumModulus` states that decode to the same value or to `nan`.
- `68f393f2` makes `QuantumModulus` arithmetic robust when cached traced operands flow into arithmetic methods by reducing non-`QuantumModulus` / non-`QuantumFloat` operands modulo the current modulus before arithmetic.
- `60623f3e` resets recycled qubits after delete verification, so qubits returned to the free pool are actually clean after a reset/delete flow.
- `ed71868d` fixes recursive quantum-session discovery so `QuantumVariable` instances are treated as leaves instead of being iterated into qubits, which was breaking under tracing.

- `b9b62ae6` introduces the first BigInteger compatibility pass for quantum modular arithmetic. It adds BigInteger-safe modulus comparisons, BigInteger-aware modulus handling in Montgomery code, `int % BigInteger` support, BigInteger-aware `ctrl_state` handling in Balauca MCX, and safer `Jaspr.copy()` constant handling.
- `a536dcd8` adds a safer BigInteger coercion path for modular operations, so large Python integers are converted without relying on unsafe host-int assumptions.
- `c39d188d` fixes BigInteger `ctrl_state` handling and ensures `measure_to_big_integer` stores limbs with the correct unsigned type.
- `86c8e59a` adds a focused regression test for `measure_to_big_integer`.
- `0454de8f` completes the BigInteger Montgomery / `QuantumModulus` fixes by making modulus sizing, operand coercion, and Montgomery multiplication work correctly with BigInteger moduli.
- `2f1b09b0` fixes the controlled-Jaspr signature by keeping `constvars` internal instead of exposing them as runtime arguments. This resolves argument-count mismatches in controlled JASP paths that involve constants.

- `7aeeb69c` adds regression coverage for the QuantumModulus uniform-superposition / invalid-basis-state measurement behavior.
- `d9b9e551` renames the temporary size placeholder in the QuantumModulus constructor for clarity without changing behavior.
- `5b4895ca` adds docstrings, comments, and clearer exceptions around the newer Montgomery-decoding helpers and related control-transform plumbing.
- `2b6b313c` adds traced BigInteger compatibility for power-of-two modular arithmetic via `bi_pow2mod`, and rewrites the QuantumModulus decode path so BigInteger-backed decoding no longer depends on forcing a concrete modulus value on the host.
- `9dfc2f46` removes the earlier hotfix from `jasp_montgomery` once the helper-based traced decode path is in place.
- `c0447564` tightens `_moduli_neq` by making it raise when used on traced moduli, so callers must guard that static-only check explicitly instead of using it accidentally under tracing.
- `89fd570f` fixes BigInteger operand coercion so modular arithmetic returns operands with the correct limb width, and updates the associated docstrings.
- `6b3c5148` adds focused JAX regression tests for the newer BigInteger, decoding, and comparison behavior in `QuantumModulus`.
- `ce1c03f8` refines `_moduli_neq`, clarifies the comparison exceptions, and expands the tests around mismatched moduli and Montgomery shifts.
- `e993d528` removes a remaining `int()` cast from Montgomery-shift computation so the shift logic stays compatible with traced values.
- `ecdf0742` starts the Montgomery-decoding refactor by introducing helper logic in `jasp_mod_tools` to express decoding in terms of the shift exponent.
- `75d89f39` completes that refactor by moving `QuantumModulus` decode usage onto the new helper path and adding targeted decode tests.
- `ed5d6638` follows up with docstring and formatting cleanup for the new Montgomery decode helpers.
- `5589faf8` restructures the `QuantumModulus` constructor so size computation and default-adder selection are separated more clearly between tracing and non-tracing paths.
- `f662daaf` adds a reduction helper in `QuantumModulus` so scalar modular operands are normalized consistently before arithmetic, with focused tests for that path.
- `3e8a7dbf` lightens the decoding call logic by routing `QuantumModulus` decoding through `new_montgomery_decoder` while preserving the public `decoder` behavior that maps invalid basis states to `nan` during measurement labeling.

## Issues Tackled

- [#418 Quantum-Quantum modular multiplication between QuantumModuli](https://github.com/eclipse-qrisp/Qrisp/issues/418)  
  This is addressed by `d727974b`, with follow-up support from `a6300212` and regression coverage in `26ee5498`. The main point is that `qq_montgomery_multiply_modulus` now behaves correctly even when inputs are not already in the same Montgomery form. The PR fixes #418.

- [#469 qq_montgomery_multiply_modulus does not work with BigInteger moduli](https://github.com/eclipse-qrisp/Qrisp/issues/469)  
  This is addressed primarily by `0454de8f`, with supporting groundwork in `b9b62ae6` and `a536dcd8`. Together these commits make the Montgomery multiplication path BigInteger-aware. The PR fixes #469.

- [#328 Bug: initialization of QuantumModulus with BigInteger in static mode](https://github.com/eclipse-qrisp/Qrisp/issues/328)  
  This is also addressed by `0454de8f`, because the fix replaces the `np.log2(modulus)`-style sizing path with logic that works for BigInteger moduli. The PR fixes #328.

- [#492 Bug: out-of-range basis states in QuantumModulus cause NaN measurements and simulator crash](https://github.com/eclipse-qrisp/Qrisp/issues/492)  
  The probability-accounting side of this issue is addressed by `97c8e186`. That commit changes decoded measurement handling so probabilities are accumulated instead of overwritten when several raw basis states decode to the same outcome. In practice, this changes the reported probability distribution, including the probability mass assigned to invalid / `nan`-decoded outcomes. Thi PR fixes #492.

- [#301 `count_ops` gets stuck when QuantumModulus.decoder is called](https://github.com/eclipse-qrisp/Qrisp/issues/301)  
  This is likely addressed by `e4420bd8` and #505 , since that commit fixes the BigInteger decoding path inside `QuantumModulus`.

- [#412 Bug: controlled/inverted state preparation with method='qswitch'](https://github.com/eclipse-qrisp/Qrisp/issues/412)  
  This incidently by `2f1b09b0`, with additional supporting groundwork in `b9b62ae6`. The issue is an argument/constant-handling mismatch in controlled JASP path. Altough not a fix.

- [#312 Multi-control in `control` environment in jasp-mode](https://github.com/eclipse-qrisp/Qrisp/issues/312)  
  This PR likely improves the general controlled-JASP path via `2f1b09b0`. Although not a fix.

The remaining commits are mostly robustness and compatibility work that came up while fixing the Montgomery and BigInteger arithmetic issues, even when they do not map one-to-one to a currently open issue.


src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_bigintiger.py
```diff
@@ -205,6 +205,21 @@ def create_dynamic(n, size):
         """
         return BigInteger.create(n, size)
 
+    @staticmethod
+    def coerce(n, size):
+        """
+        Coerce ``n`` into a fixed-width BigInteger of ``size`` limbs.
```
renezander90, line 211: If `n` is a BigInteger, the method returns `n`. Hence, it is not guaranteed that `n` will have `size` limbs in this case. This should be mentioned in the doc string, or (if useful?) addressed by creating a new BigInteger with `size` limbs. If `n` has more than `size` limbs, an error could be thrown or the last limbs could be cut off.

Please also add ``Parameters``and ``Returns`` to the doc string.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
```diff
@@ -109,6 +109,9 @@ def montgomery_decoder(
             N if isinstance(N, BigInteger) else BigInteger.create(N, Rb.digits.shape[0])
         )
         return bi_montgomery_decode(yb, Rb, Nb)
+    # Handle fractional R (from negative Montgomery shifts)
+    if isinstance(R, float) and 0 < R < 1:
```
renezander90, line 160: This is inconsistent with the typing ` R: Union[int, BigInteger]`. Why can this method receive a float as Montgomery radix? Shouldn't `R` be always an integer?

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
```diff
@@ -326,15 +356,25 @@ def qq_montgomery_multiply_modulus(x: QuantumModulus, y: QuantumModulus):
...
 
-    # res = qq_montgomery_multiply(x, y, N, m, inpl_adder)
-    # return res
+    # Compute the reduction shift from the modulus, exactly like montgomery_mod_mul:
+    #   m = ceil(log2((N-1)^2 + 1)) - n
+    # This is the number of extra bits needed so that the full integer product
+    # a*b fits in (n + m) bits, independent of the inputs' Montgomery shifts.
+    # NOTE: We compute n from N (not from x.size) to keep it a plain Python int,
+    # since x.size can be a JAX tracer during tracing.
+    N_int = _concrete_modulus_to_int(N, "QuantumModulus modulus")
```
renezander90:
```
from qrisp import *
from jax.core import Tracer

@jaspify
def main(N):
    
    digits = N.digits
    if isinstance(digits, Tracer):
        print("digits is a tracer")

    a = QuantumModulus(N)

N = BigInteger.create_static(1213, 1)
main(N)
```
> digits is a tracer

Hence, `qq_montgomery_multiply_modulus` will not work in this situation. Why do we have to coerce to static int? We could work with dynamic values?

This should always work? Or do we need the Montgomery shift to be a static int?

```
from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_mod_tools import smallest_power_of_two

n = smallest_power_of_two(N)
m = smallest_power_of_two((N-1) ** 2 + 1) - n
```

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
```diff
@@ -33,6 +36,27 @@
 )
 
 
+def _concrete_modulus_to_int(value, name: str) -> int:
```
renezander90: Why is there the `name` parameter?
I am not convinced yet that we need this function (see comment below).

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -26,6 +27,51 @@
 import jax
 
 
+def _moduli_neq(a, b):
+    """Compare two moduli for inequality, keeping BigInteger constants concrete."""
+    from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_bigintiger import BigInteger
+
+    if isinstance(a, BigInteger) or isinstance(b, BigInteger):
+        if not isinstance(a, BigInteger) or not isinstance(b, BigInteger):
```
renezander90: Why should this method always return `True` if we compare a BigInteger to an int?

```
N = BigInteger.create_static(5, 1)
if N!=5:
    print("N is not 5")
else:
    print("N is 5")
```
>N is 5

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -26,6 +27,51 @@
...
+            return True
+
+        if isinstance(a.digits, Tracer) or isinstance(b.digits, Tracer):
```
renezander90, line 62: It is actually possible to have dynamic moduli:

```
from qrisp import *
from jax.core import Tracer

@jaspify
def main(N):
    
    digits = N.digits
    if isinstance(digits, Tracer):
        print("digits is a tracer")

N = BigInteger.create_static(5, 1)
main(N)
```
> digits is a tracer

If I am not mistaken `_moduli_neq` is used whenever there are checks in other methods to ensure that the moduli a compatible. These checks do not work if the moduli are dynamic. I would simply return `False` which deactivates the check in this case, and not throw a warning because this could lead to excessive warnings output. 

In many situations we have checks that only work in static mode and not in dynamic mode.

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -201,8 +247,12 @@ def __init__(self, modulus, inpl_adder=None, qs=None):
             self.m = 0
 
         else:
+            from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_mod_tools import (
+                smallest_power_of_two,
+            )
+
             self.modulus = modulus
-            self.m = int(np.ceil(np.log2(modulus)))
```
renezander90, line 205: We could use this opportunity to remove some confusion:

Here, we first set `self.m = int(smallest_power_of_two(modulus))`,
then initialize `QuantumFloat.__init__(self, msize=self.m, qs=qs)`,
and then set `self.m = 0`(line 266). 
That is, we use the `self.m` attribute as temporary placeholder for the size of the variable,
and then set the Montgomery shift, the true purpose of `self.m`, to 0.

That is, we should not call it `self.m` here and in the next line.

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -239,18 +289,42 @@ def decoder(self, i):
...
 
-        return montgomery_decoder(i, 2**self.m, self.modulus)
+        if i_value >= modulus_value:  # or (np.gcd(i, self.modulus) != 1 and i != 0):
+            return np.nan
+        decoded = montgomery_decoder(i_value, 2**self.m, modulus_value)
+        if isinstance(self.modulus, BigInteger):
+            return BigInteger.create_static(decoded, self.modulus.digits.shape[0])
+        return decoded
 
+    def jdecoder(self, i):
```
renezander90, line 345: @positr0nium  what is actually the difference between `decoder` and `jdecoder`?
Now, they would behave differently in some cases, i.e., in the `decoder`, if the value is a BigInteger the shift must be positive. In the `jdecoder`, a negative shift would be allowed.

I think, we can save some code if we pass the Montgomery shift `m` or the inverse of the Montgomery radix modulo N `R^{-1} mod N` instead of the Montgomery radix `R` to the `montgomery_decoder`.  In particular, we should not calculate the radix as `2 ** m` for `m<0` which leads to `float` values. We have:

If `m<0` this means `R=(2^m)^{-1} mod N`, hence `R^{-1} = 2^m mod N`.
If `m>=0` this means `R=(2^m) mod N`, hence `R^{-1} = (2^m)^{-1} mod N`.

In both cases: `R^{-1} mod N = pow(2, -m, N)` as recognized below.

This could also be cleaned up in a separate PR.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
renezander90: I now see how `R` could be a float if `m<0`. 
But, we should avoid this behavior:
https://github.com/eclipse-qrisp/Qrisp/pull/495/changes#r3029425670

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -411,7 +492,11 @@ def __imul__(self, other):
     def __add__(self, other):
         if isinstance(other, int):
             other = self.encoder(other % self.modulus)
-        elif isinstance(other, QuantumModulus):
+        elif isinstance(other, (QuantumModulus, QuantumFloat)):
```
renezander90, line 547: I don't understand the changes here and below. Could you explain?

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
```diff
@@ -214,18 +217,18 @@ def smallest_power_of_two(n: Union[int, BigInteger]):
...
         )
-    else:
-        # Pure Python int path, exact and safe
-        if hasattr(n, "bit_length"):
-            return (n - 1).bit_length() if n > 1 else 0
-        # Fallback (should not occur in practice)
-        return 0
+
+    # Fallback (should not occur in practice)
+    return 0
```
positr0nium: What exactly does "should not occur in practice" means? Do we need an exception?

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
positr0nium: It might be helpful to link the paper here

positr0nium, line 88: Probably more efficient to do this via ``cyclic_shift`` (not linear depth) but this is a performance question and not immediately pressing.

src/qrisp/jasp/interpreter_tools/interpreters/cl_func_interpreter.py
```diff
@@ -977,7 +977,10 @@ def loop_body(
 
         qubit = qubit_reg.pop()
         # Verify the qubit is in |0⟩ state before deletion
-        cond(get_bit_array(bit_array, qubit), true_fun, false_fun)
+        dirty = get_bit_array(bit_array, qubit)
+        cond(dirty, true_fun, false_fun)
+        # Reset dirty bits to |0⟩ so recycled qubits are clean
```
renezander90, line 982: A dirty qubit would indicate an uncomputation error and result run in a warning ("WARNING: Faulty uncomputation found during simulation."). 
Here, we flip such a dirty qubit, but this does not solve the true problem which could be an error in the code that is simulated? 

However, there are situations where this warning could be thrown without a meaningful error: e.g., 1) the qubit is properly unentangled, but is still in the |1> state; 2) the qubits was measured. Then, either an `x` or `reset` would be missing to avoid the warning:).
Maybe a `reset_and_delete` method makes sense.

Regardless, I think **the change here**, i.e., flipping such a dirty qubit to |0>, **is useful**

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
positr0nium: I think the problem was indeed the logarithm not being available on big integer, but it is indeed solved by ``smallest_power_of_two``

src/qrisp/jasp/jasp_expression/centerclass.py
```diff
@@ -247,6 +247,10 @@ def copy(self):
             effects=self.effects,
             debug_info=self.debug_info,
         )
+        if self.consts:
+            kwargs["consts"] = list(self.consts)
```
renezander90: What is the difference between `constvars` and  `consts`?

src/qrisp/jasp/jasp_expression/control_transform.py
```diff
@@ -317,7 +317,7 @@ def multi_control_jaspr(jaspr, num_ctrl, ctrl_state):
     ctrl_avals = [x.aval for x in ctrl_vars]
 
     return make_jaspr(exec_multi_controlled_jaspr(jaspr, num_ctrl, ctrl_state))(
-        *(ctrl_avals + [var.aval for var in jaspr.invars[:-1] + jaspr.constvars])
+        *(ctrl_avals + [var.aval for var in jaspr.invars[:-1]])
```
positr0nium, line 325: What was the reasoning for this change?

src/qrisp/qtypes/quantum_modulus.py
positr0nium: Alternatively, the function could also just raise an Exception if called within dynamic mode. This make clear that such checks are specifically designed to be executed in static mode and skipped in dynamic mode.

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -26,6 +27,51 @@
...
+
+def _coerce_bigint_operand(value, modulus):
+    """Coerce *value* to a BigInteger matching *modulus*'s limb count.
+
+    Delegates to :meth:`BigInteger.coerce` to avoid duplicating the
+    int-vs-traced dispatch logic.
+    """
+    from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_bigintiger import BigInteger
+
+    if isinstance(value, BigInteger):
```
positr0nium: The docstring says that the function returns a BigInteger containing ``modulus``'s limb count but just returning ``value`` if it is BigInteger doesn't guarantee this.

src/qrisp/qtypes/quantum_modulus.py
positr0nium: It should be possible to compare non-zero Montgomery shifts, if the shifts agree, or?

src/qrisp/qtypes/quantum_modulus.py
positr0nium: I'd go even further and rename the attribute ``.m`` to something more telling. But maybe not in this PR.

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -239,18 +289,42 @@ def decoder(self, i):
...
+        )
 
-        if i >= self.modulus:  # or (np.gcd(i, self.modulus) != 1 and i != 0):
-            return np.nan
-        return montgomery_decoder(i, 2**self.m, self.modulus)
-
-    def jdecoder(self, i):
-        from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_mod_tools import (
-            montgomery_decoder,
+        i_value = i() if isinstance(i, BigInteger) else int(i)
```
positr0nium, line 333: It would be good to have some comments here, explaining what these things do.

src/qrisp/qtypes/quantum_modulus.py
positr0nium: The split is entirely arbitrary and I already considered fusing them such that there is only ``decoder``, which distinguishes between tracing mode and non-tracing mode. However with the upcoming plans to introduce type checks, I'm reconsidering this because in some cases these functions have different signatures (``QuantumVariable.decoder`` returns bitstrings, ``QuantumVariable.jdecoder`` returns integers because jax has no dynamic bitstring type)

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -302,22 +376,29 @@ def encoder(self, i):
                 return montgomery_encoder(i, 1 << self.m, self.modulus)
 
         else:
+            from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_bigintiger import (
+                BigInteger,
+            )
+            from qrisp.alg_primitives.arithmetic.modular_arithmetic import (
```
positr0nium, line 434: These in-function imports are getting really annoying... We should start to disentangle the inter-module interrelationship

src/qrisp/qtypes/quantum_modulus.py
positr0nium: There also seems to be a lot of duplicated code. Maybe we can perform this type based logic through some decorator.

positr0nium (changes requested):
Thanks for these fixes, they are important work towards realizing a stable system of jasp-compatible modular arithmetic. Like many files, I think QuantumModulus needs some refactoring but this is of course only the next step before the fundamental issues are fixed. So nice work, Diego! :)

renezander90 (changes requested):
- Some comments to be addressed.
- Additional test cases and mapping between issues and test cases should be provided.

diehoq:
Thanks a lot for the reviews guys, I will address the comments rn!

src/qrisp/jasp/jasp_expression/control_transform.py
diehoq: This came up in the context of the controlled addition over EC, but is of course more general than that:
inside the multi-control wrapper, Qrisp eventually calls` controlled_jaspr.embedd(...)` at control_transform.py (line 354), which in turn evaluates the inner Jaspr expecting the stored consts and the runtime invars, according to abstract_interpreter.py (line 168). 
Now, while building the controlled wrapper, the code traced it as if the captured constants were public arguments, thus effectively changing the expected signature of the function from `controlled(ctrls..., x)` to `controlled(ctrls..., x, k)`. But `control_jaspr(jaspr) `already preserves the original constvars and consts internally at control_transform.py (line 280).
The multi-control wrapper should only add control qubits to the public signature and not add the hidden constant and ask the caller to pass it, right?

This snippet:
```
import jax.numpy as jnp
from qrisp import QuantumVariable, p
from qrisp.jasp import make_jaspr

phase = jnp.array(0.5)

def f():
    qv = QuantumVariable(1)
    p(phase, qv[0])
    return qv

jaspr = make_jaspr(f)()
controlled = jaspr.control(2, "11")
```
raised
```
ValueError: Tried to evaluate jaxpr with insufficient arguments: expected 3 (including 1 consts), got 4.
```
before the fix.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
diehoq: This? https://arxiv.org/abs/1801.01081

src/qrisp/qtypes/quantum_modulus.py
renezander90: This could be considered a breaking change since `.m` is exposed in the documentation: https://www.qrisp.eu/reference/Quantum%20Types/QuantumModulus.html

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
diehoq: I agree that the typing is inconsistent. However this handling of float R was already implemented. https://github.com/eclipse-qrisp/Qrisp/blob/ee8da70ee15e5bc041a1b1b6a595af12d2a1d9ce/src/qrisp/alg_primitives/arithmetic/modular_arithmetic/mod_tools.py#L24

You are suggesting to switch to your solution already in this PR?

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
diehoq: Yes, I meant that all the cases should already be covered by one of the branches. I added an exception to handle unsupported types.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_mod_tools.py
renezander90: No this should be a separate PR.

Yes, I am aware that float R has been used. This will require changes at several places.

src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_montgomery.py
```diff
@@ -353,7 +372,10 @@ def qc_mul_inplace(operand, cl_int):
                 inpl_adder(cl_int // 2, operand[size - 1 - i :])
 
     res = QuantumModulus(N)
-    res.m = m
+    # The result's Montgomery shift after reduction: (x.m + y.m) - m
+    # (the reduction divides by 2^m, subtracting m from the accumulated shift)
+    res.m = int(x.m) + int(y.m) - m
```
renezander90, line 377: I would assume that the `int()`conversion will cause problems when the shifts are tracers/BigInteger. Could we just use: `res.m = x.m + y.m - m`?

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -226,31 +310,73 @@ def decoder(self, i):
...
+            # For m <= 0 →  decode factor = 2^|m| mod N  (identity when m == 0)
+            factor = lax.cond(
+                m > 0,
+                lambda p: bi_modinv(p, self.modulus),
+                lambda p: p,
+                power,
+            )
+            return bi_montgomery_encode(i, factor, self.modulus)
+
+        # Concrete path for int modulus
```
renezander90, line 372: This will not work if the shift is a tracer (not sure how relevant this is):

```
from qrisp import *

@jaspify
def main():

    qf = QuantumFloat(4)
    qf[:] = 7
    N = jnp.int32(measure(qf))

    qf2 = QuantumFloat(2)
    qf2[:] = 2
    m = jnp.int32(measure(qf2))

    qm = QuantumModulus(N)
    qm[:] = 3
    qm.m = m
    return measure(qm)

main()
```
> ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected:

We should also move this decoding logic to (an updated version of) the previously used `montgomery_decoder` which handles the different cases (static, dynamic, BigInteger). 
(Should be a separate PR.)

renezander90: More importantly, this will also not work if the shift is a non-zero int, but the modulus is a tracer:

```
from qrisp import *

@jaspify
def main():

    qf = QuantumFloat(4)
    qf[:] = 7
    N = jnp.int32(measure(qf))

    qm = QuantumModulus(N)
    qm[:] = 3
    qm.m = 1
    return measure(qm)

main()
```
> ConcretizationTypeError

renezander90: We could use a `pow2mod` function that distinguishes between BigInteger, tracing, non-tracing, as the `modinv` function does. And something like this:

```
def new_montgomery_decoder(i, m, modulus):
    if check_for_tracing():
        abs_m = jnp.abs(m)
        power = pow2mod(abs_m, modulus)
        factor = lax.cond(
            m > 0,
            lambda p: modinv(p, modulus),
            lambda p: p,
            power,
        )
    else:
        abs_m = abs(m)
        power = pow2mod(abs_m, modulus)
        factor = modinv(power, modulus) if m>0 else power

    montgomery_encoder(i, factor, modulus)
```

src/qrisp/qtypes/quantum_modulus.py
diehoq: So, the change was needed in order to deal with some cached/traced wrapper produced by the JASP/qache machinery. In the old code, such an operand would miss all three explicit branches and go straight into mod_adder unclassified and unnormalized.

I'll add some comments and look into making this a decorator.

src/qrisp/qtypes/quantum_modulus.py
```diff
@@ -26,27 +27,112 @@
...
 
-            if self.modulus != other.modulus:
+            if _moduli_neq(self.modulus, other.modulus):
                 raise Exception(
                     "Tried to compare QuantumModulus instances of differing modulus"
                 )
+        elif not check_for_tracing_mode() and self.m != 0:
+            # Comparing against a non-QuantumModulus (e.g. QuantumFloat)
+            # requires standard representation (m == 0).
+            raise Exception(
```
renezander90: Should be ValueError or RuntimeError rather than Exception?

renezander90 (approved):
Thank you! We made good progress here regarding reliability and structure of the QuantumModulus, with still more work for future PR's :).

src/qrisp/qtypes/quantum_modulus.py
positr0nium: This is a bug 😇

positr0nium (changes requested):
Only one minor change left :)

src/qrisp/qtypes/quantum_modulus.py
diehoq: Fixed

positr0nium:
Thanks, Diego - really nice work! :)

PR #497, opened 2026-03-29
GQSP angles upgrades

renezander90:
This PR addresses performance issues when calculating GQSP angles for high degree polynomials:

- Use `jax.lax.scan` primitive instead of static for loop in `_inlft` function: this results in constant compilation time while otherwise the XLA computation graph contains the unrolled loop.
- Use `@jax.jit`decorated post processing function `_gqsp_angles_from_nlft_sequence`

As a result, calculating angles for a degree 100_000 polynomials takes <1s (and used to result in a crash due to memory):

```
from qrisp import *
from qrisp.gqsp import gqsp_angles

angles, alpha = gqsp_angles(np.ones(100_000))
```

Note: `poly2cheb`and `cheb2poly` functions remain bottlenecks if the polynomials must to be converted between the two bases (this is not necessary for, e.g., `inversion`).

### DIFF
--- a/src/qrisp/algorithms/gqsp/gqsp_angles.py
+++ b/src/qrisp/algorithms/gqsp/gqsp_angles.py
@@ -154,7 +154,7 @@ def _complementary_polynomial(b: "ArrayLike") -> Array:
 @jax.jit
 def _inlft(a: "ArrayLike", b: "ArrayLike") -> Array:
     r"""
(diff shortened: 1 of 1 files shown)


MatP1337 (approved):
Looks good, seems to work...

Looking forward to now perform resource estimation for larger scales!

Thanks for the contribution, you da real MVP!

PR #505, opened 2026-03-30
Metric compilation upgrades

positr0nium:
# PR Summary: Port `pure_callback` Anti-Inlining to Profiling Interpreters

## Description

Ports the `jax.pure_callback` anti-inlining optimization from `boolean_simulation` (`cl_func_interpreter.py`, merged via #483) to the profiling interpreters used by `count_ops`, `depth`, and `num_qubits`. Also adds a `callback_threshold` slider to all four Jasp backends giving users control over the compile-speed vs. execution-speed tradeoff.

**Problem**: Without this optimization, XLA's `flatten-call-graph` pass inlines the HLO of reused `@qache`'d subroutines at every call site, causing superlinear compilation time (dozens of minutes for large programs like Shor's algorithm). The boolean simulation interpreter already had this fix; the profiling interpreter was missing it.

**Solution**: Before evaluating a reused sub-jaxpr (one with `call_count > 1`), the profiling interpreter checks if `call_count × inlined_eqn_count` exceeds a threshold. If so, the sub-jaxpr is called via `jax.pure_callback`, which makes the JIT-compiled HLO opaque to the parent compilation — XLA can no longer duplicate it.

The `callback_threshold` parameter exposes this as a user-controlled slider:

| Value | Behavior |
|-------|----------|
| `None` (default) | No callbacks at all — fastest execution, slowest compile |
| `0` | Every reused sub-jaxpr wrapped — fastest compilation, slowest execution |
| `500` | Suggested middle ground for large algorithms |
| `10^9` | Practically disables callbacks |

## Related Issues

Closes the performance regression where 128-bit Shor QRE took ~7s on `main` but ~95s with hardcoded callbacks (reported by @renezander90). Enables ECDLP dynamic pipeline (requested by @diehoq).

Related to #505.

## Type of Change

- [x] Feature (new functionality) — `pure_callback` anti-inlining for profiling interpreters
- [x] Feature (new functionality) — `callback_threshold` parameter for all four Jasp backends
- [x] Performance improvement — drastically reduced XLA compile time for large programs with reused subroutines
- [x] Refactoring — `get_compiled_profiler` cache changed from `@lru_cache` to `OrderedDict`-based LRU

## Breaking Change?
- [x] No

`callback_threshold` defaults to `None` (no callbacks), preserving pre-PR behavior for existing code.

---

### Core optimization (`profiling_interpreter.py`)

The heart of the PR. Several new components were added:

#### `_should_use_profiling_callback(jaxpr, call_graph_stats, callback_threshold)` → bool
Decides whether a sub-jaxpr should be wrapped in `jax.pure_callback`. Returns `True` when:
- `call_graph_stats` is not `None` (callbacks enabled)
- The sub-jaxpr appears in the call graph stats
- `call_count > 1` (the sub-jaxpr is reused)
- `call_count × inlined_eqn_count >= callback_threshold`

The `AbstractQuantumState` check from the boolean simulator is intentionally dropped because profiling-transformed jaxprs no longer carry quantum types.

#### `_get_result_shapes(jaxpr_evaluator, invalues)` → pytree of `ShapeDtypeStruct`
The main design challenge. `jax.pure_callback` requires a pytree of `ShapeDtypeStruct` describing the output shape/dtype *before* the callback executes. The raw sub-jaxpr outvars carry quantum abstract types (`AbstractQuantumState`, `AbstractQubitArray`) that have no `.shape`/`.dtype`. Instead, we trace the **profiling-transformed** evaluator via `jax.make_jaxpr(return_shape=True)`, which produces `ShapeDtypeStruct` for the classical metric-state outputs. Results are cached by evaluator `id()`.

#### `get_compiled_profiler()` — refactored
- Cache changed from `@lru_cache` to an `OrderedDict`-based LRU (max 10k entries) because `call_graph_stats` is an unhashable `dict`.
- Now returns `(profiler, jaxpr_evaluator)` — the profiler is always JIT-compiled (critical: when inside `pure_callback`, JAX calls it with concrete arrays outside the trace, so the JIT is eager and the compiled HLO stays local).
- Accepts `call_graph_stats` and `callback_threshold` parameters.

#### `make_profiling_eqn_evaluator()` — extended
- Accepts `call_graph_stats` and `callback_threshold`.
- In the `jit` primitive handler: calls `_should_use_profiling_callback()` and, when `True`, wraps the profiler call in `jax.pure_callback(profiler, result_shapes, *invalues)` instead of calling `profiler(*invalues)` directly.

### Metric modules — call graph analysis integration

All three metric modules (`count_ops_metric.py`, `depth_metric.py`, `num_qubits_metric.py`) were updated to:
1. Import and call `analyze_call_graph(jaspr)` to get `call_graph_stats`.
2. Pass `call_graph_stats` to `make_profiling_eqn_evaluator()`.

This is the minimal hook needed to enable the callback optimization.

### Public API (`profiler.py`, `boolean_simulation.py`)

- **`count_ops(meas_behavior, callback_threshold=None)`** — new `callback_threshold` kwarg.
- **`depth(meas_behavior, max_qubits=1024, callback_threshold=None)`** — new `callback_threshold` kwarg.
- **`num_qubits(meas_behavior, max_allocations=1000, callback_threshold=None)`** — new `callback_threshold` kwarg.
- **`boolean_simulation(func, bit_array_padding=2**16, callback_threshold=None)`** — new `callback_threshold` kwarg.

All public functions have updated docstrings explaining the parameter.

### `centerclass.py`

`Jaspr.count_ops()`, `.depth()`, `.num_qubits()` methods accept and thread `callback_threshold`.

### `cl_func_interpreter.py`

The boolean simulation's `_should_use_callback` was refactored to accept `callback_threshold` instead of hardcoding `500`.

---

#### Existing tests (basic correctness)
All four test files each had one `test_callback_threshold_*` test verifying correctness across threshold values.

#### New tests added (9 total)

**`test_count_ops.py`** — 3 new:

| Test | Coverage |
|------|----------|
| `test_callback_threshold_nested_qache` | Nested `@qache` (outer calls inner, both reused) — thresholds 0, 500 + `meas_behavior="1"` |
| `test_callback_threshold_with_jrange` | `jrange` loop (20 iters) with qache'd subroutine — scan-primitive path, thresholds 0, 500 |
| `test_callback_threshold_edge_cases` | `threshold=1`, determinism (3 calls), single-use subroutine (no reuse) |

**`test_depth.py`** — 2 new:

| Test | Coverage |
|------|----------|
| `test_callback_threshold_depth_nested_qache` | Nested qache depth metric — thresholds 0, 500, 1 |
| `test_callback_threshold_depth_with_jrange` | `jrange` loop (15 iters) depth — thresholds 0, 500 |

**`test_boolean_simulation.py`** — 2 new:

| Test | Coverage |
|------|----------|
| `test_callback_threshold_boolean_simulation_nested` | Nested qache boolean sim — thresholds 0, 500 |
| `test_callback_threshold_boolean_simulation_jrange` | `jrange` loop (30 iters) — thresholds 0, 500, 1 |

**`test_num_qubits.py`** — 2 new + import fix (`cx`, `x` added):

| Test | Coverage |
|------|----------|
| `test_callback_threshold_num_qubits_nested` | Nested qache with qubit alloc/dealloc — thresholds 0, 500, 1 |
| `test_callback_threshold_num_qubits_jrange` | `jrange` loop (10 iters) with qubit allocation — thresholds 0, 500 |

---

### `id()`-based caching vs `@lru_cache`

The `@lru_cache` decorator was replaced with an `OrderedDict`-based LRU cache because `call_graph_stats` (a `dict`) is unhashable and must be threaded through the caching layer. The cache is keyed by `(id(jaxpr), metric_cls, cache_key)`.

A review comment from @PietropaoloFrisoni raised the concern that `id()` reuse after GC eviction could theoretically cause a cache collision. This would require >10k distinct `@qache`'d subroutines — so many that entries are evicted, the old jaxpr is GC'd, and a new one coincidentally gets the same memory address. The probability is astronomically low, and the only fully correct alternative (deep Jaxpr equality comparison) would be prohibitively slow.

### `jax.make_jaxpr(return_shape=True)` for result shapes

The `_get_result_shapes` function traces the profiling-transformed evaluator with `jax.make_jaxpr(return_shape=True)`. This is the only way to get output shapes because the raw sub-jaxpr outvars carry quantum abstract types without `.shape`/`.dtype`. This tracing is performed at most once per evaluator (cached by `id`). The cost is amortized across all call sites of that sub-jaxpr.

### Compile-time vs. run-time tradeoff

Callback wrapping prevents XLA from inlining HLO, which speeds up compilation but slows down execution (extra kernel launch overhead). The `callback_threshold` slider lets the user decide where on this spectrum they want to be:

- For small programs: leave at `None` (fastest execution).
- For very large programs (Shor, ECDLP): `0` or `500` to prevent multi-minute compilation.
- Default `None` preserves backward compatibility.

---

## How was it tested?

Every test follows the same pattern:
1. Compute baseline with default `callback_threshold=None`.
2. Compute with several threshold values (`0`, `500`, `1`, `10^9`).
3. Assert identical results — callback wrapping must never change the semantic outcome.

Test coverage across four scenarios per metric (basic, nested `@qache`, `jrange` loop, edge cases):

| Test area | Status |
|-----------|--------|
| `count_ops` — basic correctness | ✅ |
| `count_ops` — nested qache (new) | ✅ |
| `count_ops` — jrange loop (new) | ✅ |
| `count_ops` — edge cases (new) | ✅ |
| `depth` — basic correctness | ✅ |
| `depth` — nested qache (new) | ✅ |
| `depth` — jrange loop (new) | ✅ |
| `boolean_simulation` — basic correctness | ✅ |
| `boolean_simulation` — nested qache (new) | ✅ |
| `boolean_simulation` — jrange loop (new) | ✅ |
| `num_qubits` — basic correctness | ✅ |
| `num_qubits` — nested qache (new) | ✅ |
| `num_qubits` — jrange loop (new) | ✅ |


positr0nium:
This is the drawback of this technique. It trades compile speed for execution speed. Maybe we can add some kind of "slider" that allows the user to go from one extrem to the other.

PietropaoloFrisoni:
Thanks @positr0nium , I will take a look tomorrow 🚀

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -278,47 +278,139 @@ def get_handlers(self) -> Dict[str, Callable[..., Any]]:
         }
 
 
-# This reconstructs the metric inside the cached function so caching not keyed
-# by the metric object identity.
-@lru_cache(int(1e5))
```
PietropaoloFrisoni, line 283: `lru_cache` should hold a strong reference to a `jaxpr` (so as long as the cache entry exists, `jaxpr` cannot be garbage collected). Now we are caching using `(id(jaxpr), ...)` in `_profiler_cache` and with `id(jaxpr_evaluator)` in `_result_shapes_cache`. I am trying to understand if it is somehow possible that the following happens:

- `jaxpr_evaluator_A` (built from `jaxpr_A`) is stored in `_profiler_cache` and its result shapes are stored in `_result_shapes_cache` using `id(jaxpr_evaluator_A)` as key.

- 10k other entries are added to `_profiler_cache`, evicting the entry above. `jaxpr_evaluator_A` has no more references and is garbage collected. Its memory address is freed.

- Later, `jaxpr_evaluator_B` (built from a different `jaxpr_B`) is created and Python happens to reuse the same memory address (which is used by `id` in CPython). Now `id(jaxpr_evaluator_B) == id(jaxpr_evaluator_A)`.

- `_get_result_shapes(jaxpr_evaluator_B, invalues)` finds a cache hit and returns shapes computed for `jaxpr_A` (which would be the wrong shapes for `jaxpr_B`).

But I don't think this can ever happen because, if I understand correctly, this would be the case with a quantum program with more than 10k distinct `@qached` subroutines. Do you agree?

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -278,47 +278,139 @@ def get_handlers(self) -> Dict[str, Callable[..., Any]]:
...
+# ---------------------------------------------------------------------------
+# Callback helper – decides whether a sub-jaxpr should be wrapped in
+# ``jax.pure_callback`` to prevent XLA from inlining it at every call site.
+# ---------------------------------------------------------------------------
+
+def _should_use_profiling_callback(jaxpr, call_graph_stats):
```
PietropaoloFrisoni, line 286:
```suggestion
def _should_use_profiling_callback(jaxpr : Jaxpr | ClosedJaxpr, call_graph_stats: dict | None) -> bool:
```

Actually, these are internal functions and the type is already specified in the docstring, so I think that the type hint is not very relevant here (we might decide on a policy for such things and write it in the internal doc guidelines)

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -278,47 +278,139 @@ def get_handlers(self) -> Dict[str, Callable[..., Any]]:
...
+    """
+    Decide whether *jaxpr* should be called via ``jax.pure_callback``.
+
+    A sub-jaxpr benefits from callback wrapping when it is **reused**
+    (``call_count >= 2``) and large enough to matter
+    (``call_count * inlined_eqn_count >= 500``).  Wrapping prevents XLA's
```
PietropaoloFrisoni, line 292: I am curious about why these were chosen as the parameters (which are the same as in `_should_use_callback` used by the boolean simulator).  When you proposed to add some kind of 'slider' to go from one extreme to the other, were you referring to parametrizing such parameters?

PietropaoloFrisoni (commented):
Extremely instructive stuff, thanks a lot @positr0nium !

I am not sure I understand all the implications of this, so I am very curious to ask a few questions later on 🚀

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -34,11 +34,11 @@
 
 
 from abc import ABC, abstractmethod
-from functools import lru_cache
+from collections import OrderedDict
 from typing import Any, Callable, Dict, Sequence, Tuple
 
 import jax
-import numpy as np
+from jax import pure_callback, ShapeDtypeStruct
```
PietropaoloFrisoni, line 41: Is `ShapeDtypeStruct` used?

renezander90:
> This is the drawback of this technique. It trades compile speed for execution speed. Maybe we can add some kind of "slider" that allows the user to go from one extrem to the other.

We should have this slider feature. Otherwise resource estimation will be significantly slower in many other situations.

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
positr0nium: This can only happen if the ID's indeed match, which probably less likely than winning the lottery. Things like this can happen with any hashing mechanism and the only truly solid solution would be to compare for equality by comparing equation by equation. This would make everything really slow.

diehoq:
This should be completed and merged if we want to merge the ECDLP implementation into main, otherwise the dynamic pipeline would not work.

renezander90:
> This should be completed and merged if we want to merge the ECDLP implementation into main, otherwise the dynamic pipeline would not work.

Do you want to introduce the ECDLP still with 0.9? I am still concerned that this PR will slow down resource estimation in other cases. We should retest for Shor's algorithm.

diehoq:
It's what I suggested some weeks ago. However it's not an hard requirement and it would be nice if some tutorial came with it, so I guess we can postpone it to the next release.

positr0nium:
Implemented the slider and added some tests for it.

src/qrisp/jasp/evaluation_tools/boolean_simulation.py
```diff
@@ -64,7 +64,7 @@
 from qrisp.jasp import Jaspr
 
 
-def boolean_simulation(*func: Callable, bit_array_padding: int = 2**16) -> Callable:
+def boolean_simulation(*func: Callable, bit_array_padding: int = 2**16, callback_threshold=None) -> Callable:
```
PietropaoloFrisoni:
```suggestion
def boolean_simulation(
    *func: Callable,
    bit_array_padding: int = 2**16,
    callback_threshold: int | None = None,
) -> Callable:
```

Just type hint + formatting

src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -111,7 +111,7 @@ def _normalize_meas_behavior(meas_behavior: str | Callable) -> Callable:
 # Keeping them here for now to avoid circular imports.
 
 
-def count_ops(meas_behavior: str | Callable) -> Callable:
+def count_ops(meas_behavior: str | Callable, callback_threshold=None) -> Callable:
```
PietropaoloFrisoni:
```suggestion
def count_ops(
    meas_behavior: str | Callable, callback_threshold: int | None = None
) -> Callable:
```

src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -278,15 +285,15 @@ def ops_counter(*args):
...
+                *args, meas_behavior=meas_behavior, callback_threshold=callback_threshold
             )
 
         return ops_counter
 
     return count_ops_decorator
 
 
-def depth(meas_behavior: str | Callable, max_qubits: int = 1024) -> Callable:
+def depth(meas_behavior: str | Callable, max_qubits: int = 1024, callback_threshold=None) -> Callable:
```
PietropaoloFrisoni:
```suggestion
def depth(
    meas_behavior: str | Callable,
    max_qubits: int = 1024,
    callback_threshold: int | None = None,
) -> Callable:
```

src/qrisp/jasp/evaluation_tools/profiler.py
```diff
@@ -418,15 +433,15 @@ def depth_counter(*args):
...
+                *args, meas_behavior=meas_behavior, max_qubits=max_qubits, callback_threshold=callback_threshold
             )
 
         return depth_counter
 
     return depth_decorator
 
 
-def num_qubits(meas_behavior: str | Callable, max_allocations: int = 1000) -> Callable:
+def num_qubits(meas_behavior: str | Callable, max_allocations: int = 1000, callback_threshold=None) -> Callable:
```
PietropaoloFrisoni:
```suggestion
def num_qubits(
    meas_behavior: str | Callable,
    max_allocations: int = 1000,
    callback_threshold: int | None = None,
) -> Callable:
```

src/qrisp/jasp/interpreter_tools/interpreters/cl_func_interpreter.py
```diff
@@ -86,7 +86,7 @@
 BooleanQuantumState = tuple[BitArray, Jlist]  # (bit_array, free_qubits)
 
 
-def make_cl_func_eqn_evaluator(call_graph_stats=None):
+def make_cl_func_eqn_evaluator(call_graph_stats=None, callback_threshold=None):
```
PietropaoloFrisoni:
```suggestion
def make_cl_func_eqn_evaluator(
    call_graph_stats: dict | None = None,
    callback_threshold: int | None = None,
) -> Callable:
```

src/qrisp/jasp/evaluation_tools/boolean_simulation.py
```diff
@@ -97,6 +97,14 @@ def boolean_simulation(*func: Callable, bit_array_padding: int = 2**16) -> Calla
         and will always be ``1/64`` of the ``bit_array_padding``.
         The default here is therefore ``2**16/2**6 = 1024``.
 
+    callback_threshold : int or None, optional
+        For very large algorithms, compile time can blow up due to aggressively
+        inlining of the Jax pipeline. ``callback_threshold`` allows to mitigate
```
PietropaoloFrisoni:
```suggestion
        For very large algorithms, compile time can blow up due to aggressive
        inlining of the Jax pipeline. ``callback_threshold`` allows to mitigate
```

I don't trust my English skills, but according to Gemini: 

```
"aggressively"` is an adverb modifying "inlining" (used as a noun), which is ungrammatical
```

(same for all the other cases)

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -278,47 +278,151 @@ def get_handlers(self) -> Dict[str, Callable[..., Any]]:
...
-    Returns:
-        callable: A profiler function that takes the same arguments as the Jaxpr
-            and returns the evaluated result with metric information.
+    Returns
+    -------
+    tuple[Callable, Callable]
+        ``(profiler, jaxpr_evaluator)`` – the compiled profiler function and the
+        raw evaluator (used for computing result shapes for ``pure_callback``).
     """
+    key = (id(jaxpr), metric_cls, cache_key)
```
PietropaoloFrisoni:
```suggestion
    key = (id(jaxpr), metric_cls, cache_key, callback_threshold)
```

I did not try to run the tests with this, but if I understand correctly, with 2 calls with the same circuit but different thresholds, without this the returned the stale cached profiler is only compiled with the first threshold. Am I missing something? I am still trying to figure out if this is indeed the case (I will continue the review after the stand-up)

src/qrisp/jasp/interpreter_tools/interpreters/profiling_interpreter.py
```diff
@@ -329,6 +433,16 @@ def make_profiling_eqn_evaluator(metric: BaseMetric) -> Callable:
...
 
+    call_graph_stats : dict[int, JaxprStats] | None, optional
+        Call graph analysis results from ``analyze_call_graph``. When provided,
+        enables callback-based compilation optimization for reused sub-jaxprs.
+
+    callback_threshold : int | None, optional
+        Minimum value of ``call_count * inlined_eqn_count`` required to trigger
+        ``jax.pure_callback`` wrapping.  ``None`` (default) disables callbacks
+        entirely (fastest execution).  ``0`` wraps every reused sub-jaxpr
+        (fastest compilation).  The PR's original heuristic was ``500``.
```
PietropaoloFrisoni:
```suggestion
        (fastest compilation). 
```

src/qrisp/jasp/interpreter_tools/interpreters/cl_func_interpreter.py
```diff
@@ -590,7 +594,7 @@ def no_warning():
     context_dic[eqn.outvars[0]] = jnp.array(result, dtype=bool)
 
 
-def process_while(eqn: JaxprEqn, context_dic: ContextDict, call_graph_stats=None) -> bool | None:
+def process_while(eqn: JaxprEqn, context_dic: ContextDict, call_graph_stats=None, callback_threshold=None) -> bool | None:
```
PietropaoloFrisoni:
```suggestion
def process_while(
    eqn: JaxprEqn,
    context_dic: ContextDict,
    call_graph_stats: dict | None = None,
    callback_threshold: int | None = None,
) -> bool | None:
```

src/qrisp/jasp/interpreter_tools/interpreters/num_qubits_metric.py
```diff
@@ -278,7 +279,7 @@ def extract_num_qubits(res: Tuple, jaspr: Jaspr, _) -> dict:
 
 @lru_cache(int(1e5))
 def get_num_qubits_profiler(
-    jaspr: Jaspr, meas_behavior: Callable, max_allocations: int = 1000
+    jaspr: Jaspr, meas_behavior: Callable, max_allocations: int = 1000, callback_threshold=None
```
PietropaoloFrisoni:
```suggestion
def get_num_qubits_profiler(
    jaspr: Jaspr,
    meas_behavior: Callable,
    max_allocations: int = 1000,
    callback_threshold: int | None = None,
) -> tuple[Callable, None]:
```

src/qrisp/jasp/interpreter_tools/interpreters/num_qubits_metric.py
```diff
@@ -278,7 +279,7 @@ def extract_num_qubits(res: Tuple, jaspr: Jaspr, _) -> dict:
 
 @lru_cache(int(1e5))
 def get_num_qubits_profiler(
```
PietropaoloFrisoni, line 281:
```suggestion
```

PietropaoloFrisoni (commented):
Great stuff! 

I will continue the review after the stand-up (I still have some questions, but they are mostly for my own understanding).

I suggest a few minor changes, mostly cosmetic

PietropaoloFrisoni (approved):
🚀

PR #518, opened 2026-04-03
Sofware engineering - Issue Templates

bastibock:
## Description

This PR adds Issue templates for:
- Feature requests
- Change requests
- Bug reports
- Documentation
- Epics

## Type of Change
- Adds GitHub Functionality

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?
With these changes users can directly select one of the templates to fill in the according issue.

## How was it tested?
Unfortunately there is no testing possible, since GitHub only allows this feature if the Yaml templates are in main.
There are however no breaking changes to the actual Qrisp Code.

## Checklist
- [ ] I have added/updated tests (referencing issue Test-IDs) (not applicable)
- [ ] I have updated the documentation (not applicable)
- [ ] I have added a changelog entry (not applicable)
- [ ] Breaking changes are documented with migration path (not applicable)


.github/issue_template/bug_report.yml
PietropaoloFrisoni, line 1: These templates are placed under `.github/issue_template/`, but, looking at the  [GitHub documentation on issue and pull request templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), the correct path seems to be `.github/ISSUE_TEMPLATE/` (plural, uppercase).

I am not 100% sure, but with the current path, GitHub would probably silently ignore the templates, so no form will appear.

.github/issue_template/epic.yml
PietropaoloFrisoni, line 1: Is it intentional that there is no `priority` field here?

.github/issue_template/epic.yml
```diff
@@ -0,0 +1,67 @@
...
+    id: dependencies
+    attributes:
+      label: Dependencies & Order
+      description: "Describe dependencies between sub-issues. Use a diagram if helpful."
+    validations:
+      required: false
+
+  - type: textarea
+
+    id: done
```
PietropaoloFrisoni, line 71: Is it intentional that here there is `done` as `id` instead of `definition-of-done` as in the other workflows?

.github/issue_template/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+      required: false
+
+  - type: textarea
+
+    id: tests
+    attributes:
+      label: Means of Verification
+      description: "For example 1 positive, 1 negative, and 2 boundary tests."
+      value: |
+        | Test-ID | Type     | Description | Expected Result |
```
PietropaoloFrisoni, line 53: Compared to the other tables, this does not have an `Input` column. I don't know if this is intentional or not, but I am not sure I understand what the purpose of the `Input` field should be. Can you please explain? Thanks!

PietropaoloFrisoni (commented):
Thank you so much @bastibock , I am happy to see that there is some work going on in this direction! 

In general, I think these templates are an important contribution and a component that we definitely need 🚀 

I may miss some context, as I was not present in the last meeting, so I will just comment on what I see in this PR. It seems to me that these templates describe what to test in prose, but they say nothing about where the tests should live, how they should be written, or which framework to use. Someone new to the project, reading these templates, has no idea whether to write `pytest` tests, `unittest`, `doctest`, or something else entirely.

To circumvent this problem, we can probably write a `CONTRIBUTING.md` or a testing guide, and maybe add a reference in this template (this was probably already your plan).

.github/issue_template/bug_report.yml
bastibock: Then we should change it to make sure, GitHub accepts the templates.

.github/issue_template/epic.yml
bastibock: No, this is not intentional. I will add it, thank you for making me aware!

.github/issue_template/epic.yml
bastibock: This is actually somewhat intentional. But thinking about this again, it would make sense to have this here as well. We can also discuss this on Thursday.

bastibock:
Hi Pietro,
Thank you very much for checking the templates and adding your comments!
These templates are meant to help with gathering and having an overview over the requirements to go to a more planned and structured development of Qrisp.

So this is not only about testing but adding new features, changes etc in Qrisp in general.
But the testing part was also the biggest discussion last Thursday and I agree that we should definitely add some guide/ section for this in the Conributing.md file.

.github/issue_template/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+      description: "Tests to verify the fix. At least 1 positive, 1 negative, and 2 boundary tests."
+      value: |
+        | Test-ID | Type     | Description | Input | Expected Result |
+        |---------|----------|-------------|-------|-----------------|
+        | T-001   | Positive |             |       |                 |
+        | T-002   | Negative |             |       |                 |
+        | T-003   | Boundary |             |       |                 |
+        | T-004   | Boundary |             |       |                 |
+    validations:
+      required: false
```
purva-thakre, line 118: my 2 cents on this template is that we don't need an additional `Means of verification` section when there's already a `Steps to reproduce` section above.

.github/issue_template/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+    id: definition-of-done
+    attributes:
+      label: Definition of Done
+      description: "To be checked off during development and review."
+      options:
+        - label: Root cause identified
+        - label: Fix implemented and peer-reviewed
+        - label: Regression test added
+        - label: All tests pass (CI green)
+        - label: Changelog entry added
```
purva-thakre, line 140: question: do we update the changelog as the PRs come in or only before a release?

.github/issue_template/change_request.yml
```diff
@@ -0,0 +1,199 @@
...
+  - type: dropdown
+
+    id: priority
+    attributes:
+      label: Priority
+      options:
+        - Critical
+        - High
+        - Medium
+        - Low
```
purva-thakre, line 14: is this going to automate adding labels to created issues? Pretty cool! 

We need to ensure the labels are added before merging this PR.

.github/issue_template/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+      required: true
+
+  - type: input
+
+    id: responsible
+    attributes:
+      label: Responsible
+      placeholder: "e.g. Jane Doe (Fraunhofer FOKUS)"
+    validations:
+      required: false
```
purva-thakre, line 26: is this for internal purposes only? Because people without repo permissions cannot `@` a Qrisp maintainer.

.github/issue_template/change_request.yml
```diff
@@ -0,0 +1,199 @@
...
+      required: true
+
+  - type: input
+
+    id: target-version
+    attributes:
+      label: Target Version / Milestone
+      placeholder: "e.g. v0.10, Backlog"
+    validations:
+      required: false
```
purva-thakre, line 35: we probably might not need this if we instead use github milestone. Go to an issue, on the RHS you will see the following. It is straightforward to add an issue to a milestone. 

Image

I tried creating a milestone without a due date: https://github.com/eclipse-qrisp/Qrisp/milestone/1

https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones

.github/issue_template/change_request.yml
```diff
@@ -0,0 +1,199 @@
...
+      description: "At least 1 positive, 1 negative, and 2 boundary tests."
+      value: |
+        | Test-ID | Type     | Description | Input | Expected Result |
+        |---------|----------|-------------|-------|-----------------|
+        | T-001   | Positive |             |       |                 |
+        | T-002   | Negative |             |       |                 |
+        | T-003   | Boundary |             |       |                 |
+        | T-004   | Boundary |             |       |                 |
+    validations:
+      required: true
```
purva-thakre, line 175: same as my other comment, I do not think we need this field given that there is a `desired behavior` section above.

purva-thakre (commented):
I am not one of the requested reviewers for the PR. Even then, I had a couple of comments. :)

.github/issue_template/bug_report.yml
bastibock: I would still update the changelog only with a new release, as we did before. But this way we can gather all changes directly and more easily create the changelog for a new release.

.github/issue_template/bug_report.yml
bastibock: The term might not be the best, but I think it makes sense to have a section, where it is tracked how the bug is fixed.
This is not necessarily needed when one user creates the bug issue, but can then be added in later comments/fixes.

.github/issue_template/change_request.yml
bastibock: I'm not sure if it does but I hope so. We can only use the templates, when they are in main, so I couldn't test it unfortunately.

.github/issue_template/bug_report.yml
bastibock: Good point! 
They could still add the name of a person but tagging people with @ would be even better of course. But Eclipse Foundation has some restriction on the settings which we also cannot change.

.github/issue_template/change_request.yml
bastibock: Yes, if we have a milestone the target version is not necessarily needed. One of each is fine here from my opinion.

.github/issue_template/change_request.yml
bastibock: Similar to the template above, I think it makes sense to keep it.
``desired behaviour`` is what it is supposed to be like, and the tests are there to make sure it actually works this way. And it can be added later.

.github/issue_template/feature_request.yml
bastibock: I think I had some general remarks in mind when adding the input field.
But looking at it again it doesn't make too much sense.
I will delete it from the other templates.

.github/issue_template/bug_report.yml
PietropaoloFrisoni: Good point! So if I understand correctly, every person who merges a PR should create a changelog entry (with a super brief description of the PR and a link to the original PR). If so, I think this can save a lot of time for the release manager. That is, it seems a much more scalable approach to me compared to retrieving all the merged PRs since the latest release

.github/issue_template/bug_report.yml
ntchol: I agree with Pietropaulo - we should also have a place to capture and enumerate the tests. This is an open part

.github/issue_template/bug_report.yml
ntchol: "Means of verification" and "Steps to reproduce" are actually two different things. The means of verification would mean that we have (a) test(s) that verify that the bug is fixed or any other approach (e.g. documentation or workflow) that verifies that a bug is removed or a feature is probably implemented.

"Means of verification" relates not only to bugs but also to change requests or feature requests.

.github/issue_template/change_request.yml
ntchol: The "Means of verification" are very important. We need tests or other means that make sure that our features/changes/bug-fixes are properly implemented. This means could even come from a different person. We should have as many as possible "automated" means of verfication such that these can run regularly as regression tests in the CI/CD.

ntchol:
I see one open point: where do we manage the tests ?

bastibock:
@ntchol What do you mean by managing the tests? 
Like where are they gathered or how they should be implemented? Or how they reflect the fix/ implementation?
In what sense to you want to manage them (with little to no overhead)?

PR #519, opened 2026-04-04
QuantumArray upgrades

renezander90:
**Scope:** This PR focusses on **element-wise operations for QuantumArrays**. It includes improvements of typing, documentation, testing, and  bug fixes for existing methods, and implementations of new methods.

**Bug Fixes:**
- Fixed several minor issues that prevented proper functioning of element-wise arithmetic for QuantumArrays.

**New Methods:**
- Element-wise and (`&`), or (`|`), xor (`^`) for QuantumArrays of QuantumBools
- `_reduce_over_axes` helper method
- `all` and `any` methods for QuantumArrays of QuantumBools with support for `axis`

**Refactored Methods:**
- `matmul`, `rmatmul`

**Tests:**

Arithmetic methods for QuantumFloat are tested for quantum-quantum and quantum-classical use (when applicable).

_Static_ (`test_quantum_array.py`)
- `test_quantum_array_element_wise_ops`: Test element-wise operations on QuantumArrays of QuantumFloat/QuantumModulus against their classical counterparts. (`+`, `-`, `*`, `==`, `!=`, `>`, `>=`, `<`, `<=`)
- `test_quantum_array_element_wise_bool_ops`: Test element-wise boolean operations on QuantumArrays of QuantumBool against their classical counterparts. (`&`, `|`, `^`)
- `test_quantum_array_element_wise_inplace_ops`: Test element-wise in-place operations on QuantumArrays of QuantumFloat/QuantumModulus against classical counterparts. (`+=`, `-=`, `*=`)
- `test_quantum_array_all`: Test the all() method on QuantumArrays of QuantumBool against their classical counterparts.
- `test_quantum_array_any`: Test the any() method on QuantumArrays of QuantumBool against their classical counterparts.

_Dynamic_ (`test_jasp_quantum_array.py`)
- `test_quantum_array_element_wise_ops`: Test element-wise operations on QuantumArrays of QuantumFloat against their classical counterparts. (`+`, `-`, `*`, `==`, `!=`, `>`, `>=`, `<`, `<=`)
- `test_quantum_array_element_wise_ops_qm`: Test element-wise operations on QuantumArrays of QuantumModulus against their classical counterparts. (`+`, `-`, `*`, `==`, `!=`, `>`, `>=`, `<`, `<=`)
- `test_quantum_array_element_wise_bool_ops`: Test element-wise boolean operations on QuantumArrays of QuantumBool against their classical counterparts. (`&`, `|`, `^`)
- `test_quantum_array_element_wise_inplace_ops`: Test element-wise in-place operations on QuantumArrays of QuantumFloat against classical counterparts. (`+=`, `-=`, `*=`)
- `test_quantum_array_element_wise_inplace_ops_qm`: Test element-wise in-place operations on QuantumArrays of QuantumModulus against classical counterparts. (`+=`, `-=`, `*=`)
- `test_quantum_array_all`: Test the all() method on QuantumArrays of QuantumBool against their classical counterparts.
- `test_quantum_array_any`: Test the any() method on QuantumArrays of QuantumBool against their classical counterparts.

**Documentation and Typing:**

Improved documentation including examples, and type-checking for element-wise operations:

QuantumArray.__sub__(other) | Performs element-wise subtraction.
QuantumArray.__matmul__(other) | Performs matrix multiplication.
QuantumArray.__mul__(other) | Performs element-wise multiplication.
QuantumArray.__eq__(other) | Performs element-wise == comparison.
QuantumArray.__ne__(other) | Performs element-wise != comparison.
QuantumArray.__gt__(other) | Performs element-wise > comparison.
QuantumArray.__ge__(other) | Performs element-wise >= comparison.
QuantumArray.__lt__(other) | Performs element-wise < comparison.
QuantumArray.__le__(other) | Performs element-wise <= comparison.
QuantumArray.__and__(other) | Performs element-wise & (bitwise AND) operation.
QuantumArray.__or__(other) | Performs element-wise \| (bitwise OR) operation.
QuantumArray.__xor__(other) | Performs element-wise ^ (bitwise XOR) operation.
QuantumArray.__iadd__(other) | Performs element-wise in-place addition.
QuantumArray.__isub__(other) | Performs element-wise in-place subtraction.
QuantumArray.__imul__(other) | Performs element-wise in-place multiplication.
QuantumArray.all([axis]) | Performs an element-wise logical AND reduction, returning True if all elements are True.
QuantumArray.any([axis]) | Performs an element-wise logical OR reduction, returning True if any element is True.

**Notes:** 
- Per-method imports in arithmetic methods, e.g., `from qrisp.qtypes.quantum_float import QuantumFloat`, are required to break circular imports. This is caused by the `QuantumString` which extends `QuantumArray` being a `QuantumType`. We could consider refactoring import structure/ move `QuantumString` to avoid this.


src/qrisp/core/quantum_array.py
```diff
@@ -757,23 +827,34 @@ def __str__(self):
...
 
         Returns
         -------
         QuantumArray
             A new QuantumArray containing the multiplication result.
+                The ``qtype`` of the output array is the same as the ``qtype`` of self.
+
+        .. warning::
+
+            Matrix multiplication is currently not supported in tracing mode if ``qtype`` is QuantumFloat.
```
purva-thakre:
```suggestion
            Matrix multiplication is currently not supported in tracing mode if ``qtype`` of self is QuantumFloat.
```

src/qrisp/core/quantum_array.py
```diff
@@ -757,23 +827,34 @@ def __str__(self):
...
+            If ``qtype`` is QuantumModulus, matrix multiplication is supported in tracing mode and other must be a classical jax numpy array of integers.
```
purva-thakre:
```suggestion
            If ``qtype`` of self is QuantumModulus, matrix multiplication is supported in tracing mode and other must be a classical jax numpy array of integers.
```

purva-thakre:
@renezander90  thanks for the request for review. I am slowly making my way through the PR as I am trying to understand how the`QuantumArray` class works. 

In the following, how is `a_array` defined?

```py
a_array = QuantumArray(QuantumModulus(7), shape=(2,2))
a_array[:] = np.array([[1, 2], [3, 4]])
```
I understand `QuantumModulus(7)` based [on the docs](https://qrisp.eu/reference/Quantum%20Types/QuantumModulus.html#qrisp.QuantumModulus) and `shape=(2,2)` in the context of numpy. However, I do not understand what a `QuantumArray` of these quantities mean. The encoding value being of shape (2,2) makes sense. But why did you have to specify the `QuantumModulus` have size 7?

Is it creating a QuantumArray with QuantumModulus objects of size 7?

renezander90:
@purva-thakre  thanks for reviewing.
This documentation could be helpful: https://qrisp.eu/reference/Core/QuantumArray.html
In Qrisp, a QuantumArray has as first argument `qtype` which is a QuantmVariable that serves as template for the variables in the QuantumArray. 

```
a_array = QuantumArray(QuantumModulus(7), shape=(2,2))
a_array[:] = np.array([[1, 2], [3, 4]])
```

This means that we have a QuantumArray of shape 2 x 2, where each entry is a QuantumModulus with modulus 7. A QuantumModulus must have a fixed modulus from which the number of qubits of the variable is derived, e.g., 3 qubits for modulus 7. (something like ceil(log2(modulus)) in general).

It is creating a QuantumArray with QuantumModulus objects of modulus 7 and size (number of qubits) 3.

In another example, each entry could be a QuantumFloat of 3 qubits:
```
a_array = QuantumArray(QuantumFloat(3), shape=(2,2))
a_array[:] = np.array([[1, 2], [3, 4]])
```

The `a_array[:] = np.array([[1, 2], [3, 4]])` statement prepares the (0,0) variable in 1, the (0,1) variable in 2, etc.

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+
+    @jaspify
+    def main():
+        q_array = QuantumArray(QuantumBool(), shape=(2,2))
+        q_array[:] = a_c
+        qbl = q_array.any(axis=axis)
+        return measure(qbl)
+
+    measured = main()
+    assert np.array_equal(measured, a_c.any(axis=axis)), f"Expected {a_c.any(axis=axis)}, got {measured}"
```
purva-thakre: nit: you do not need to use f-string to catch for unit test failures. If there is a failure, pytest will print these values while letting you know there is an `AssertionError`. For this reason, I am going to suggest some changes in other unit tests to remove the f-string. 

I have noticed Claude likes to add these fstrings. Not really sure why. You have to explicitly let it know to skip adding these. 

```suggestion
    assert np.array_equal(measured, a_c.any(axis=axis))
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+
+    @jaspify
+    def main():
+        q_array = QuantumArray(QuantumBool(), shape=(2,2))
+        q_array[:] = a_c
+        qbl = q_array.all(axis=axis)
+        return measure(qbl)
+
+    measured = main()
+    assert np.array_equal(measured, a_c.all(axis=axis)), f"Expected {a_c.all(axis=axis)}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c.all(axis=axis))
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+        return measure(a_array)
+    
+    # Validate measurements
+    measured = main()
+
+    # Calculate classical reference
+    op(a_c, b_c)
+    a_c %= modulus
+
+    assert np.array_equal(measured, a_c), f"Failed on operator {op.__name__} with {rhs_type} RHS. Expected {a_c}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c)
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+        op(a_array, rhs_operand)
+        return measure(a_array)
+    
+    # Validate measurements
+    measured = main()
+
+    # Calculate classical reference
+    op(a_c, b_c)
+
+    assert np.array_equal(measured, a_c), f"Failed on operator {op.__name__} with {rhs_type} RHS. Expected {a_c}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c)
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
     r_array = main()
 
     assert np.array_equal(r_array, expected_c), f"Failed on operator {op.__name__}. Expected {expected_c}, got {r_array}"
```
purva-thakre:
```suggestion
    assert np.array_equal(r_array, expected_c)
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -185,18 +185,135 @@ def test():
...
+        r_array = op(a_array, rhs_operand)
+        return measure(r_array)
+    
+    # Calculate classical reference
+    expected_c = op(a_c, b_c) % modulus
+    
+    # Validate measurements
+    r_array = main()
+
+    assert np.array_equal(r_array, expected_c), f"Failed on operator {op.__name__}. Expected {expected_c}, got {r_array}"
```
purva-thakre:
```suggestion
    assert np.array_equal(r_array, expected_c)
```

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -185,18 +185,135 @@ def test():
...
+        r_array = op(a_array, rhs_operand)
+        return measure(r_array)
+    
+    # Calculate classical reference
+    expected_c = op(a_c, b_c)
+    
+    # Validate measurements
+    r_array = main()
+
+    assert np.array_equal(r_array, expected_c), f"Failed on operator {op.__name__}. Expected {expected_c}, got {r_array}"
```
purva-thakre:
```suggestion
    assert np.array_equal(r_array, expected_c)
```

tests/core_tests/quantum_array/test_quantum_array.py
```diff
@@ -145,35 +147,192 @@ def main(k):
...
+    pytest.param(np.array([[True, True], [True, True]]), 1, id="All True, axis 1"),
+    pytest.param(np.array([[True, False], [True, True]]), 1, id="One False, axis 1"),
+])
+def test_quantum_array_any(a_c, axis):
+    """Test the any() method on QuantumArrays of QuantumBool against their classical counterparts."""
+    q_array = QuantumArray(QuantumBool(), shape=(2,2))
+    q_array[:] = a_c
+    qbl = q_array.any(axis=axis)
+    measured = qbl.most_likely()
+    assert np.array_equal(measured, a_c.any(axis=axis)), f"Expected {a_c.any(axis=axis)}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c.any(axis=axis))
```

tests/core_tests/quantum_array/test_quantum_array.py
```diff
@@ -145,35 +147,192 @@ def main(k):
...
+    pytest.param(np.array([[True, True], [True, True]]), 1, id="All True, axis 1"),
+    pytest.param(np.array([[True, False], [True, True]]), 1, id="One False, axis 1"),
+])  
+def test_quantum_array_all(a_c, axis):
+    """Test the all() method on QuantumArrays of QuantumBool against their classical counterparts."""
+    q_array = QuantumArray(QuantumBool(), shape=(2,2))
+    q_array[:] = a_c
+    qbl = q_array.all(axis=axis)
+    measured = qbl.most_likely()
+    assert np.array_equal(measured, a_c.all(axis=axis)), f"Expected {a_c.all(axis=axis)}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c.all(axis=axis))
```

tests/core_tests/quantum_array/test_quantum_array.py
```diff
@@ -145,35 +147,192 @@ def main(k):
...
+    op(a_array, rhs_operand)
+
+    # Calculate classical reference
+    op(a_c, b_c)
+    if isinstance(qtype, QuantumModulus):
+        a_c = a_c % qtype.modulus
+
+    # Validate measurements
+    measured = a_array.most_likely()
+    assert np.array_equal(measured, a_c), f"Failed on operator {op.__name__} with {rhs_type} RHS. Expected {a_c}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, a_c)
```

tests/core_tests/quantum_array/test_quantum_array.py
```diff
@@ -145,35 +147,192 @@ def main(k):
...
+    r_array = op(a_array, rhs_operand)
     
+    # Calculate classical reference
+    expected = op(a_c, b_c)
+    if isinstance(qtype, QuantumModulus):
+        expected = expected % qtype.modulus
+    
+    # Validate measurements
+    measured = r_array.most_likely()
+    assert np.array_equal(measured, expected), f"Failed on operator {op.__name__}. Expected {expected}, got {measured}"
```
purva-thakre:
```suggestion
    assert np.array_equal(measured, expected)
```

tests/core_tests/quantum_array/test_quantum_array.py
```diff
@@ -145,35 +147,192 @@ def main(k):
...
+
+@pytest.mark.parametrize("op", ops)
+@pytest.mark.parametrize("rhs_type", rhs_types)
+@pytest.mark.parametrize("instance", instances)
+def test_quantum_array_element_wise_inplace_ops(op, rhs_type, instance):
+    """Test element-wise in-place operations on QuantumArrays of QuantumFloat/QuantumModulus against classical counterparts."""
     
-    assert len(results) > 0, "No measurement results found"
+    if op == operator.imul and rhs_type == "quantum":
+        pytest.skip("Quantum-quantum inplace multiplication is unsupported.")
```
purva-thakre: why not raise an error in `imul` for this and then testing whether the error is raised as expected rather than using `pytest.skip`?

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

purva-thakre (commented):
Publishing my older comments because github won't let me add a new one for some reason. Still working on reviewing the code.

src/qrisp/core/quantum_array.py
```diff
@@ -1183,12 +1394,24 @@ def __mul__(self, other: QuantumArray) -> QuantumArray:
...
 
+        Multiplying a scalar with a QuantumArray of QuantumFloats:
+
+        >>> import numpy as np
+        >>> from qrisp import QuantumArray, QuantumFloat
+        >>> a_array = QuantumArray(QuantumFloat(2), shape=(2,2))
+        >>> a_array[:] = np.eye(2)
+        >>> r_array = a_array * 2
+        >>> print(r_array)
+        # {OutcomeArray([[2, 0], [0, 2]]): 1.0}
```
positr0nium: The actual output here is

```
{OutcomeArray([[1, 0],                                          
              [0, 1]]): 1.0}
```

which seems to be a bug. Apart from that, really solid work!

positr0nium (changes requested):

src/qrisp/core/quantum_array.py
```diff
@@ -519,6 +525,70 @@ def swapaxes(self, axis_1, axis_2):
...
+
+        Parameters
+        ----------
+        operation : Callable
+            A callable that takes a 1D list/array of QuantumVariables
+            and returns a single QuantumVariable.
+        qtype : QuantumVariable
+            The type of the quantum variable to return.
+        axis : int or tuple of ints, optional
+            The axes to reduce over.
```
purva-thakre: should add something here that when `axis = None`, the reduction appears to be over all axis (according to L556)

this function also does not have unit tests.

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+        >>> c_array = QuantumArray(QuantumBool(), shape=(2,2))
+        >>> c_array[:] = np.array([[True, False], [False, False]])
+        >>> print(c_array.any(axis=0))  # Output: [True, False]
+        """
+        from qrisp.qtypes.quantum_bool import QuantumBool
+
+        if not isinstance(self.qtype, QuantumBool):
+            raise TypeError(
+                "The 'any' method is only defined for QuantumArrays of QuantumBools."
+            )
```
purva-thakre, line 2031: minor nit: this should be mentioned in the docstring under a `Raises` section. I know you already have it mentioned in the description but this should be moved to its own section.

purva-thakre: I think there is also a unit test missing for this. You should verify an error is raised when the method is used for QuantumArrays of non-QuantumBools. 

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+        >>> c_array = QuantumArray(QuantumBool(), shape=(2,2))
+        >>> c_array[:] = np.array([[True, False], [True, True]])
+        >>> print(c_array.all(axis=0))  # Output: [True, False]
+        """
+        from qrisp.qtypes.quantum_bool import QuantumBool
+
+        if not isinstance(self.qtype, QuantumBool):
+            raise TypeError(
+                "The 'all' method is only defined for QuantumArrays of QuantumBools."
+            )
```
purva-thakre, line 1976: Same as my comment for `any`. There should be a `Raises` section in the docstring and a unit test for the unallowed cases. Verify the error is raised as expected in static mode.

src/qrisp/core/quantum_array.py
```diff
@@ -784,38 +865,80 @@ def __matmul__(self, other: QuantumArray) -> QuantumArray:
...
+        from qrisp.qtypes.quantum_float import QuantumFloat
+        from qrisp.qtypes.quantum_modulus import QuantumModulus
 
-        if isinstance(self.qtype, QuantumModulus):
-            from qrisp.alg_primitives.arithmetic.jasp_arithmetic.jasp_montgomery import (
-                cq_montgomery_mat_multiply,
+        if self.shape[1] != other.shape[0]:
+            raise ValueError(
+                f"Incompatible shapes for matrix multiplication: {self.shape} and {other.shape}"
             )
```
purva-thakre, line 908: you should add a unit test for this.

src/qrisp/core/quantum_array.py
```diff
@@ -784,38 +865,80 @@ def __matmul__(self, other: QuantumArray) -> QuantumArray:
...
-            n2 = other.shape[1]
-            out = QuantumArray(qtype=self[0, 0], shape=(n1, n2))
-            cq_montgomery_mat_multiply(self, other, out)
-            return out
+        if isinstance(self.qtype, QuantumModulus):
+            if isinstance(other, QuantumArray):
+                raise TypeError(
+                    "Cannot apply matrix multiplication between a QuantumArray of QuantumModulus "
+                    "and another QuantumArray. The second array must be a classical array of integers."
+                )
```
purva-thakre, line 927: needs a unit test.

src/qrisp/core/quantum_array.py
```diff
@@ -784,38 +865,80 @@ def __matmul__(self, other: QuantumArray) -> QuantumArray:
...
+                out = QuantumArray(qtype=self.qtype, shape=(n1, n2))
+                cq_montgomery_mat_multiply(self, other, out)
+                return out
+
         elif isinstance(self.qtype, QuantumFloat):
+            if check_for_tracing_mode():
+                raise NotImplementedError(
+                    "Matrix multiplication between QuantumArrays of QuantumFloat in tracing mode "
+                    "is currently not supported."
+                )
```
purva-thakre, line 945: needs a unit test

src/qrisp/core/quantum_array.py
```diff
@@ -1048,22 +1177,34 @@ def _element_wise_out_of_place_injection(self, other, fun, out_type):
                     (out_view[i] << fun)(self_view[i], other)
             return out
 
-    def _validate_arithmetic(self, other):
+    def _validate_arithmetic(
```
purva-thakre, line 1203: This is an older function that is also missing unit tests. We should also verify errors are raised as expected by this function in a unit test.

src/qrisp/core/quantum_array.py
```diff
@@ -1075,23 +1216,31 @@ def _validate_arithmetic(self, other):
...
         """
         Performs element-wise addition.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be added.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be added.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1252: There should be a unit test verifying an error is raised when there is a shape mismatch in this function. I see it is courtesy of `validate_arithmetic`.

Also, add a `Raises` section to the docstring to mention the errors raised in this function when the inputs are invalid. 

```py
>>> import numpy as np
>>> from qrisp import QuantumArray, QuantumFloat
>>> a_array = QuantumArray(QuantumFloat(2), shape=(2,2))
>>> a_array[:] = np.eye(2)
>>> r_array = a_array + np.eye(3)
>>> print(r_array)
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 5
      1 import numpy as np
      2 from qrisp import QuantumArray, QuantumFloat
      3 a_array = QuantumArray(QuantumFloat(2), shape=(2,2))
      4 a_array[:] = np.eye(2)
----> 5 r_array = a_array + np.eye(3)
      6 print(r_array)
      7 # {OutcomeArray([[2, 0], [0, 2]]): 1.0}

File ~/Qrisp/src/qrisp/core/quantum_array.py:1277, in QuantumArray.__add__(self, other)
   1222 """
   1223 Performs element-wise addition.
   1224 
   (...)   1273 
   1274 """
   1275 from qrisp.qtypes.quantum_float import create_output_qf
-> 1277 self._validate_arithmetic(other)
   1278 if isinstance(other, QuantumArray):
   1279     out_type = create_output_qf([self.qtype, other.qtype], "add")

File ~/Qrisp/src/qrisp/core/quantum_array.py:1215, in QuantumArray._validate_arithmetic(self, other, mode)
   1213 elif isinstance(other, (np.ndarray, jnp.ndarray)):
   1214     if self.shape != other.shape:
-> 1215         raise ValueError(f"Shape mismatch: {self.shape} vs {other.shape}.")

ValueError: Shape mismatch: (2, 2) vs (3, 3).
```

src/qrisp/core/quantum_array.py
```diff
@@ -1116,23 +1287,31 @@ def __add__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise subtraction.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be subtracted.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be subtracted.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1330: Same feedback as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1143,36 +1322,68 @@ def __sub__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise multiplication.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be multiplied.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be multiplied.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1409: same feedback as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1143,36 +1322,68 @@ def __sub__(self, other: QuantumArray) -> QuantumArray:
...
+            If a scalar is provided, it will be multiplied with each element of the QuantumArray.
 
         Returns
         -------
         QuantumArray
             A new QuantumArray containing the element-wise product.
+            If a QuantumArray or QuantumVariable is provided, the ``qtype`` of the output will be determined by the qtypes of the two input objects to prevent overflow.
+            If a classical scalar or numpy array is provided, the ``qtype`` of the output will be the same as the ``qtype`` of self.
```
purva-thakre: I think this line is incorrect. Here, you are referring to numpy array is provided on its own, right? Not that the value is encoded in a QuantumArray? I tried multiplying a numpy array to a QuantumArray and that raised an error. 

```py
>>> import numpy as np
>>> from qrisp import QuantumArray, QuantumFloat
>>> b_array = QuantumArray(QuantumFloat(2), shape=(2,2))
>>> a_array = np.eye(2)
>>> b_array[:] = np.eye(2)
>>> r_array = a_array * b_array
>>> print(r_array)
```

```
TypeError                                 Traceback (most recent call last)
Cell In[13], line 7
      3 # a_array = QuantumArray(QuantumFloat(2), shape=(2,2))
      4 b_array = QuantumArray(QuantumFloat(2), shape=(2,2))
      5 a_array = np.eye(2)
      6 b_array[:] = np.eye(2)
----> 7 r_array = a_array * b_array
      8 print(r_array)

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'multiply'>, '__call__', array([[1., 0.],
       [0., 1.]]), <qrisp.core.quantum_array.QuantumArray object at 0x797722b31e10>): 'ndarray', 'QuantumArray'
```

Switching the order of multiplication i.e. `r_array = b_array * a_array` raised a different error. `Exception: QuantumFloat multiplication for type <class 'float'> not implemented (available are QuantumFloat and int)`

src/qrisp/core/quantum_array.py
```diff
@@ -1197,14 +1420,18 @@ def __mul__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``==`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1475: same comments as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1214,32 +1441,48 @@ def __eq__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``!=`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1533: Same comment as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1249,32 +1492,48 @@ def __ne__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``>`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1591: same comment as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1284,32 +1543,48 @@ def __gt__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``>=`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1649: same comment as __add__

src/qrisp/core/quantum_array.py
```diff
@@ -1319,32 +1594,48 @@ def __ge__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``<`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1707: same comment as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1354,32 +1645,48 @@ def __lt__(self, other: QuantumArray) -> QuantumArray:
...
         """
         Performs element-wise ``<=`` comparison.
 
         Parameters
         ----------
-        other : QuantumArray
-            The QuantumArray to be compared to.
+        other : QuantumArray | QuantumVariable | ArrayLike
+            The array or scalar to be compared to.
+            If an array is provided, it must have the same shape as the original QuantumArray.
```
purva-thakre, line 1765: same comment as `__add__`

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
 
         self._validate_arithmetic(other)
         return self._element_wise_out_of_place_injection(
             other, lambda a, b: a <= b, QuantumBool()
         )
 
+    def __and__(self, other: QuantumArray | QuantumVariable) -> QuantumArray:
+        """
+        Performs element-wise ``&`` (bitwise AND) operation.
+        This operation is only defined for QuantumArrays of QuantumBools.
```
purva-thakre: this is better mentioned in a `Raises` section in the docstring as there are multiple invalid cases. Plus, add unit tests to verify the errors are raised as expected. I think a large number of these unit tests for all the functions can be combined into one with `@pytest.mark.parametrize` as all we are doing is verifying `with pytest.raises(SomeError)`

The code block below helped me raise errors courtesy of `validate_arithmetic`. I think we should explicitly mention these in public facing functions given that the function raising these errors is private. 

- TypeError: Element-wise operations require qtype 'QuantumBool'. Got QuantumFloat.
- TypeError: Element-wise operations require both arrays to have qtype 'QuantumBool'. Got QuantumBool and QuantumFloat.
- ValueError: Shape mismatch: (2, 2) vs (3, 3).
- TypeError: Element-wise operations require qtype 'QuantumBool'. Got QuantumModulus.

```py
import numpy as np
from qrisp import QuantumArray, QuantumBool, QuantumFloat, QuantumModulus

# Valid case (sanity check)
a = QuantumArray(QuantumBool(), shape=(2,2))
b = QuantumArray(QuantumBool(), shape=(2,2))
a[:] = np.array([[True, False], [False, True]])
b[:] = np.array([[True, True], [False, False]])
r = a & b

# 1. Wrong qtype: QuantumFloat instead of QuantumBool
a_float = QuantumArray(QuantumFloat(3), shape=(2,2))
b_float = QuantumArray(QuantumFloat(3), shape=(2,2))
a_float[:] = np.array([[1, 2], [3, 4]])
b_float[:] = np.array([[5, 6], [7, 0]])
r = a_float & b_float

# 2. Mixed qtypes: QuantumBool & QuantumFloat
r = a & a_float

# 3. Shape mismatch
c = QuantumArray(QuantumBool(), shape=(3,3))
c[:] = np.array([[True, False, True], [False, True, False], [True, True, False]])
r = a & c

# 4. QuantumModulus instead of QuantumBool
a_mod = QuantumArray(QuantumModulus(7), shape=(2,2))
a_mod[:] = np.array([[1, 2], [3, 4]])
r = a_mod & a_mod
```

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+
+        self._validate_arithmetic(other, mode="bool")
+        return self._element_wise_out_of_place_injection(
+            other, lambda a, b: a & b, QuantumBool()
+        )
+
+    def __or__(self, other: QuantumArray | QuantumVariable) -> QuantumArray:
+        """
+        Performs element-wise ``|`` (bitwise OR) operation.
+        This operation is only defined for QuantumArrays of QuantumBools.
```
purva-thakre: same comment as `and`.

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+
+        self._validate_arithmetic(other, mode="bool")
+        return self._element_wise_out_of_place_injection(
+            other, lambda a, b: a | b, QuantumBool()
+        )
+
+    def __xor__(self, other: QuantumArray | QuantumVariable) -> QuantumArray:
+        """
+        Performs element-wise ``^`` (bitwise XOR) operation.
+        This operation is only defined for QuantumArrays of QuantumBools.
```
purva-thakre: same comment as `and`

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+        return self._element_wise_out_of_place_injection(
+            other, lambda a, b: a ^ b, QuantumBool()
+        )
+
+    def all(
+        self, axis: int | tuple[int, ...] | None = None
+    ) -> QuantumArray | QuantumVariable:
+        """
+        Performs an element-wise logical AND reduction, returning True if all elements are True.
+        This operation is only defined for QuantumArrays of QuantumBools.
```
purva-thakre: same comment as `and`

src/qrisp/core/quantum_array.py
```diff
@@ -1389,31 +1696,253 @@ def __le__(self, other: QuantumArray) -> QuantumArray:
...
+            return res
+
+        return self._reduce_over_axes(_all, QuantumBool(), axis=axis)
+
+    def any(
+        self, axis: int | tuple[int, ...] | None = None
+    ) -> QuantumArray | QuantumVariable:
+        """
+        Performs an element-wise logical OR reduction, returning True if any element is True.
+        This operation is only defined for QuantumArrays of QuantumBools.
```
purva-thakre: add a `Raises` section for all invalid inputs. We want to make this clear in a public function.

src/qrisp/core/quantum_array.py
```diff
@@ -1444,7 +1973,43 @@ def _element_wise_in_place_call(self, other, fun):
...
                     fun(self_view[i], other)
 
-    def __iadd__(self, other):
+    def __iadd__(
+        self, other: QuantumArray | QuantumVariable | "ArrayLike"
+    ) -> QuantumArray:
+        """
+        Performs element-wise in-place addition.
+        Note that this modifies the original QuantumArray and does not create a new one.
+
```
purva-thakre, line 2092: needs a `Raises` section courtesy of the errors raised by `_element_wise_in_place_call` or `validate_Arithmetic`; not really sure which one as there is an overlap in the errors raised by both functions.
 
Plus, unit tests to verify all the expected errors are raised in this function without any problems.

src/qrisp/core/quantum_array.py
```diff
@@ -1444,7 +1973,43 @@ def _element_wise_in_place_call(self, other, fun):
...
+        --------
+
+        Adding a scalar to a QuantumArray of QuantumFloats,
+        and adding two QuantumArrays:
+
+        >>> import numpy as np
+        >>> from qrisp import QuantumArray, QuantumFloat
+        >>> qa = QuantumArray(QuantumFloat(8,-1), shape=(2,2))
+        >>> qa[:] = np.array([[1.0, 2.0], [3.0, 4.0]])
+        >>> qa += 1.0
```
purva-thakre, line 2123: if I replace this with `-2`, the output corrects to `{OutcomeArray([[127.,   0.],[  1.,   2.]]): 1.0}` due to overflow. You should mention the cause for this in the docstring. 

I don't really understand why it is `127` though. Can you help me understand this? In the clockwork analogy it is correcting backwards but where is the number `127` coming from?

src/qrisp/core/quantum_array.py
```diff
@@ -1453,7 +2018,43 @@ def f(a, b):
         self._element_wise_in_place_call(other, f)
         return self
 
-    def __isub__(self, other):
+    def __isub__(
+        self, other: QuantumArray | QuantumVariable | "ArrayLike"
+    ) -> QuantumArray:
+        """
+        Performs element-wise in-place subtraction.
+        Note that this modifies the original QuantumArray and does not create a new one.
```
purva-thakre, line 2143: same comments as `iadd`.

src/qrisp/core/quantum_array.py
```diff
@@ -1462,18 +2063,61 @@ def f(a, b):
         self._element_wise_in_place_call(other, f)
         return self
 
-    def __imul__(self, other):
+    def __imul__(self, other: "ArrayLike") -> QuantumArray:
+        """
+        Performs element-wise in-place multiplication.
+        Note that this modifies the original QuantumArray and does not create a new one.
```
purva-thakre, line 2193: same comments as `iadd`

src/qrisp/core/quantum_array.py
```diff
@@ -1462,18 +2063,61 @@ def f(a, b):
...
+        --------
+
+        Multiplying a scalar with a QuantumArray of QuantumFloats,
+        and scaling a QuantumArray by a numpy array:
+
+        >>> import numpy as np
+        >>> from qrisp import QuantumArray, QuantumFloat
+        >>> qa = QuantumArray(QuantumFloat(8,-1), shape=(2,2))
+        >>> qa[:] = np.array([[1.0, 2.0], [3.0, 4.0]])
+        >>> qa *= 2.0
```
purva-thakre, line 2231: mention in the docstring: scalar must be positive. After I changed this line to `-2.0`, I got `Exception: Tried to inplace-multiply unsigned QuantumFloat with negative factor`. 

I think it is the same if I make one of the elements on L2096 negative i.e. `qa *= np.array([[-0.5, 1.5], [2.5, 3.5]])`. I am not sure if this is a bug though. 

Edit: I see this is due to the in-place multiplication function in SBP arithmetic. Is there an alternative to using this? Our plan is to move away from the functionality in this module anyway.

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+@pytest.mark.parametrize("rhs_type", rhs_types)
+@pytest.mark.parametrize("instance", instances)
+def test_quantum_array_element_wise_inplace_ops(op, rhs_type, instance):
+    """Test element-wise in-place operations on QuantumArrays of QuantumFloat against classical counterparts."""
+    
+    if op == operator.imul and rhs_type == "quantum":
+        pytest.skip("Quantum-quantum inplace multiplication is unsupported.")
+
+    if op == operator.imul and rhs_type == "classical":
+        pytest.skip("Quantum-classical inplace multiplication currently is unsupported in Jasp.")
```
purva-thakre: rather than using `skip`, add unit tests to verify with `pytest.raises`

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -185,18 +185,135 @@ def test():
...
+
+@pytest.mark.parametrize("op", ops)
+@pytest.mark.parametrize("rhs_type", rhs_type)
+@pytest.mark.parametrize("instance", instances)
+def test_quantum_array_element_wise_ops_qm(op, rhs_type, instance):
+    """Test element-wise operations on QuantumArrays of QuantumModulus against their classical counterparts."""
+
+    if op == operator.mul and rhs_type == "quantum":
+        # qq multiplication fixed in separate pull request, but for now we skip this test to avoid CI failures
+        pytest.skip("Quantum-quantum multiplication for QuantumModulus is currently unsupported in Jasp.")
```
purva-thakre, line 273: use `with pytest.raises`

tests/jax_tests/test_jasp_quantum_array.py
```diff
@@ -214,3 +331,166 @@ def main():
...
+
+@pytest.mark.parametrize("op", ops)
+@pytest.mark.parametrize("rhs_type", rhs_types)
+@pytest.mark.parametrize("instance", instances)
+def test_quantum_array_element_wise_inplace_ops_qm(op, rhs_type, instance):
+    """Test element-wise in-place operations on QuantumArrays of QuantumModulus against classical counterparts."""
+    
+    if op == operator.imul and rhs_type == "quantum":
+        # qq multiplication fixed in separate pull request, but for now we skip this test to avoid CI failures
+        pytest.skip("Quantum-quantum inplace multiplication is currently unsupported in Jasp.")
```
purva-thakre: use `with pytest.raises`

purva-thakre (changes requested):

tests/jax_tests/test_jasp_quantum_array.py
renezander90: This will be fixed in https://github.com/eclipse-qrisp/Qrisp/pull/495 which we anticipate to be finalized soon (I any case before 0.9). Hence, I would not introduce a new exception for this particular case and keep as is. The other suggestions regarding ``with pytest.raises`` were implemented in bbb25a472d3554c0656cd896a3a89378c7e7c748

src/qrisp/core/quantum_array.py
renezander90: Good catch! Actually, there is another problem which is that, different from ``add``, ``sub``, for ``mul`` the ``qtype`` of the result is not the ``qtype`` of self. I will raise an exception and leave quantum-classical multiplication for a separate PR.

renezander90: And also this issue: #520

src/qrisp/core/quantum_array.py
renezander90: Done in 914f533f8cead0ec9624054a68b9fe1be667f861

src/qrisp/core/quantum_array.py
renezander90: Done in 914f533f8cead0ec9624054a68b9fe1be667f861

src/qrisp/core/quantum_array.py
renezander90: Done in ac8d804cbccd27458ce6026594b7a94423ef6894 and b26a664e894a24f83e6caa6676c65f7d5cb0505f for ``add``and the other arithmetic methods.

src/qrisp/core/quantum_array.py
renezander90: Done in 89fef2213bd6f43d57cb082db902b86266c94c3e

src/qrisp/core/quantum_array.py
renezander90: Done in ac8d804cbccd27458ce6026594b7a94423ef6894

src/qrisp/core/quantum_array.py
renezander90: In your example, the QuantumArray is on the right. The error is because numpy's ``mul`` is called first. Adding an ``rmul`` for QuantumArray also does not help.

src/qrisp/core/quantum_array.py
renezander90: 8a7a6c698e2478fcf96f2cdf4e0fd3ed7a2173f6 adds overflow hint. ``QuantumFloat(8,-1)`` represents numbers 0, 0.5, 1, 1.5, ..., 127, 127.5, then -1 + 128 = 127.

src/qrisp/core/quantum_array.py
renezander90: Added hint: 2727cf15cb89369d41e77149d6d2e2e8fd16d37f. As far as I know, there is only the quantum-quantum multiplier: https://github.com/eclipse-qrisp/Qrisp/blob/8d550e326d9f8428785122df105761d27e7d8858/src/qrisp/alg_primitives/arithmetic/jasp_arithmetic/jasp_multiplyers.py#L61

positr0nium (approved):
Thanks!

purva-thakre (approved):

PR #526, opened 2026-04-13
QuantumCircuit refactoring Part 4

PietropaoloFrisoni:
## Description

This PR is the final refactoring of the core `QuantumCircuit` class, which includes a partial refactoring of  the `Operation` and `standard_operations` modules to improve code quality, consistency, and maintainability. 

The `QuantumCircuit.append` method is decomposed into focused private helpers, exception types are narrowed to semantically correct subtypes, all standard-operation factory functions receive `NumPy`-style docstrings, and the helper functions `convert_to_qb_list` / `convert_to_cb_list` are rewritten for clarity. Substantial new test coverage is added for both the gate factories and the `QuantumCircuit` API.

## Related Issues

Closes / addresses the known bug (see the corresponding discussion on DIscord) in the `r()` gate method where the `theta` and `ph`i parameter names were swapped relative to `RGate` (as mentioned before, this was previously discussed on Discord).

## Type of Change

- Bug fix
- Refactoring (no new feature, behaviour-preserving except where noted)
- Tests
- Documentation / docstrings

## Breaking Change?

Yes, there are two intentional breaking changes:

- `QuantumCircuit.r()`: the positional argument order is corrected from `r(phi, theta, qubits)` to `r(theta, phi, qubits)` to match the underlying `RGate(theta, phi)` convention. The previous order was a bug (parameter names were internally reversed). Callers using positional arguments must swap their `theta` and `phi` values. Keyword callers are unaffected if they were already using the documented names.

- `QuantumCircuit.barrier()`: the `clbits` parameter is removed. This parameter has always been documented as "currently ignored" and has never had any effect on the circuit. Code passing clbits to `barrier()` will now raise a `TypeError`. Since the argument was a no-op, no circuit behaviour changes.

## What was changed?

- `src/qrisp/circuit/quantum_circuit.py`

`append()` refactored, `convert_to_qb_list` / `convert_to_cb_list` rewritten, `r(phi, theta, ...)` corrected, `barrier() drops the unused clbits parameter`, `to_qasm()` bare except etc.

The other changed files are a consequence/extension of such changes (see the changed files) for details.

## How was it tested?

All existing tests pass unchanged.
New unit tests added for every changed component (see above).


positr0nium (approved):
Thanks :)

Issue #527, opened 2026-04-13
Mid-level interfacing with the FullStaQD compiler


julfarn:
Hello everybody, 

as qrisp will play an important role in the [FullStaQD compilation framework](https://github.com/FullStaQD/compiler), at least in the form of a privileged frontend, we should start efforts to connect the two softwares. Ultimately, we would like to offer FullStaQD as one possible compilation path for qrisp code, handing over the quantum program at a high level of abstraction. 

But to first establish some minimal working pipeline, we aim to connect to the existing mid-level jasp dialect, and an IR similar to the one emitted by the `.to_mlir()` method. To reach our [first milestone in June](https://github.com/FullStaQD/compiler/milestone/1), we need to be able to compile simple qrisp programs with classical control flow down to QIR and the [HiSEP-Q](https://github.com/caps-tum/HiSEP-Q-2.0) QISA. As of now, we can parse IR containing the `jasp` and `linalg` dialects, and translate this to the dialects of [MQT core](https://github.com/munich-quantum-toolkit/core).

---

Here's how we imagine the integration from the qrisp side. Most of these ideas have been discussed before, but hopefully we can use this issue to make some decisions. 

### User interaction
From the user perspective, I believe it would be idiomatic for qrisp to settle the compiler calls next to the existing `to_mlir()`, `to_qasm()`, etc. methods of `Jaspr` objects. These could be new methods `to_hisepq()`, or flags of existing ones, like `to_qir(use_qcc=True)`. 

### Dependency
We offer a `qcc` binary of the compiler. As qrisp likely does not want to formally introduce a dependency on the FullStaQD compiler this early in the development, our suggestion is to delegate this problem to the user. Qrisp could assume that there is a `qcc` executable on the `PATH`, and simply run it for the compilation. 

### IR Interface
@positr0nium was nice enough to join a discussion in the FullStaQD repository, and figured out how to lower from `stablehlo` to `linalg` within the qrisp codebase in [this comment](https://github.com/FullStaQD/compiler/pull/18#issuecomment-4209658674). We therefore expect as input IR format the result of the `to_mlir()` function, puth through the `stablehlo-convert-to-linalg` pass. Input and output of `qcc` are textual MLIR through stdin and stdout, or to/from files.

--- 

Do you as qrisp developers agree with this picture, or do you have other ideas? 
If we go along as proposed here, would it be realistic that the task of creating a `to_hisepq()` method, which 
 - [ ] lowers to the interface IR,
[...]

PR #528, opened 2026-04-15
Linalg lowering

positr0nium:
# PR: Add StableHLO Lowering Option for MLIR Emission

## Summary

This PR introduces a `lower_stablehlo` parameter to the MLIR emission pipeline, allowing StableHLO operations to be lowered to lower-level dialects (linalg, arith, tensor) for improved compatibility with downstream MLIR tooling.

### New Feature: `lower_stablehlo` Parameter

- **`Jaspr.to_mlir(lower_stablehlo=False)`** - Added optional parameter to control StableHLO lowering
- **`jaspr_to_mlir(jaspr, lower_stableHLO=False)`** - Updated internal function signature
- **`jaxpr_to_xdsl(jaxpr, lower_stableHLO=False, ...)`** - Updated xDSL conversion function

### MLIR Pass Pipeline

When `lower_stablehlo=True`, the following MLIR passes are applied:

1. `symbol-dce` - Removes unused private shadow functions that JAX emits
2. `stablehlo-convert-to-signless` - Converts to signless integers
3. `stablehlo-legalize-to-linalg` - Converts arithmetic/data ops to linalg

### Files Modified

- `src/qrisp/jasp/jasp_expression/centerclass.py` - Updated `to_mlir()` method signature and docstring
- `src/qrisp/jasp/mlir/jaxpr_lowering.py` - Added `lower_jaxpr_to_linalg_MLIR()` function and updated docstrings
- `src/qrisp/jasp/mlir/mlir_emission.py` - Updated `jaspr_to_mlir()` signature and added docstring
- `tests/jax_tests/test_mlir.py` - Added parametrized tests for `lower_stablehlo=True/False`

### Test Coverage

All MLIR tests now run with both `lower_stablehlo=False` (default) and `lower_stablehlo=True`:

- `test_mlir_generation`
- `test_mlir_basic_dialect_operations`
- `test_mlir_quantum_control_flow_rewriting`
- `test_mlir_grovers_algorithm`
- `test_mlir_qae_algorithm`
- `test_mlir_iqpe`
- `test_mlir_iqae`
- `test_mlir_hamiltonian_simulation`
- `test_mlir_qaoa`
- `test_mlir_array_operations`
- `test_mlir_jasp_dialect_registration`

### Known Limitations

The `stablehlo-legalize-to-linalg` pass only lowers arithmetic and element-wise data operations. Some operations like `stablehlo.scatter` remain unlowered as they require specialized lowering patterns not available in the standard pass.

## Usage Example

```python
from qrisp import *
from qrisp.jasp import make_jaspr

def example():
    qv = QuantumVariable(2)
    h(qv[0])
    cx(qv[0], qv[1])
    return measure(qv)

jaspr = make_jaspr(example)()

# Default: StableHLO operations preserved
mlir_stablehlo = jaspr.to_mlir(lower_stablehlo=False)

# New: StableHLO arithmetic lowered to linalg/arith
mlir_linalg = jaspr.to_mlir(lower_stablehlo=True)
```

## Relevant Issues:

https://github.com/eclipse-qrisp/Qrisp/issues/527


positr0nium:
Who wants to review this? @PietropaoloFrisoni? @renezander90? Any takers?

src/qrisp/jasp/mlir/jaxpr_lowering.py
```diff
@@ -120,40 +122,59 @@ def generic_mlir_to_xdsl(mlir_string: str) -> builtin.ModuleOp:
     """
     # Create context with unregistered operations allowed so our custom ops
     # remain intact when parsed by xDSL.
+    from xdsl.dialects import builtin, func, linalg, arith, tensor, scf
```
renezander90:
```suggestion
    from xdsl.dialects import builtin, func, linalg, arith, tensor, scf, math
```

src/qrisp/jasp/mlir/jaxpr_lowering.py
```diff
@@ -120,40 +122,59 @@ def generic_mlir_to_xdsl(mlir_string: str) -> builtin.ModuleOp:
...
     ctx = Context()
     ctx.allow_unregistered = True
 
     # Register essential dialects used by the wrapper module and functions.
     ctx.load_dialect(builtin.Builtin)
     ctx.load_dialect(func.Func)
+    ctx.load_dialect(linalg.Linalg)
+    ctx.load_dialect(arith.Arith)
+    ctx.load_dialect(tensor.Tensor)
+    ctx.load_dialect(scf.Scf)
```
renezander90, line 140:
```suggestion
    ctx.load_dialect(scf.Scf)
    ctx.load_dialect(math.Math)
```

Register math dialect. Test:

```
from qrisp import *

def main():

    a = jnp.pow(2.0, 3.0)
    return a

jaspr = make_jaspr(main)()
mlir = jaspr.to_mlir(lower_stablehlo=True) 
print(mlir)
```

Before:
> %4 = "math.powf"(%1, %2) <{fastmath = #arith.fastmath<none>}> : (f64, f64) -> f64

After:
>%4 = math.powf %1, %2 : f64

renezander90 (changes requested):
Thanks! Please note that ``math`` dialect should be registered.

ctfle:
I just tried to install this branch and ran into problems when using `pip`. The MLIR exporter seems to require xDSL v.0.55 but `pip` goes for v.0.64 since its not specified in the dependencies. `uv pip` resolves this correctly and installs xDSL v.0.55. Maybe you can add the explicit dependencies and/or recommend using `uv`?

purva-thakre:
> The MLIR exporter seems to require xDSL v.0.55 but pip goes for v.0.64 since its not specified in the dependencies. uv pip resolve this correctly and installs xDSL v.0.55. Maybe you can add the explicit dependencies and/or recommend using uv

Yes, we are planning to add explicit dependencies soon in #553. I have been busy with other things. Adding `uv` is also on my todos.

renezander90:
> > The MLIR exporter seems to require xDSL v.0.55 but pip goes for v.0.64 since its not specified in the dependencies. uv pip resolve this correctly and installs xDSL v.0.55. Maybe you can add the explicit dependencies and/or recommend using uv
> 
> Yes, we are planning to add explicit dependencies soon in #553. I have been busy with other things. Adding `uv` is also on my todos.

Actually, dependency should be xDSL 0.59.0 which is required for the latest pennylane-catalyst

renezander90 (approved):

PR #534, opened 2026-04-20
Linalg lowering upgrades

renezander90:
### Overview

This PR adds a new `mlir_rewrites/` subpackage with three xDSL pattern-rewrite passes that clean up the IR produced by `stablehlo-legalize-to-linalg`. It also fixes an internal API issue where `jaxpr_to_xdsl` / `MLIR_str_to_xdsl` did not return the xDSL `Context` needed to run xDSL transform passes.

Fixes issue https://github.com/eclipse-qrisp/Qrisp/issues/535.

### Changes

**Internal API fix** (jaxpr_lowering.py, mlir_emission.py)

`MLIR_str_to_xdsl` and `jaxpr_to_xdsl` now return a `(ctx, module)` tuple instead of just the module. The `Context` is required to run xDSL transform passes such as `CanonicalizePass` and `DeadCodeElimination`.

**New rewrite passes** (mlir_rewrites/)

Three composable xDSL pattern-rewrite passes targeting artifacts from `stablehlo-legalize-to-linalg`:

- **`scalar_linalg_folding`** (`FoldScalarLinalgGeneric`) — Folds 0-dimensional `linalg.generic` ops back into plain scalar arithmetic with `tensor.extract` / `tensor.from_elements`. The linalg lowering wraps every scalar arithmetic op in a `linalg.generic` region, which carries significant overhead for what is semantically a single instruction. Run only when `lower_stablehlo=True`.

- **`scalar_tensor_folding`** (`FoldExtractFromElements`, `EraseDeadFromElements`) — Eliminates `tensor.extract(tensor.from_elements(x))` round-trips on 0-D tensors, replacing them with the underlying scalar directly. `EraseDeadFromElements` is needed as a fallback because xDSL's standard DCE skips `tensor.from_elements` (it lacks the `Pure` trait). Run unconditionally, followed by `CanonicalizePass` and `DeadCodeElimination`.

- **`cmpi_extui_folding`** (`FoldCmpiExtui`) — Folds comparisons of zero-extended booleans (`extui i1 → cmpi`) against constants 0 or 1 back to the original `i1` value, its `NOT`, or a constant `true`/`false`. This eliminates multi-op chains like `measure (i1) → extui → cmpi eq 0 → extui → extract → cmpi ne 0` which JAX regularly emits for measurement-based control flow. Run unconditionally, followed by `CanonicalizePass` and `DeadCodeElimination`.

### Test

One new test in **test_mlir.py**:

**`test_mlir_linalg_folding_after_stablehlo_lowering`** — verifies that after `to_mlir(lower_stablehlo=True)`, no `linalg.generic` operations remain in the output MLIR string.

**test_mlir_rewrites.py** — Unit tests for the xDSL rewrite passes, operating directly on hand-constructed IR (no Qrisp tracing required).

**`TestFoldScalarLinalgGeneric`** (3 tests)
- `test_successful_fold` — verifies that a valid 0-D `linalg.generic` is replaced by `tensor.extract` + scalar arithmetic + `tensor.from_elements`
- `test_no_folds[non-0d]` — verifies that a generic with iterator types (non-scalar) is left untouched
- `test_no_folds[multi-output]` — verifies that a generic producing more than one output is left untouched

**`TestScalarTensorFolding`** (3 tests):
- `test_successful_bypass_extract` — verifies that `tensor.extract` on a 0-D `tensor.from_elements` is bypassed entirely, with the result pointing directly to the original scalar, and that the now-stranded `from_elements` is cleaned up by `EraseDeadFromElements`
- `test_erase_dead_from_elements` — verifies that a `tensor.from_elements` with no uses is erased (targeted DCE fallback for ops lacking the `Pure` trait)
- `test_preserve_live_from_elements` — verifies that a `tensor.from_elements` that has an active user (a `ReturnOp`) is left untouched

**`TestCmpiExtUIFolding`** (23 tests)
- `test_all_fold_cases` (20 parametrized) — exhaustively covers every `(predicate, rhs ∈ {0, 1})` combination from the fold table, asserting the correct outcome: `IDENTITY` (no new ops), `NOT` (XOR with 1), `TRUE` (constant `1 : i1`), or `FALSE` (constant `0 : i1`)
- `test_no_folds` (3 parametrized) — verifies the three early-return guard conditions: LHS is not an `extui`, RHS is outside `{0, 1}`, and predicate index is out of range — all leave the `cmpi` untouched


src/qrisp/jasp/mlir/mlir_rewrites/cmpi_extui_folding.py
julfarn, line 1: I think you are missing an `__init__.py` in `mlir_rewrites`.

julfarn (commented):
Hi, I tried out this functionality and it's great to see how polished the output looks now. Since we want to connect to this exporter, I have a few questions to help us plan:

- Are you planning on making further changes to the output, or do you expect this to get merged roughly as is? If the latter, I would use this branch to produce example input for us for now. 
- The target branches' PR is tagged with a Milestone with deadline June 5. Is this a hard date, so that we can expect this feature released by then? Or is it just a rough internal timeline?

Thanks for the efforts!

src/qrisp/jasp/mlir/mlir_rewrites/cmpi_extui_folding.py
renezander90: Thanks for noticing 👍. This PR still needs review which could result in changes, but I would expect roughly as is. June 5th is the deadline for the feature freeze for 0.9 which will be released on June 12th.

renezander90:
Close polluted PR and open new PR.

PR #537, opened 2026-04-22
Add Gidney's venting adder

purva-thakre:
As defined in https://arxiv.org/abs/2507.23079


documentation/source/reference/Primitives/Gidney venting adder.rst
```diff
@@ -0,0 +1,20 @@
...
+
+.. currentmodule:: qrisp
+
+.. autofunction:: gidney_cq_venting_adder
+
+.. autofunction:: carry_venting_adder
+
+.. autofunction:: dirty_ancillae_adder
+
+.. autofunction:: carry_xor_block
```
positr0nium: I think we don't need to expose these internal functions in the documentation. They're not that interesting for users who just want to use the adder. People who are willing to dig into the specifics have to clone the repo anyway.
Exposing the helper functions therefore just creates an unnecessary risk for breaking changes if we want to refactor the implementation.

src/qrisp/alg_primitives/arithmetic/adders/gidney_venting_adder.py
```diff
@@ -0,0 +1,861 @@
...
+
+    Notes
+    -----
+    The function exists so that callers don't need to repeat the
+    ``if a_int_is_bigint … else …`` pattern at every callsite and to
+    guarantee the result is a JAX scalar (``jnp.bool_``) rather than a
+    plain Python ``bool`` — JAX cannot trace through Python scalars
+    inside compiled loops.
+    """
+    if a_int_is_bigint:
```
positr0nium: Wouldn't it be simpler to do this via ``isinstance(a_int, BigInteger)``? It might also be beneficial to import this function from the gidney adder implementation.

src/qrisp/alg_primitives/arithmetic/adders/gidney_venting_adder.py
```diff
@@ -0,0 +1,861 @@
...
+    """
+    with control(b):
+        if ctrl is None:
+            qrisp.x(parity_check_ctrl)
+        else:
+            qrisp.cx(ctrl, parity_check_ctrl)
+
+    # MCX stays 2-control — ctrl is NOT added as a third control because
+    # ctrl conditions the parity flips above, not the MCX itself.
+    # When ctrl=|0⟩ and b=1, the CX flips are no-ops → MCX sees un-flipped
```
positr0nium: MCX sees un-flipped input but this doesn't means that the MCX doesn't flip it's target. Effectively this means that this function can still have non-trivial behavior ctrl = |0> and inputs are unlucky. I guess it works so inputs are never unlucky but this is very hacky and should have a big warning in the docstring.

src/qrisp/alg_primitives/arithmetic/adders/gidney_venting_adder.py
```diff
@@ -0,0 +1,861 @@
...
+                                     │
+        simple_ctrl: ────────────────•─────────
+                                     │
+        target: ─────────────────────X─────────
+
+    Used in the carry chain when the classical addend bit decides whether the control fires on zero or one.
+    """
+    with control(b):
+        if ctrl is None:
+            qrisp.x(parity_check_ctrl)
```
positr0nium: The ``qrisp.`` syntax is used nowhere else in the codebase

purva-thakre: Yes. I was being lazy in avoiding importing them at the top of the file. Will apply a final commit regarding this.

documentation/source/reference/Primitives/Gidney venting adder.rst
purva-thakre: from my perspective, the `venting` step might have more applications. which is why I wanted to keep them public. But perhaps I am jumping ahead. We could make them public when the need arises.

src/qrisp/alg_primitives/arithmetic/adders/gidney_venting_adder.py
```diff
@@ -0,0 +1,861 @@
...
+
+    Parameters
+    ----------
+    target : list[qrisp.Qubit] | qrisp.QuantumVariable
+        Target register, little-endian (index 0 = LSB). Modified in place.
+    d : int
+        Classical addend (compile-time constant).
+    ancilla : qrisp.QuantumVariable
+        2-qubit clean ancilla register for the streaming carry chain.
+    c_in : qrisp.Qubit | qrisp.QuantumVariable | None
```
positr0nium: Similar point as in the Cuccaro PR: ``QuantumVariable`` could in theory also hold more than one qubit and is therefore unsuitable here.

tests/primitives_tests/arithmetic_tests/test_gidney_venting_adder.py
```diff
@@ -0,0 +1,557 @@
...
+    (1, True,  0, None),
+    # ctrl=1 → matches uncontrolled
+    (0, True,  1, 1),
+    # ctrl=0 → parity flips suppressed, MCX fires on un-flipped state (= behaves as b=False)
+    (1, True,  1, 0),
+]
+
+@pytest.mark.parametrize("ctrl_val, b, expected_tgt, ctrl_qbl_val", ALL_BIT_INVERTED_CASES)
+def test_bit_inverted_mcx_jasp(ctrl_val, b, expected_tgt, ctrl_qbl_val):
+    @jaspify
```
positr0nium, line 78: Why not ``boolean_simulation``? Note that for the caching to work, you should not use ``pytest.mark.parametrize`` but perform the loop inside the test such that the compiled version can be retrieved from cache.

tests/primitives_tests/arithmetic_tests/test_gidney_venting_adder.py
```diff
@@ -0,0 +1,557 @@
...
+    assert int(measured_always_one) == 1
+    # target toggles when (ctrl_val XOR b) = 1; always_one=1 ensures simple_ctrl fires
+    assert int(measured_tgt) == expected_tgt
+    if ctrl_qbl_val is not None:
+        assert int(measured_ctrl_qbl) == ctrl_qbl_val
+
+
+@pytest.mark.parametrize("q0v, q1v, expected_tgt", ZZ_PARITY_CASES)
+def test_zz_mcx_jasp(q0v, q1v, expected_tgt):
+    @jaspify
```
positr0nium, line 110: Same here and below

tests/primitives_tests/arithmetic_tests/test_gidney_venting_adder.py
```diff
@@ -0,0 +1,557 @@
...
+        ctrl_qbl[:] = 1
+        with control(ctrl_qbl[0]):
+            gidney_cq_venting_adder(3, target)
+        return measure(target)
+
+    a = unctrl()
+    b = ctrl_on()
+
+    assert a.get("t", 0) == b.get("t", 0)
+    assert a.get("t", 0) > 0
```
positr0nium: I'm missing a large loop of ``boolean_simulation`` calls that goes through an exhaustive set of parameter combintations :)

positr0nium (changes requested):
Nice work :) We should soon think about how we combine this with a qq adder. (maybe with Cuccaro since they both have constant space requirements?)

documentation/source/reference/Primitives/Gidney venting adder.rst
positr0nium: It would really cool if it had more applications, but yes lets keep it private until then.

positr0nium:
I just noticed: There seems to be also a problem with the phase behavior. The following script should always return ``{0 : 1.0}``:

```python
from qrisp import *

@terminal_sampling
def unctrl(i):
    target = QuantumFloat(i)
    h(target)

    gidney_cq_venting_adder(1, target)
    with invert():
        gidney_adder(1, target)

    h(target)
    return target

for i in range(3, 64):
    print(unctrl(i))
```

tests/primitives_tests/arithmetic_tests/test_gidney_venting_adder.py
purva-thakre: @positr0nium This function relies on Hadamard to make mid-circuit X basis measurements. The following fails with `Exception: Classical function simulator can't process gate h`. is there a reccomended workaround that allows me to use `boolean_simulation`? 

```py
from qrisp import QuantumFloat, measure, h
from qrisp.jasp import boolean_simulation


@boolean_simulation
def apply_h():
    qf = QuantumFloat(1)
    h(qf[0])
    return measure(qf)


result = apply_h()
```

purva-thakre: I guess one option could be to do Z basis measurement instead and redo the phase correction calculations from the Gidney paper with the new corrections.

positr0nium: Ahh, yeah right. Okay in this case boolean simulation will not work at all

purva-thakre:
> I just noticed: There seems to be also a problem with the phase behavior. The following script should always return {0 : 1.0}:

@positr0nium I wonder if this is a limitation of the adder itself. If I edit your example to use the older gidney adder, it works without any issues. The code block works fine with the Gidney venting adder as long as the target input is not in superposition. The phase corrections are applied after mid-circuit measurements and maybe the algorithm itself does
not preserve phase coherence in superposition. 

I can try to run Gidney's original code to see if this is the issue. The paper does not mention anything related to this.

src/qrisp/alg_primitives/arithmetic/adders/gidney_venting_adder.py
purva-thakre: > It might also be beneficial to import this function from the gidney adder implementation.

Yes, since there was an open PR for the Gidney function when I created this PR, I wanted to make sure there were not conflicts between the two branches.

purva-thakre:
There is a weird circular import issue that occurs when the venting adder is added to the `init` file. It mostly affects the Remaud adder unit tests. 

When I removed the venting adder in [fa5f380](https://github.com/eclipse-qrisp/Qrisp/pull/537/commits/fa5f380fb3633ebdfe64d17f51cbe6013a5757e5), the remaud unit tests passed. 

When I added back the venting adder in https://github.com/eclipse-qrisp/Qrisp/pull/537/commits/87de6dcc972480ea6f387f26dba59a35a30949db, the tests are currently passing for both 3.11 and 3.12. But they were previously failing in https://github.com/eclipse-qrisp/Qrisp/pull/537/commits/9216f98053930ca2a9aef916ea2b27a157c2998b. The only difference between the two commits is that I am not running the entire test suite.

tests/primitives_tests/arithmetic_tests/test_gidney_venting_adder.py
```diff
@@ -0,0 +1,1032 @@
...
+        pytest.param(
+            4,
+            marks=pytest.mark.skip(
+                reason="stochastic behavior under @terminal_sampling"
+            ),
+        ),
+        pytest.param(
+            5,
+            marks=pytest.mark.skip(
+                reason="stochastic behavior under @terminal_sampling"
```
purva-thakre: right now the venting adder behaves non-deterministically w/terminal sampling when the quantum target for in-place addition is of size 3, 4, 5. The functions need to have a special case to consider the Z-phase corrections for inputs of these sizes.

purva-thakre: @positr0nium Digging through the code of the original paper, I think the adder is not supposed to work safely on small inputs. The non-deterministic behavior I saw for n= 2,...,5 is the norm and not an issue of terminal sampling. For these sizes, the Z phase corrections cannot be applied properly. In these cases, there are way too many ancillas compared to the input sizes. The special cases I have been considering also fail to work due to this. 

You can see this [in the code of the paper](https://zenodo.org/records/15866587) as well. The venting adder only works at n>=6. Before this, the paper uses a non-venting adder. The functions you are looking for is `circuit_to_iadd_using_3_clean_qubits` and `circuit_to_iadd_using_linear_clean_qubits` in the Zenodo upload. My function also follows the same approach by using the older Gidney adder for inputs of small sizes. The failures currently in the PR are due to https://github.com/eclipse-qrisp/Qrisp/pull/537#issuecomment-4730379906

purva-thakre:
@positr0nium Thanks for the bug fix! This PR is also no longer failing with the `Found jrange with external carry value` error for the Remaud adder unit tests. 

I know you discussed the bug briefly this morning. But I am still not sure how the fixes for the bug also speed up the QCLA adder unit tests a bit. Neither the QCLA adder functions nor the unit tests utilize qache in their files.

positr0nium (approved):
Really nice work! :)

Issue #540, opened 2026-04-26
Jax fails to run xxyy gate under conjugate or invert environment.


Redperij:
We have been struggling with jax error trying to use `BE.apply_rus()` function with our basic implementation of FOQCS-LCU.
I traced it down to `xxyy` gate used in `unbalanced_W_state()`. Had to remove all `len(qv)` or `qv.size` calls before that.

So I decided to run a small test with just the built-in Qrisp functionality. Using the `xxyy` gate under conjugate environment.

The test fails with the error in the [test_xxyy_prep_conjugate_jasp_error.txt](https://github.com/user-attachments/files/27097780/test_xxyy_prep_conjugate_jasp_error.txt). Commenting out `@terminal_sampling` makes the test pass, as well as using `prep(qv)`without conjugate.

```python
def test_xxyy_prep_conjugate_jasp():
    import numpy as np
    from qrisp import QuantumVariable, conjugate, xxyy
    from qrisp.jasp import terminal_sampling

    coeffs = np.ones(6, dtype=complex)

    def prep(qv):
        xxyy(coeffs[0], np.pi / 2, qv[0], qv[1])

    @terminal_sampling
    def main():
        qv = QuantumVariable(6)
        with conjugate(prep)(qv):
            pass
        return qv

    main()
```

renezander90:
Fixed in 6a703c59d7bc95c81d480bf528e43594f3b12e66

PR #541, opened 2026-04-26
Fixing ``Operation.inverse`` on composite gates with abstract parameters.

positr0nium:
Consider the following code:

```python

from qrisp import *
from sympy import Symbol

phi = Symbol("phi")
psi = Symbol("psi")

qv = QuantumVariable(2)
xxyy(phi, psi, qv[0], qv[1])

op = qv.qs.data[2].op

inv_op = op.inverse()

print(inv_op.params)
```

Before the fix, the output would be an empty list, which cause issue such as https://github.com/eclipse-qrisp/Qrisp/issues/540.
This is now fixed. The inverse gate now carries precisely the same paramers as the original. This is a valid fix because the underlying definition circuit is already daggered, such that the parameters need to undergo no further transformation to reflect a properly daggered unitary.

### DIFF
--- a/src/qrisp/circuit/operation.py
+++ b/src/qrisp/circuit/operation.py
@@ -279,6 +279,11 @@ class Operation:
                 res = inverse_circ.to_op(name=self.name[:-3])
             else:
                 res = inverse_circ.to_op(name=self.name + "_dg")
+            
+            res.params = list(self.params)
+            res.abstract_params = set(self.abstract_params)
+            res.permeability = dict(self.permeability)
+            res.is_qfree = self.is_qfree
 
         elif self.name == "qb_alloc":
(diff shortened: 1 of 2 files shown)


src/qrisp/circuit/operation.py
```diff
@@ -279,6 +279,11 @@ def inverse(self):
                 res = inverse_circ.to_op(name=self.name[:-3])
             else:
                 res = inverse_circ.to_op(name=self.name + "_dg")
+            
+            res.params = list(self.params)
+            res.abstract_params = set(self.abstract_params)
+            res.permeability = dict(self.permeability)
```
renezander90: This should be redundant: permeability and is_qfree are already set in line 303/304 below.

PR #542, opened 2026-04-27
Refactoring the `maxCut.py` module

PietropaoloFrisoni:
A simple PR to add some minimal and non-breaking improvements (mostly based on `pylint` with a few type hints here and there) to the `maxCut.py` module in the QAOA section.

I purposely didn't implement all the possible type hints as I was not sure about most of them.

### DIFF
--- a/src/qrisp/algorithms/qaoa/problems/maxCut.py
+++ b/src/qrisp/algorithms/qaoa/problems/maxCut.py
@@ -16,20 +16,33 @@
 ********************************************************************************
 """
 
-from qrisp import *
-import numpy as np
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Callable
+
 import jax.numpy as jnp
-
-from numba import njit, prange
+import numpy as np
 from jax import jit, vmap
+from numba import njit, prange
+
+from qrisp.algorithms.qaoa import QAOAProblem, RX_mixer
+from qrisp.core import rzz
+from qrisp.jasp import check_for_tracing_mode
+
+if TYPE_CHECKING:
+    import networkx as nx
+
+    from qrisp.core import QuantumVariable
 
 
-def maxcut_obj(x, G):
+def maxcut_obj(x: str, G: nx.Graph) -> int:
+    """Compute the MaxCut objective value for a single bitstring and graph G."""
     return maxcut_obj_jitted(int(x[::-1], 2), list(G.edges()))
 
 
 @njit(cache=True)
-def maxcut_obj_jitted(x, edge_list):
+def maxcut_obj_jitted(x: int, edge_list: list[tuple[int, int]] | np.ndarray) -> int:
+    """JIT-compiled MaxCut objective: count edges crossing the cut encoded by integer x."""
     cut = 0
     for i, j in edge_list:
         # the edge is cut
@@ -40,8 +53,12 @@ def maxcut_obj_jitted(x, edge_list):
 
 
 @njit(parallel=True, cache=True)
-def maxcut_energy(outcome_array, count_array, edge_list):
-
+def maxcut_energy(
+    outcome_array: np.ndarray,
+    count_array: np.ndarray,
+    edge_list: np.ndarray,
+) -> float:
+    """Compute the weighted sum of MaxCut objectives over all measurement outcomes."""
     res_array = np.zeros(len(outcome_array))
     for i in prange(len(outcome_array)):
         res_array[i] = maxcut_obj_jitted(outcome_array[i], edge_list) * count_array[i]
@@ -49,9 +66,10 @@ def maxcut_energy(outcome_array, count_array, edge_list):
     return np.sum(res_array)
 
 
-def create_maxcut_cl_cost_function(G):
+def create_maxcut_cl_cost_function(G: nx.Graph) -> Callable[[dict], float]:
     """
-    Creates the classical cost function for an instance of the maximum cut problem for a given graph ``G``.
+    Creates the classical cost function for an instance of the maximum cut
+    problem for a given graph ``G``.
 
     Parameters
     ----------
@@ -61,11 +79,12 @@ def create_maxcut_cl_cost_function(G):
     Returns
     -------
     cl_cost_function : function
-        The classical cost function for the problem instance, which takes a dictionary of measurement results as input.
+        The classical cost function for the problem instance, which takes a
+        dictionary of measurement results as input.
 
     """
 
-    def cl_cost_function(counts):
+    def cl_cost_function(counts: dict) -> float:
 
         edge_list = np.array(list(G.edges()), dtype=np.uint32)
 
@@ -75,7 +94,7 @@ def create_maxcut_cl_cost_function(G):
         if not isinstance(counts_keys[0], str):
 
             for c_array in counts_keys:
-                integer = int("".join([c for c in c_array])[::-1], 2)
+                integer = int("".join(list(c_array))[::-1], 2)
                 int_list.append(integer)
         else:
             for c_str in counts_keys:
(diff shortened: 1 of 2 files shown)


src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -16,20 +16,33 @@
...
-def maxcut_obj(x, G):
+
+def maxcut_obj(x: str, G: nx.Graph) -> int:
+    """Compute the MaxCut objective value for a single bitstring and graph G."""
     return maxcut_obj_jitted(int(x[::-1], 2), list(G.edges()))
 
 
 @njit(cache=True)
-def maxcut_obj_jitted(x, edge_list):
+def maxcut_obj_jitted(x: int, edge_list: list[tuple[int, int]] | np.ndarray) -> int:
```
PietropaoloFrisoni: The `edge_list` name made me think that this is supposed to be a `list`, but it can actually be a `numpy` array as well. The type hint reflects both options. If this is not intentional, we can cast it to be one or the other

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -134,18 +184,19 @@ def create_maxcut_cost_operator(G):
...
     -------
-    cost_operator : function
+    maxcut_cost_operator : Callable[[QuantumVariable, float], None]
         A function receiving a :ref:`QuantumVariable` and a real parameter $\gamma$.
         This function performs the application of the cost operator.
 
     """
 
-    def maxcut_cost_operator(qv, gamma):
+    def maxcut_cost_operator(qv: QuantumVariable, gamma: float) -> None:
```
PietropaoloFrisoni: For `gamma` and similar parameters I would like to introduce a separate `Param` (or a similar name) type alias which include `float`, `sympy.Symbol`, `int`, etc.

renezander90: There is a distinction between static and dynamic mode. In static mode they can be ``sympy.Symbol``, but not in dynamic mode. In dynamic mode, they can also be jnp.float/jnp.int/Tracer. Not sure how/if to reflect this distinction in the typing.

renezander90: In the GQSP and BlockEncoding modules (no sympy here), we mostly used jax.typing.ArrayLike:
Bildschirmfoto 2026-05-01 um 14 17 19

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -85,45 +105,74 @@ def cl_cost_function(counts):
...
-        cut = jnp.sum(bools)  # Count the number of edges crossing the cut
+        # Count the number of edges crossing the cut
+        cut = jnp.sum(bools)
         return -cut
 
     return cut_computer
 
 
-def create_maxcut_sample_array_post_processor(G):
+def create_maxcut_sample_array_post_processor(G: nx.Graph) -> Callable:
```
renezander90: We should also include this function in the documentation: https://qrisp.eu/reference/Algorithms/qaoa/implementations/maxCut.html

and add a note that it is only relevant for dynamic (Jasp) QAOA usage. (See example and tutorial:
https://www.qrisp.eu/reference/Algorithms/qaoa/JaspQAOA.html
https://www.qrisp.eu/general/tutorial/JaspQAOAtutorial.html)

The Callable should take "ArrayLike" (array of integers) as input and return jax.Array (0-D array of float) as output.

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -85,45 +105,74 @@ def cl_cost_function(counts):
...
+    Create a JIT-compiled function that computes the cut value for a given graph G.
+
+    Parameters
+    ----------
+    G : nx.Graph
+        The graph for the problem instance.
+
+    Returns
+    -------
+    Callable[[int], jnp.ndarray]
```
renezander90: Gemini says:

Image

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -85,45 +105,74 @@ def cl_cost_function(counts):
...
+        A JIT-compiled function mapping an integer bitstring to its (negated) cut value.
+
+    """
     edge_list = jnp.array(G.edges())
 
     @jit
-    def cut_computer(x):
+    def cut_computer(x: int) -> jnp.ndarray:
```
renezander90: ``def cut_computer(x: ArrayLike) -> jax.Array``

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -85,45 +105,74 @@ def cl_cost_function(counts):
...
         outcome_array = np.array(int_list, dtype=np.uint32)
 
-        return maxcut_energy(outcome_array, counts_array, edge_list)
+        return maxcut_energy(outcome_array, counts_array, edge_array)
 
     return cl_cost_function
 
 
 @jit
 def extract_boolean_digit(integer, digit):
```
renezander90, line 116: ``def extract_boolean_digit(integer: ArrayLike, digit: ArrayLike) -> jax.Array:``

src/qrisp/algorithms/qaoa/problems/maxCut.py
```diff
@@ -85,45 +105,74 @@ def cl_cost_function(counts):
...
+    """Extract a single boolean digit from an integer at the given bit position."""
     return (integer >> digit) & 1
 
 
-def create_cut_computer(G):
+def create_cut_computer(G: nx.Graph) -> Callable[[int], jnp.ndarray]:
+    """
+    Create a JIT-compiled function that computes the cut value for a given graph G.
```
renezander90, line 123: We should mention JAX JIT-complied to hint JAX compatibility, and distinguish between the numba JIT functions above.

src/qrisp/algorithms/qaoa/problems/maxCut.py
renezander90: Should also rather be ``Callable(["ArrayLike"], jax.Array]`` since the input can be a Tracer.

renezander90 (changes requested):
Nice work, thanks! Some minor comments.

src/qrisp/algorithms/qaoa/problems/maxCut.py
PietropaoloFrisoni: Thanks, @renezander90. I reply here, but this also applies to most of the other comments. This is also closely related to the [type hints PR](https://github.com/eclipse-qrisp/Qrisp/pull/543), but we can start the discussion here.

I'm generally skeptical about using `ArrayLike` or `Array`, as they are very broad and also include simple scalars such as `int`, `complex`, etc. For example, it happens many times that we declare a variable `x` as `ArrayLike` and then we use an index-based notation such as `x[...]`, which is inconsistent and would raise an error for pure scalars.

The way I tried to address this is to create two type aliases, one for scalar-only parameters and one for array-only parameters. But if we have a function that truly works with all the types included in `ArrayLike`, then I don't see any problem with using it (as is the case for some of the functions in this module, such as `cut_computer`)

src/qrisp/algorithms/qaoa/problems/maxCut.py
PietropaoloFrisoni: > Not sure how/if to reflect this distinction in the typing.

I think the `|` syntax is perfect here. So far I just omitted the type hint for `gamma`, but I will address this in the PR that introduces typing across the codebase.

src/qrisp/algorithms/qaoa/problems/maxCut.py
PietropaoloFrisoni: This is an example of the issue I was referring to:
```
Operator ">>" not supported for types "ArrayLike" and "ArrayLike"
  Operator ">>" not supported for types "bool_" and "number[Any, int | float | complex]"
  Operator ">>" not supported for types "bool_" and "float"
  Operator ">>" not supported for types "bool_" and "complex"
  Operator ">>" not supported for types "number[Any, int | float | complex]" and "bool_"
  Operator ">>" not supported for types "number[Any, int | float | complex]" and "number[Any, int | float | complex]"
  Operator ">>" not supported for types "number[Any, int | float | complex]" and "bool"
  Operator ">>" not supported for types "number[Any, int | float | complex]" and "int"
  Operator ">>" not supported for types "number[Any, int | float | complex]" and "float"
  ...Pylance[reportOperatorIssue](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportOperatorIssue.md)
  ```

src/qrisp/algorithms/qaoa/problems/maxCut.py
PietropaoloFrisoni: I am not sure what you mean. Are you referring to the docstring or the type hint (or both)?

src/qrisp/algorithms/qaoa/problems/maxCut.py
renezander90: Yes, I agree. The distinction between scalar-only and array-only parameters would be helpful.

src/qrisp/algorithms/qaoa/problems/maxCut.py
PietropaoloFrisoni: I would just omit the type hint here for now and then address this in a following PR

renezander90 (approved):

PR #543, opened 2026-04-27
Introducing a `typing` module

PietropaoloFrisoni:
## Description

This Pr Introduces `qrisp.typing`. That is, a new module that provides a small set of curated type aliases for use in function signatures throughout the codebase. This software-engineering improvement was discussed on Discord.

The scalar/array distinction in Jasp mode (where Python scalars become jax.Array tracers at trace time) was discussed [in a prior PR](https://github.com/eclipse-qrisp/Qrisp/pull/542#discussion_r3180254709). The plan is to extend coverage to more of the source code in future PRs.

## Type of Change

This is a non-breaking change as it only affects documentation and type hints.

## What was changed?

We have a new `src/qrisp/typing.py` module with six type aliases, all importable from `qrisp.typing`.

All the public circuit methods in `quantum_circuit.py `  now carry QubitLike, ClbitLike, and/or Param annotations and matching docstring entries. We also annotate all paraetric-gate factory functions in `standard.operations.py`  and `U3Gate` in `operation.py`.

New reference page describing all six aliases, their values, and intended use. Linked from the Utilities index.

## How was it tested?

`tests/test_typing.py` covers runtime `isinstance` checks for all six type aliases against representative accepted values. The annotations were verified statically with `Pylance`: no new warnings are introduced in the annotated files.


documentation/source/reference/Circuit Manipulation/Clbit.rst
```diff
@@ -1,3 +1,5 @@
+.. _Clbit:
```
PietropaoloFrisoni, line 1: This was missing. Now :ref:`Clbit` should work across the codebase (it is also used in `typing.py`)

documentation/source/reference/Utilities/typing.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. py:data:: NDArrayLike
+   :type: TypeAlias
+   :value: np.ndarray | jax.Array | jax.core.Tracer
+
+   A multi-dimensional array value. Covers NumPy arrays, JAX arrays, and JAX
+   tracers. Use this when a parameter is expected to be an array, 
+   and the function should be compatible with both NumPy and JAX. 
+
+.. py:data:: ArrayLike
```
PietropaoloFrisoni, line 50: These is almost the same as `jax.ArrayLike`. Since a lot of code in `qrisp` is already using `jax.ArrayLike` as a type hint, this should hopefully be backward-compatible.

src/qrisp/alg_primitives/state_preparation/qswitch_state_preparation.py
PietropaoloFrisoni, line 1: This file finally does not have any warning from `Pylance`

renezander90:
Please note the upcoming changes with new JAX versions:

**JAX 0.8.2**
Bildschirmfoto 2026-05-20 um 18 25 02

**JAX 0.9.1**
Bildschirmfoto 2026-05-20 um 18 23 58

- Does `ScalarLike` intentionally only allow static scalars? Otherwise, it should include `jax.core.Tracer`.
- `Param` should explicitly include `jax.core.Tracer`. If `Param` only refers to scalar parameters, we should remove `jax.Array`. Traced scalars are recognized as `jax.core.Tracer` (also with JAX 0.7.1):

Bildschirmfoto 2026-05-20 um 18 33 59

PietropaoloFrisoni:
Thanks!

The type hints are now forward-compatible with `jax 0.8.2`. `ScalarLike` now explicitly includes `jax.core.Tracer`, since scalar values might indeed become tracers during. Similarly, `jax.core.Tracer` is now explicit in `Param` (without this, the old definition would have silently broken in `jax ≥ 0.8.2`, where `Tracer` no longer inherits from `jax.Array`). I kept `jax.Array` as well, since 0-d concrete jax arrays (e.g. `jnp.array(1.5)`) are valid gate parameters distinct from traced values.

renezander90 (approved):

Issue #466, opened 2026-03-05
Meta: Open quantum arithmetic issues (tracking)


diehoq:
## Summary
This meta-issue tracks all currently open issues in `eclipse-qrisp/Qrisp` that are directly related to quantum arithmetic (`QuantumFloat`/`QuantumModulus`, adders, multiplication, division, modular arithmetic).

Snapshot date: **2026-03-08**

### Correctness bugs
- [ ] #492 
- [ ] #468
- [ ] #437
- [ ] #422
- [ ] #418
- [ ] #328
- [ ] #337
- [ ] #111
- [ ] #103
- [ ] #20

### Refactors / feature completion
- [ ] #395
- [ ] #393
- [ ] #302
- [ ] #467
- [ ] #469 
- [ ] #499 
- [ ] #500 

### Testing / tooling gaps around arithmetic paths
- [ ] #392
- [ ] #301
- [ ] Missing/insufficient unit tests for adder implementations and variants.

### Performance
- [ ] #211

## Notes on overlap
- #111 and #392 concern the same `q_int_div` surface (runtime correctness vs missing test coverage).
- #20 is explicitly referenced as related work in #395.
- #422 and #302 both point to multiplication behavior in Jasp contexts (`QuantumFloat` inplace multiplication path).
- #467 tracks replacement of legacy SBP-based `qf_inversion`/`q_div` paths and is related to division/inversion modernization.
- #468 tracks static-mode uncomputation faults from missing modulus propagation in injection (`<<`) flows.

## Proposed use of this issue
Use this as an umbrella tracker for prioritization and cross-linking PRs that touch arithmetic internals.

purva-thakre:
> The gidney adder

Yes, as I said in my previous comment, there are only a few for this. 

I missed the qcla adder yesterday. Indeed, it does. https://github.com/eclipse-qrisp/Qrisp/blob/ee999922e76a00f7b08175e71594f3f37c0c9a0b/tests/primitives_tests/arithmetic_tests/test_qcla.py#L19

purva-thakre:
@diehoq @renezander90 What are your thoughts on adding an `Adder` class in the quantum arithmetic module?

Yesterday, I discussed with @positr0nium how the main in-place adder function could work. The previous idea was to have `inpl_add` utilize the other adders available in Qrisp. In order to allow a custom adder option to this function, we would need to use the `ammend_inpl_adder` and `inpl_adder_test` to verify the custom function is a valid adder. 

[...]

PR #545, opened 2026-04-27
Introduce `Adder` class

purva-thakre:
As discussed in https://github.com/eclipse-qrisp/Qrisp/issues/466#issuecomment-4176108216


MatP1337:
Great idea, lfg!!!

Pull request of the highest class (ba dum tss)

edit: @purva-thakre this means having different adders as methods, similarly to ``BlockEncoding``? That's imo the way to go to make qrisp closer to classical devs/domain experts struggling with the bandwidth needed to "get quantum".

purva-thakre:
> this means having different adders as methods

No. The new `Adder` class is just a wrapper to the existing adders. By using the `Adder` class as a decorator of the existing in-place adder, I simply get to verify the existing in-place adder is a valid function performing in-place addition.

MatP1337:
> > this means having different adders as methods
> 
> No. The new `Adder` class is just a wrapper to the existing adders. By using the `Adder` class as a decorator of the existing in-place adder, I simply get to verify the existing in-place adder is a valid function performing in-place addition.

Gotcha... great idea imo! Big up!

purva-thakre:
Closing due to rebasing issues.

PR #549, opened 2026-05-02
Add Jasp → Quake memory-semantics lowering backend

renezander90:
## Description

Implements a complete Jaspr → Quake MLIR lowering backend that compiles Qrisp/Jasp functions to native CUDA-Q kernels. This covers a multi-pass MLIR lowering pipeline, the CUDA-Q ingestion/bridge layer, a static qubit-register allocator, and the user-facing `@cudaq_kernel` decorator — including `"sample"` and `"run"` (default) execution modes.

Compatibility: `cudaq>=0.16.0` and `xdsl==0.59.0` (new optional dependency group: `pip install -e ".[cudaq]"`).

### Examples

```python
import cudaq
from qrisp import *

@cudaq_kernel
def bell():
    qv = QuantumVariable(2)
    h(qv[0])
    cx(qv[0], qv[1])
    return measure(qv)

# Run the kernel once (statevector simulation)
result = bell()
print("Single run result:", result)

# Run the kernel multiple times (shots)
results = cudaq.run(bell, shots_count=10)
print("10 shots:", results)
```

```python
from qrisp.operators import X, Y, Z
import networkx as nx

@cudaq_kernel
def hybrid_example():
    G = nx.Graph()
    G.add_nodes_from([0, 1, 2, 3])
    G.add_edges_from([(0, 1), (1, 2), (2, 3), (3, 0)])

    H = sum(X(i) for i in G.nodes) + sum(Z(i)*Z(j) for i, j in G.edges)
    U = H.trotterization()

    a = QuantumFloat(4)
    b = QuantumFloat(4)

    a[:] = 5
    h(b)
    U(a, t=1.0, steps=10)

    # Real-time control flow based on measurement results
    c = measure(b) < 5
    with control(c):
        a += 1

    return measure(a)

result = hybrid_example()
print("Single run result:", result)

results = cudaq.run(hybrid_example, shots_count=10)
print("10 shots:", results)
```

But, **does it run BlockEncodings?**

```python
import cudaq
import numpy as np
    
from qrisp import *
from qrisp.block_encodings import BlockEncoding

A = np.array([[0.73255474, 0.14516978, -0.14510851, -0.0391581],
            [0.14516978, 0.68701415, -0.04929867, -0.00999921],
            [-0.14510851, -0.04929867, 0.76587818, -0.03420339],
            [-0.0391581, -0.00999921, -0.03420339, 0.58862043]])
b = np.array([0, 1, 1, 1])
kappa = np.linalg.cond(A)

BA = BlockEncoding.from_array(A)
BA_inv = BA.inv(0.01, 2)

# Prepares operand variable in state |b>
def prep_b():
    operand = QuantumVariable(2)
    prepare(operand, b)
    return operand

@cudaq_kernel
def main():
    qv = prep_b()
    ancs = BA_inv.apply(qv)
    measured_ancs = [measure(anc) for anc in ancs]
    return measure(qv), *measured_ancs
    
results = cudaq.run(main, shots_count=10)
print("10 shots:", results)
```
> 10 shots: [(2, True, 14.0), (3, True, 4.0), (3, True, 4.0), (0, True, 4.0), (3, True, 0.0), (0, True, 4.0), (3, True, 0.0), (3, True, 4.0), (3, True, 0.0), (3, True, 4.0)]

## Reviewer Notes

- Qrisp's `cudaq_kernel` decorator eagerly compiles a Qrisp function into a proper CUDA-Q kernel that can be executed with `cudaq.run` like a native kernel. The `run` variant is aligned with Qrisp's/Jasp programming model and the default. Optionally, a Qrisp function can be compiled into a kernel for `cudaq.sample` via `cudaq_kernel(qrisp_func, execution_mode="sample")`. `cudaq.sample` is more aligned with the quantum circuit programming paradigm, it does not support real-time control. Today's quantum hardware is however mostly compatible with `sample` and not (yet) with `run`. Hence, we provide the option to use `sample`.

- `register_size` (static-register allocation) is **opt-in only** via `cudaq_kernel(..., register_size=N)`. This is only relevant for the `sample` variant and **not** for the (default) `run` variant: `cudaq.sample` supports only limited circuit optimization/memory management compared to `cudaq.run`. This necessitates the conversion into a static register for circuits with many allocation/deallocations. 

- `cudaq` 0.16 throws the following warning: "FutureWarning: The CUDA-Q `sample` and `observe` algorithmic primitives will change in a future release. Existing code may require updates. See https://nvidia.github.io/cuda-quantum/latest/using/migration/upcoming_changes.html for details." The link leads to an (almost) empty page. 

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

**Dialect definitions** (`quake_lowering/dialects/`)

- `quake_dialect.py` — xDSL dialect for all Quake ops used in the pipeline (`quake.alloca`, `quake.dealloc`, `quake.extract_ref`, `quake.veq_size`, `quake.subveq`, `quake.concat`, `quake.relax_size`, `quake.mz`, `quake.discriminate`, `quake.reset`, gate ops) with correct CUDA-Q assembly formats (functional-type, flat operand lists, no bracket-prefix controls).
- `cc_dialect.py` — xDSL dialect for Classical Compute ops (`cc.alloca`, `cc.load`, `cc.store`, `cc.loop`, `cc.if`, `cc.switch`, `cc.condition`, `cc.continue`, `cc.sequence_data/size`, `cc.compute_ptr`, `cc.log_output`, `cc.insert_value`, `cc.undef`, etc.).

**Jasp → Quake lowering helpers** (`quake_lowering/jasp_to_quake/`)

- `gate_mapping.py` — mapping table from Qrisp gate names to Quake op constructors, including decompositions for gates without a direct Quake equivalent (e.g. `sx → h/sdg/h`, `xxyy`, `gphase`).
- `helper_functions.py` — shared rewrite-pattern utilities used across passes 1a/1b.
- `jasp_to_quake.py` — thin orchestrator chaining pass 1a and 1b.
- `lower_jasp_to_quake.py` — **op lowering.** Rewrites every `jasp.*` op to its Quake equivalent (`create_qubits→alloca`, `get_qubit→extract_ref`, `get_size→veq_size`, `slice→subveq`, `fuse→concat`, `delete_qubits→dealloc`, `quantum_gate→quake.<gate>`, `measure→mz+discriminate`, `reset→reset`), threading `!jasp.QuantumState` values backwards so they become dead (structural removal happens in 1b). Carries a `mode: "run"|"sample"` parameter through all rewrites (`"run"`: measurement bit-packing loop; `"sample"`: raw `quake.mz` only).
- `strip_qst.py` — **QuantumState structural elimination** from `scf.while/if/for/index_switch` operands/block-args/yields/results, `func.call`/`func.return`, and `func.func` signatures/attributes; special-cases `jasp.create_quantum_kernel`/`consume_quantum_kernel` wrapper ops.

**Remaining lowering passes** (`quake_lowering/lowering_passes`)

- `scf_to_cc.py` — **SCF → CC dialect lowering.** Converts `scf.while`, `scf.if`, `scf.for` to `cc.loop`/`cc.if` via pure SSA threading (`cc.continue`/`cc.condition`); unwraps loop-carried `tensor<T>` scalars at region boundaries; includes a hotfix for a CUDA-Q compiler bug with inclusive loop bounds (`sge → sgt(B-1)`).
- `scalar_tensor_unwrap.py` — folds rank-0 `tensor<T>` round-trips (`tensor.from_elements`/`tensor.extract`) and rewrites function signatures/return types to bare scalar types.
- static_veq_alloca.py *(new since the original draft)* — rewrites `quake.alloca !quake.veq<?>[%n : i64]` into `quake.alloca !quake.veq<N>` whenever `%n` is constant-fed, inserting `quake.relax_size` casts only where a consumer (control-flow block args, `func.call`/`func.return`) still requires the dynamic type.
- `ranked_tensor_to_array.py` — lowers `tensor<NxT>` constants/element accesses to `cc.alloca`/`cc.compute_ptr`/`cc.load`; rewrites rank-1 tensor arguments to `!cc.ptr<!cc.array<T x N>>` for dynamic indexing across function-call boundaries.
- `array_to_sequence.py` — rewrites `!cc.ptr<!cc.array<T x N>>` entrypoint parameters to `!cc.sequence<T>` (with `cc.sequence_data` extraction) so kernels accept NumPy array arguments at the C++ boundary, then casts back to `!cc.ptr<...>` internally.
- `safeguard_no_ranked_tensor_linalg.py` — validation pass rejecting IR with `ranked_tensor`/`linalg` ops that wouldn't survive the CUDA-Q compiler.
- `validation_tools.py` — `validate_quake_mlir` helper: parses generated MLIR against the live CUDA-Q context, raising a descriptive error on failure.
- `jaspr_to_quake.py` — `jaspr_to_quake_mlir(jaspr, execution_mode="run")` orchestrates all passes in order and returns a valid Quake MLIR string.

**Static qubit-register allocation** (Only relevant for `samlpe` variant.)

- `interpreter_tools/interpreters/static_register_interpreter.py` + `interpreter_tools/scalar_list.py` — transforms a Jaspr to pre-allocate a single fixed-size qubit register and manage qubit create/delete via index bookkeeping (mirroring the Catalyst interpreter's strategy), replacing every `AbstractQuantumState` with a `(pre_alloc_array, free_indices, abs_qst)` triple. `ScalarList` represents its contents as individual scalar leaves rather than a ranked array, so it lowers cleanly to CUDA-Q/Quake (no tensor/linalg equivalent needed). This works around CUDA-Q's lack of proper qubit deallocation/reuse, exposed via the new `register_size` argument on `cudaq_kernel`/`Jaspr.to_quake_mlir`.

**`Jaspr.to_quake_mlir`** (`jasp_expression/centerclass.py`)

- New public method exposing the Quake lowering pipeline directly on `Jaspr`, with `execution_mode="run"|"sample"`.

**CUDA-Q bridge** (`cudaq_interface/`)

- `annotations.py` — `FixedShapeNDArray[dtype, size]` annotation class for typed NumPy array kernel parameters, including dummy-value generation for Jaspr tracing.
- `cudaq_ingestion/` (cudaq_prep.py, `xdsl_ingestion.py`) — synthesizes `.run`/`.run.entry` variants for `"run"` mode (incl. multi-value `cc.struct` packing for tuple returns); emits a minimal void-return module for `"sample"` mode.
- cudaq_kernel.py — `cudaq_kernel(func=None, execution_mode="run", register_size=None)`: user-facing decorator, supporting `@cudaq_kernel` and `@cudaq_kernel(execution_mode=..., register_size=...)` forms; traces eagerly at decoration time via `make_jaspr`; supports `int`, `float`, `bool`, and `FixedShapeNDArray` parameters; compiled kernels are directly callable (single-shot) or usable with `cudaq.run`/`cudaq.sample`.
- `__init__.py` — re-exports public API.

**Additional Changes**

- `algorithms/gqsp/gqsp.py` — replaced `theta[::-1]`/`phi[::-1]` reversed-array indexing (unsupported by this lowering path) with equivalent forward index arithmetic (`theta[d - i]`, etc.), needed for GQSP to trace/lower correctly to CUDA-Q.

**Packaging / CI**

- pyproject.toml — new `cudaq = ["cudaq>=0.16.0"]` optional dependency group; added `pytest-timeout==2.4.0` to `dev-testing`.

**Docs**

- New `CUDA-Q Interface.rst` reference page (`cudaq_kernel`, `FixedShapeNDArray`), `Jaspr.to_quake_mlir` generated API page, and a full CUDAQ.ipynb tutorial (Bell state, hybrid quantum-classical logic with operators/trotterization, mid-circuit measurement with feed-forward, while-loop with accumulated measurement, kernel parameters/arrays, variational workflows), linked from the tutorial index and changelog-dev.rst.

**Test cleanup**

- Removed an accidental duplicate `test_reflection_quantum_array` definition in test_reflection.py.

## How was it tested?

New test suite under cudaq_interface_tests plus test_static_register_interpreter.py 

(Note: Actual number of test is likely higher than reported here.)

| File | Tests | Coverage |
|---|---|---|
| test_quake_lowering.py | 37 | Quake assembly-format validity (functional-type, no bracket-prefix controls), alloc/dealloc, static vs. dynamic `veq` staticization + `relax_size`, gate/measurement lowering, qubit indexing, classical math lowering, negative tests |
| test_cudaq_kernel.py | 30 | Decorator forms (`run`/`sample`), multiple return values, scalar (`int`/`float`/`bool`) and `FixedShapeNDArray` parameters (static/dynamic indexing, nested functions, loops, conditionals), multi-shot execution, annotation validation/error handling |
| test_cudaq_control_flow.py | 16 | Invert/conjugate, classical/quantum control, `q_cond`, `q_switch` (2/3-branch, classical/quantum operand), `q_while_loop` (incl. accumulation), `q_fori_loop` (incl. nested), `jax`-native `fori_loop`, `jrange` |
| test_cudaq_static_register.py | 16 | Static-register allocation, qubit reuse, classical control, slicing, extend/append/prepend, while-loop + `qache` interaction, `QuantumBool`, Bell-state correlation under register reuse |
| test_cudaq_gates.py | 11 | Exact-statevector equivalence (vs. Qrisp's simulator) for all `GATE_MAP` gates, controlled variants, and composite/decomposed 2-qubit gates (`cp`, `swap`, `rxx`, `rzz`, `xxyy`) |
| test_cudaq_alg_primitives.py | 11 | Higher-level algorithm primitives lowered through the pipeline |
| test_cudaq_arrays.py | 6 | Array parameter/return handling, incl. `qache` interaction |
| test_cudaq_arithmetic.py | 3 | `QuantumFloat` arithmetic and comparisons (classical/quantum RHS) |
| test_cudaq_algorithms.py | 3 | End-to-end algorithm execution on the CUDA-Q backend |
| `test_static_register_interpreter.py` | 17 | Static-register interpreter correctness independent of the Quake backend |

## Checklist
- [x] All tests pass locally


documentation/source/general/tutorial/CUDAQ.ipynb
```diff
@@ -0,0 +1,978 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "970a3ec8",
+   "metadata": {},
+   "source": [
+    "# Qrisp CUDA-Q Interface\n",
+    "\n",
+    "Are you utilizing CUDA-Q in your quantum workflows? Have you wondered if this would also be possible with Qrisp as a frontend? What cane first, the chicken or the egg?\n",
```
positr0nium: This gives me a bit of an awkward vibe... Maybe something like this?

_Ever wanted to combine Qrisp's expressive, Pythonic quantum programming with CUDA-Q's high-performance simulation and execution? You're in the right place.

This tutorial walks you through Qrisp's new CUDA-Q integration — from writing your first quantum kernel to building full hybrid variational workflows. Along the way you'll mix classical Python, quantum gates, and real-time control flow, all running natively on CUDA-Q.

Let's dive in!_

documentation/source/general/tutorial/CUDAQ.ipynb
```diff
@@ -0,0 +1,978 @@
...
+     "text": [
+      "Single run result: 3\n",
+      "10 shots: [3, 0, 0, 0, 3, 0, 0, 3, 3, 0]\n"
+     ]
+    }
+   ],
+   "source": [
+    "import cudaq\n",
+    "from qrisp import *\n",
+    "from qrisp.jasp.cudaq_interface import cudaq_kernel, FixedShapeNDArray\n",
```
positr0nium, line 45: I'm not sure if I like the import path. Being able to import everything "from qrisp" makes it convenient because you only have to memorize function name and not also the import path.

documentation/source/general/tutorial/CUDAQ.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   "id": "ff331ca9",
+   "metadata": {},
+   "source": [
+    "## Kernel Basics: Returns, Parameters, and Arrays\n",
+    "\n",
+    "Before diving into larger algorithms, it is worth covering three practical building blocks that come up in almost every non-trivial kernel:\n",
+    "\n",
+    "- **Multiple return values**: a kernel can return several measurement results at once, packed into a tuple.\n",
+    "- **Parametrized kernels**: scalar values (`int`, `float`, `bool`) and fixed-length arrays can be passed into a kernel as classical parameters, enabling runtime-configurable circuits without recompilation.\n",
+    "- **Static vs dynamic arrays**: arrays created inside the kernel from `np.array` are resolved at compile time and support NumPy transformations; arrays passed in as `FixedShapeNDArray` parameters or created with `jnp.array` are dynamic and are primarily used for indexed element access inside the kernel.\n",
```
positr0nium, line 298: The "inference through type annotation" mechanism seems hacky to me. It mixes the Python type annotation system, which usually doesn't really interfere with runtime logic into the actual runtime logic. Is the ``jnp.array`` path not enough? What do you think @PietropaoloFrisoni

documentation/source/general/tutorial/CUDAQ.ipynb
```diff
@@ -0,0 +1,978 @@
...
+   "cell_type": "code",
+   "execution_count": 12,
+   "id": "9f74f1bd",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Error defining kernel: Failed to compile Qrisp function 'dynamic_array_bad_kernel' to MLIR: Cannot lower this program to CUDA-Q because it still contains a ranked-tensor linalg.generic operation.\n",
```
positr0nium: It can be a bit confusing to users what the problem is here exactly. I think it would help if you describe precisely what doesn't work and that this is due to a shortcoming of the quake dialect.

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,727 @@
...
+# imports qrisp.jasp.cudaq_interface, so qrisp.jasp is not yet fully initialised here.
+from qrisp.jasp.jasp_expression import make_jaspr
+from qrisp.jasp.mlir.quake_lowering.jaspr_to_quake import jaspr_to_quake_mlir
+from qrisp.jasp.cudaq_interface.annotations import FixedShapeNDArray
+
+# ------------------------------------------------------------------ #
+# Platform-aware LLVM attribute defaults
+# ------------------------------------------------------------------ #
+
+_PLATFORM_DEFAULTS = {
```
positr0nium: It would help to have a comment what this actually does.

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,727 @@
...
+
+    The MLIR is expected to already have array parameters in
+    ``!cc.stdvec<T>`` form (as produced by pass4_array_to_stdvec).
+
+    The returned kernel is a first-class CUDA-Q kernel object and supports
+    all standard CUDA-Q execution patterns:
+
+    * ``kernel()`` — single-shot execution, returning the measurement result.
+    * ``cudaq.run(kernel, shots_count=N)`` — multi-shot sampling.
+    * ``cudaq.sample(kernel, shots_count=N)`` — histogram sampling
```
positr0nium: It would be nice to have an explanation what the difference is between the two? is it like terminal sampling vs sample in jaspify?

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,727 @@
...
+            return measure(qv)
+
+        mlir_str = str(jaspr_to_quake_mlir(make_jaspr(bell)()))
+        kernel = cudaq_kernel_from_mlir(mlir_str)
+
+        print(kernel())                          # single-shot, e.g. 0 or 3
+        print(cudaq.run(kernel, shots_count=100))
+
+    """
+    kernel = cudaq.make_kernel()
```
positr0nium: This function needs comments. Its not transparent on what is happening

documentation/source/general/tutorial/CUDAQ.ipynb
PietropaoloFrisoni: First of all, congrats @renezander90 as this is really an amazing PR (I would have paid myself to implement such a feature, as this is really exciting stuff).

Regarding the inference through type annotation, I was puzzled as well, even though I know that even in standard Python there are cases in which type annotations are not just a signal (such as in `@dataclass`, where they change indeed the behaviour at runtime).

However, I looked a bit [into the CUDA-Q documentation](https://nvidia.github.io/cuda-quantum/latest/using/examples/building_kernels.html) and had a few converstations with Gemini. It seems to me that even this framework uses a type-hint system to infer the behaviour at runtime. That is, the `FixedShapeNDArray` annotation serves exactly the role that the first concrete argument plays in `jax.jit` (it provides the shape and dtype needed to construct the abstract tracer). 

I don't have enough context to know whether this is indeed the case, but if I understand correctly, the annotation mechanism is necessary here because a compiler needs to know the type and shape of all values before it can generate typed IR (and `jnp.array` only creates a new array from values that we already have inside the kernel body).

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,727 @@
...
+        mlir_str = jaspr_to_quake_mlir(make_jaspr(bell)())
+        result = run_quake_mlir(mlir_str, shots=10)
+        print(result)  # e.g. [0, 0, 3, 0, 3, 0, 3, 3, 0, 0]
+
+    """
+    pykd = cudaq_kernel_from_mlir(mlir_str, execution_mode="run")
+    return cudaq.run(pykd, shots_count=shots)
+
+
+def sample_quake_mlir(mlir_str: str, shots: int = 100) -> dict[str, int]:
```
positr0nium: What is the purpose of these wrappers?

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,727 @@
...
+
+
+def cudaq_kernel(
+    func: Callable | None = None, execution_mode: Literal["run", "sample"] = "run"
+) -> PyKernelDecorator:
+    """
+    Decorator that compiles a Qrisp function to a native CUDA-Q kernel.
+
+    Mirrors ``@cudaq.kernel`` exactly: the decorated name is bound directly
+    to a ``PyKernelDecorator`` — compiled **eagerly at decoration time** —
```
positr0nium, line 89: Why do it eagerly at decoration time? What is wrong with doing it at tracing time?

documentation/source/general/tutorial/CUDAQ.ipynb
PietropaoloFrisoni: However, I don't think we should use an instance in type annotations, as in:

```
def static_index_kernel(angles: FixedShapeNDArray(float, 3)):
```

That is, as the name itself suggests, we should use type (classes). Even `pylance` flags this. This is something I have personally never seen and concerns me a bit to do this in a tutorial document :sweat_smile: 

P.S. According to Gemini, the cleanest fix  would be to implement `__class_getitem__` on `FixedShapeNDArray` so users write:

```
def kernel(angles: FixedShapeNDArray[float, 3]):
```

instead of:

```
def kernel(angles: FixedShapeNDArray(float, 3)):
```

Square bracket syntax is what Python's type system uses for generics, so Pylance would treat it as a subscripted generic rather than an instance.

renezander90: Thanks @PietropaoloFrisoni !

Yes, it is inspired by how native parametrised CUDA-Q kernels use type annotation: https://nvidia.github.io/cuda-quantum/latest/using/examples/building_kernels.html#parameterized-kernels

In our case, we additionally need the shape known statically at compile time for JAX tracing. Hence, the `FixedShapeNDArray` annotation. On could of course also ask the user to provide a dummy `np.array` directly (that is what is then created internally based on the annotation and used for tracing), but my impression was the the type annotation is more suitable for the user interface and more aligned with native CUDA-Q behaviour.

src/qrisp/jasp/cudaq_interface/quake_lowering/lowering_passes/jasp_to_quake/gate_mapping.py
```diff
@@ -0,0 +1,356 @@
...
+
+@dataclass(frozen=True)
+class GateInfo:
+    """Lowering information for a single Jasp gate.
+
+    For simple 1:1 mappings, set ``quake_gate`` and leave ``emit`` as None.
+    For multi-op decompositions, provide an ``emit`` callable.
+    """
+
+    #: Number of leading qubit operands that are *control* qubits.
```
positr0nium, line 60: I'm very confident that our mcx lowerings are more efficient than Quake's. We should not use this beyond just one control qubit

src/qrisp/jasp/mlir/quake_lowering/jasp_to_quake/gate_mapping.py
```diff
@@ -0,0 +1,356 @@
...
+    #: When provided, this is called *instead* of ``make_gate_op``.
+    emit: Optional[Callable] = field(default=None, repr=False, compare=False)
+
+
+# ===================================================================
+# Decomposition emitters
+# ===================================================================
+
+
+def _emit_sx(controls, params, targets):
```
positr0nium: It would help to have type annotations here

src/qrisp/jasp/mlir/quake_lowering/pass1_jasp_to_quake.py
```diff
@@ -0,0 +1,1056 @@
...
+   ``jasp.get_size``         ``quake.veq_size``
+   ``jasp.slice``            ``quake.subveq``
+   ``jasp.fuse``             ``quake.concat``
+   ``jasp.delete_qubits``    ``quake.dealloc``
+   ``jasp.quantum_gate``     ``quake.<gate>``  (dispatched via gate_mapping)
+   ``jasp.measure``          ``quake.mz`` + ``quake.discriminate``
+   ``jasp.reset``            ``quake.reset``
+   ``jasp.create_qk``        *(dropped; function attr added)*
+   ``jasp.consume_qk``       *(dropped)*
+   ``jasp.parity``           *(left in place / not lowered)*
```
positr0nium: This can be lowered to a XOR expression. Not necessarily in this PR though.

src/qrisp/jasp/cudaq_interface/quake_lowering/lowering_passes/jasp_to_quake/gate_mapping.py
renezander90: We use our `mcx` decompositions. In practice, this should not be used with control qubits since we run the `decompose_composite_gates(jaspr)`in the MLIR lowering pipeline.

documentation/source/general/tutorial/CUDAQ.ipynb
PietropaoloFrisoni: Thanks for the clarification. I update my comment above regarding the instance issue instead of type (not sure if you saw it)

src/qrisp/jasp/mlir/quake_lowering/pass1_jasp_to_quake.py
```diff
@@ -0,0 +1,1056 @@
...
+
+SCF ops are handled inside-out:
+
+1. Block arg types are updated first (``QubitArray → !quake.veq<?>``,
+   ``Qubit → !quake.ref``; QuantumState args are left for later removal).
+2. Inner ops are then processed recursively.
+3. Finally qst is stripped from the SCF op's structural components.
+"""
+
+
```
positr0nium: I think this file would profit from having it split into several smaller files. The amount of code and little helper functions makes it intransparent.

src/qrisp/jasp/mlir/quake_lowering/pass1_jasp_to_quake.py
```diff
@@ -0,0 +1,1056 @@
...
+2. **QuantumState threading elimination** – after the jasp-op rewrites,
+   ``!jasp.QuantumState`` values are removed from:
+
+   * op operands and results (done in-line during rewriting);
+   * ``scf.while`` / ``scf.if`` block arguments and yield/condition operands;
+   * function argument lists and return types.
+
+Algorithm
+---------
+Ops are processed in **forward order** within each block.  For each jasp op,
```
positr0nium: It might also be valuable to use the pass rewriting infrastructure provided by xDSL to make it more stable and avoid boilerplate (similar to how it is done in the "control flow fix" pass for Jasp mlir emission.

positr0nium (changes requested):
Really nice work @renezander90 ! It definitely still needs some maturing but the foundation stands! Having this bridge will for sure be very valuable!

documentation/source/general/tutorial/CUDAQ.ipynb
positr0nium: While going through the PR, I saw that this is because the ``cudaq_kernel`` decorator compiles things eagerly. If it was done lazily (i.e. after Jasp tracing), all the shape/type information would be available, or?
Since it is decorating Qrisp syntax, to me it would make more sense to have this behave like the rest of the Qrisp codebase.

.github/workflows/qrisp_test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest qiskit-aer qiskit-iqm pyscf stim cirq xdsl
+          pip install pytest qiskit-aer qiskit-iqm pyscf stim cirq xdsl cudaq==0.14.0
```
PietropaoloFrisoni: I don't know how long CUDA-Q takes to be installed, but we should really handle the installation of dependencies in a more scalable and efficient way (the more feature/dependencies we add, the more urgent this becomes). Of course this should not be done in this PR, but something to keep in mind during the next cycle.

PietropaoloFrisoni:
@renezander90 @positr0nium I just provided a couple of comments, but I can review this PR as well if required (please let me know if that is the case)

documentation/source/general/tutorial/CUDAQ.ipynb
renezander90: Yes, it is because the decorator compiles eagerly. I think it is beneficial not to have to recompile every time we execute the parametrised kernel. Consider, e.g., a situation like in the QAOA example in the tutorial.

positr0nium:
```python
from jax.api_util import shaped_abstractify
from jax.tree_util import tree_flatten

from qrisp import jaspify, QuantumVariable, measure

def lower_jasp_func_to_cudaq(func, *args):
    print("Lowering was called with args ", args)
    return jaspify(func)

def cudaq_kernel(func):

    compile_cache = {}

    def return_func(*args):

        args_flat, in_tree = tree_flatten(args)
        in_avals = tuple(shaped_abstractify(a) for a in args_flat)
        abstract_signature = hash((in_avals, in_tree))

        if abstract_signature not in compile_cache:
            compile_cache[abstract_signature] = lower_jasp_func_to_cudaq(func, *args)

        return compile_cache[abstract_signature](*args)

    return return_func
        
@cudaq_kernel
def main(a):
    qv = QuantumVariable(5)
    return measure(qv)

main(np.zeros(5))
main(np.zeros(5))
main(np.zeros(5))
main(np.zeros(4))
# Yields
# Lowering was called with args  (array([0., 0., 0., 0., 0.]),)
# Lowering was called with args  (array([0., 0., 0., 0.]),)   
```

renezander90: Thanks! This looks like a better solution :)

Edit: I might have been too excited. If we want it to behave like a native `cudaq.kernel`, e.g., like

`cudaq.run(main, np.array([1.,2.,3.]), shots_count=100)` 

we still need to compile eagerly. Otherwise, the user would have to call 

```
kernel = main(np.array([1.,2.,3.])) # or with some dummy value
cudaq.run(kernel, np.array([1.,2.,3.]), shots_count=100)
```

PietropaoloFrisoni:
@renezander90 Hi René,

Before merging this PR, could you please run `ruff/pylint` and, if possible (**highly** recommended), verify that the new code is covered by tests?

These checks are obviously not a substitute for a proper review, but they can help reduce the risk of introducing additional technical debt while we are actively trying to improve the quality and maintainability of qrisp.

As I already mentioned multiple times, I think these are the kinds of checks that we should eventually automate in CI, as is common practice in many open-source and industrial software projects. Thanks!

PietropaoloFrisoni:
Thanks @renezander90, I am super busy now but I will take a look as soon as possible. Can we request the review from @copilot in the meantime? Usually it is quite useful for my PRs

P.S. Looks like my comment triggered Copilot automatically somehow (I certainly did not request its review explicitly)

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
renezander90: To ensure that it works with `cudaq.run` **exactly** like a native kernel.

src/qrisp/jasp/cudaq_interface/quake_lowering/lowering_passes/jasp_to_quake/lower_jasp_to_quake.py
```diff
@@ -0,0 +1,464 @@
...
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: SliceOp, rewriter: PatternRewriter) -> None:
+        """Lower a JASP register slice to a Quake sub-register operation."""
+        arr = op.operands[0]
+        start_t = op.operands[1]
+        end_t = op.operands[2]
+
+        lo = _extract_scalar_for_rewriter(start_t, i64, rewriter)
+        hi_raw = _extract_scalar_for_rewriter(end_t, i64, rewriter)
+        hi_norm = _normalize_index_for_veq_rewriter(arr, hi_raw, rewriter)
```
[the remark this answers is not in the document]
renezander90: Fixed in 6f52c40cca8358e54fe5ad27b1229c660d47e138

src/qrisp/jasp/cudaq_interface/quake_lowering/lowering_passes/array_to_sequence.py
```diff
@@ -0,0 +1,132 @@
...
+            sequence_type = CcSequenceType(elem_type)
+            dyn_ptr_type = CcPtrType(CcArrayType(elem_type, _MLIR_DYNAMIC))
+
+            # Extract a dynamic-sized array pointer from the sequence.
+            data_op = CcSequenceDataOp(arg, dyn_ptr_type)
+            rewriter.insert_op(data_op, InsertPoint.at_start(block))
+
+            # Cast dynamic pointer back to the original static pointer type
+            cast_op = CcCastOp(data_op.result, old_type)
+            rewriter.insert_op(cast_op, InsertPoint.after(data_op))
```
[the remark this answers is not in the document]
renezander90: The true input shape is not known at compile time. Hence, this cannot be checked.

src/qrisp/jasp/cudaq_interface/cudaq_kernel.py
```diff
@@ -0,0 +1,253 @@
...
+                f"'{ann_name}' for parameter '{p.name}' of "
+                f"'{func.__name__}'. Supported: {_supported}."
+            )
+
+    jaspr = make_jaspr(func)(*dummy_args)
+
+    if register_size is not None:
+        new_jaspr = jaspr_to_static_register_jaspr(jaspr, register_size)
+    else:
+        new_jaspr = jaspr
```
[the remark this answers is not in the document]
renezander90: Fixed in a88bc1a679181d7bd700652bafe9d7cc056769b0

PR #555, opened 2026-05-05
Fix documentation build requirements

### LINKED ISSUE #463: Instructions to generate the docs are out of date
As stated in the title, the instructions to generate the docs are not up to date. `pip` dependency resolver runs into issues while installing dependencies. 

1. In `Qrisp/Documentation`, I first used `pip install -r requirements_doc.txt` 
2. Then I had to compare which dependencies were missing from the `txt` file and were mentioned in the first step of the `readme`. 

Image

4. I had to install `pandoc` because `make html` raised the following. 
```
nbsphinx.NotebookError: PandocMissing in general/tutorial/BE_tutorial/BE_vol1.ipynb:
    Pandoc wasn't found.
    Please check that pandoc is installed:
    https://pandoc.org/installing.html
```
Finally `make html` worked.

diehoq:
**Context:**
The documentation build instructions were out of date and no longer worked reliably in a fresh environment. Issue #463 tracks the dependency and setup problems around generating the docs, including Sphinx version compatibility and missing build-time tools in a clean environment.

**Description of the Change:**
This PR updates the documentation requirements and setup workflow so contributors can build the docs with:

```bash
cd documentation
make docs-install
make html
```

The changes pin Sphinx below 9 for `sphinx_toolbox.sidebar_links`, add the missing notebook and autodoc dependencies, add a `docs-install` target to the documentation Makefile to install Python requirements plus the `pandoc` system package, and clean up the README to point to that workflow. The pydata-sphinx-theme option name is also updated to avoid the current warning.

As an alternative, we could have modified `documentation/source/conf.py` to detect the bundled pandoc binary from a Python package and prepend it to `PATH` during Sphinx startup. This PR intentionally avoids that approach and keeps the docs configuration free of environment bootstrapping logic.

**Benefits:**
- The docs setup is now documented in one place through the Makefile-based workflow.
- The dependency set in `requirements_doc.txt` matches the actual Python-side build requirements.
- The docs configuration stays focused on Sphinx settings instead of runtime environment mutation.
- The new `docs-install` target is straightforward to run before `make html`.

**Possible Drawbacks:**
- `docs-install` uses `sudo apt install -y pandoc`, so it is Debian/Ubuntu-oriented and assumes users have sudo access.
- The workflow now depends on a system package manager step rather than solving pandoc entirely inside Python packaging.
- I could not execute the full end-to-end build under this exact workflow in this environment because `pandoc` is not installed here, but the Make target itself was validated and the previous docs build passed before switching the installation path.

**Related GitHub Issues:**
Closes #463


documentation/build_docs_temp_env.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+REPO_ROOT=$(dirname "$SCRIPT_DIR")
+DOCS_DIR="$SCRIPT_DIR"
+
+TMPDIR=$(mktemp -d /tmp/qrisp-docs-XXXXXX)
+VENV_DIR="$TMPDIR/.venv"
```
purva-thakre: question: If I am already in a virtual environment, I would not need this step, right? 

Using the `sh` is an interesting approach. However, I am worried this might be too specific to the docs build on your device. When we finally get to the stage where the docs can be built and viewed live on Read The Docs, something like this file will be extremely helpful in writing the `yml` file for the docs build.

documentation/README.md
```diff
@@ -27,5 +27,15 @@ Of further interest is the conf.py, check out the recommended tutorial for its f
...
 Opening the index.html file in the browser displays the build version of the documentation page.
 
+# Building docs in a fresh temporary venv
+Run the helper script from the documentation folder or by absolute path:
+
+```bash
+    ./build_docs_temp_env.sh
+```
+
+The script creates a temporary virtual environment, installs the documentation
```
purva-thakre: Same as my other comment, I need to read up a bit on using the helper script. In my mind, the generic workflow is

- activate virtual environment
- navigate to the docs directory
- use pip install on the docs requirement.txt file, then install pandoc separately
- run `make html` in the docs folder
- go to the html build in the local folder

Since you have moved all the requirements to the `txt` file, you could get rid of the `pip install` instructions on L6-L10. Sorry, I can't make direct comments on it in this PR as you haven't made a change to those lines.

purva-thakre (changes requested):

documentation/build_docs_temp_env.sh
diehoq: Yes, this was actually to create a new temporary and separate virtual environment to build the doc. It is just a suggestion and I can remove this file for now. We need to deal with pandoc not being on PATH then.
I see two alternatives then:
1. We make the user install it manually, as you did
2. We add a small helper in conf.py which makes the binary installed via pypandoc-binary discoverable on the doc build

purva-thakre: > We need to deal with pandoc not being on PATH then.

Another option is to add a makefile. Add a target called `install-pandoc` such that the first step is pip install from the `requirement,.txt` file and then use `make install-pandoc`. Thoughts?

This will hopefully be a temporary fix. I think I have found a way to not add `pandoc` to the docs build pipeline. https://github.com/eclipse-qrisp/Qrisp/issues/558

diehoq: That could be also a valid option, but it is as much overhead as letting the user install pandoc directly, no? I don't have anything against it, though

purva-thakre: > but it is as much overhead as letting the user install pandoc directly,

True. As I said, I am hoping this is a temporary fix until someone works on the linked issue. Thanks to Claude, I created a target in the makefile that has both steps in one go. 

```
docs-install:
	pip install -r requirements_doc.txt
	sudo apt install -y pandoc
```
With this, the steps are:

- cd docs
- make docs-install
- make html

diehoq: Ok I'll push the fix later today!

documentation/README.md
```diff
@@ -23,8 +17,11 @@ Each page needs to be integrated via index.rst file.\
 Of further interest is the conf.py, check out the recommended tutorial for its functionality.
 
 # Creating the actual html page
+    make docs-install
     make html
```
purva-thakre, line 22:
```suggestion
```bash
    make docs-install
    make html
 ```
```

purva-thakre (approved):
Thanks! LGTM. 

Minor nit: last line of `documentation/README.md` appears to be a leftover from something else. Can you please remove it?

diehoq:
> Thanks! LGTM.
> 
> Minor nit: last line of `documentation/README.md` appears to be a leftover from something else. Can you please remove it?

You mean L26? I think that tells you how to visualize the local doc build in the browser. Could be phrased better maybe.

documentation/README.md
```diff
@@ -23,8 +17,12 @@ Each page needs to be integrated via index.rst file.\
 Of further interest is the conf.py, check out the recommended tutorial for its functionality.
 
 # Creating the actual html page
+```bash
+    make docs-install
     make html
-The make html command generates the html pages inside the build/html folder.\
```
purva-thakre, line 27: the ``` after `make html` is missing. IDK why it didn't show up in the suggestion. 🤷🏾‍♀️ Sorry.

purva-thakre:
> You mean L26? I think that tells you how to visualize the local doc build in the browser. Could be phrased better maybe.

No. I meant L29. It is unrelated to the changes you made which is why I can't suggest a change to remove it. 
https://github.com/diehoq/Qrisp/blob/13a6b68576a71da49ac51e6961599e0721a28f5d/documentation/README.md?plain=1#L29

diehoq:
> No. I meant L29. It is unrelated to the changes you made which is why I can't suggest a change to remove it.
> https://github.com/diehoq/Qrisp/blob/13a6b68576a71da49ac51e6961599e0721a28f5d/documentation/README.md?plain=1#L29

Wow completely missed line! removing it

PR #564, opened 2026-05-12
Passmanager infra

positr0nium:
## Description

Introduces a structured pass manager infrastructure (`qrisp.circuit.pass_management`) — a formal framework for organizing, composing, and applying quantum circuit transformation passes. Includes a `CircuitPass` decorator with type checking, a `PassManager` pipeline, 14 built-in circuit transformation passes, full Sphinx documentation, and comprehensive tests.

## Related Issues

Closes #
Related to #

## Type of Change

- [x] Feature (new functionality)
- [x] Documentation
- [ ] Change Request (modification of existing functionality)
- [ ] Bug Fix
- [ ] Refactoring (no behavior change)
- [ ] Performance improvement
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Added `CircuitPass` decorator (`circuit_pass.py`) to wrap `QuantumCircuit → QuantumCircuit` functions with runtime type validation.
- Added `PassManager` pipeline (`pass_manager.py`) — ordered collection of `CircuitPass` instances applied via `run()`, with `add_pass`, `insert_pass`, `remove_pass`, `clear`, `verify`, `__iadd__`, and iteration.
- Added 14 built-in passes in `passes/`:
  - `arange_swaps` — reorder SWAP qubits for better cancellation
  - `cancel_inverses` — cancel adjacent gate–inverse pairs via DAG analysis
  - `cancel_zero_controls` — remove gates controlled on `|0⟩`
  - `combine_single_qubit_gates` — fuse consecutive single-qubit gates
  - `commute_swaps` — commute single-qubit ops past SWAP gates
  - `convert_to_cz` — convert CX, CY, SWAP to CZ-based decompositions
  - `decompose` — recursively dissolve synthesized gates (configurable level + predicate)
  - `gray_synth_toffoli` / `is_toffoli` — Gray-synthesis Toffoli decomposition and predicate
  - `manual_layout` — re-index qubits via user-supplied mapping
  - `remove_barriers` — strip barrier instructions
  - `resolve_swaps` — fold SWAPs into qubit index remapping
  - `reverse_parallelize` — reverse-order parallelization for deeper optimization
  - `visualize` — no-op debug pass that prints the circuit
- Added Sphinx documentation pages for `CircuitPass`, `PassManager`, and all passes with examples inline in docstrings.
- Added 15 test files covering every built-in pass, `CircuitPass`, and `PassManager`.

## How was it tested?

| Test-ID | Status |
|---------|--------|
[...]

Issue #563, opened 2026-05-11
Adding FOQCS-LCU BlockEncoding method support


Redperij:
### Priority

Medium

### Responsible

@MSY2K , and @Redperij

### Target Milestone

v0.9, Support for Heisenberg and Spin Glass model examples with 🦊-LCU

### Goal

The goal is to introduce the [FOQCS-LCU](https://arxiv.org/abs/2507.20887) block encoding method (#448) and transfer its [Qiskit implementation](https://github.com/QuantumComputingLab/foqcs-lcu) to Qrisp. The bare minimum is to implement the optimal Heisenberg model and the more general spin-glass model. 🦊-LCU handles certain LCUs (like the Heisenberg and Spin-Glass model) more effectively by exploiting their structure, but doing so requires a custom PREP block for each model.

### Sub-Issues

- [x] #567 
- [x] #565 
- [x] #566 

### Dependencies & Order

Depends on the PRs (must be merged in top-down order):
- [x] #574 
- [x] #564 
- [x] #572 
- [x] #597 

### Definition of Done

- [x] Heisenberg model implemented and tested
- [x] Spin-glass model implemented and tested
- [x] BlockEncoding class methods behaviour with 🦊-LCU is verified
- [x] Tests are complete
- [x] All sub-issues closed
- [x] Documentation updated
- [x] Peer-reviewed

Potential future improvements using "[Practical block encodings of matrix polynomials that can also be trivially controlled](https://arxiv.org/abs/2601.18767)"?

Redperij:
<img width="640" height="434" alt="Image" src="https://github.com/user-attachments/assets/db2632f0-15bb-42a3-a35f-d32fe10f3893" />

PR #572, opened 2026-05-13
Dicke states expansion

### LINKED ISSUE #565: Dicke state cannot be used under JASP
### Priority

Medium

### Responsible

@Redperij

### Target Version / Milestone

_No response_

### Type of Change

API change (non-breaking)

### Breaking Change?

No

### Short Description

Dicke state preparation function cannot run under JASP due to the internal `len(qv)` call. Suggestion: making the `len(qv)` call optional depending on whether the Quantum variable length was specified using a new parameter.

### Current Behavior

As in short Description, JASP throws an error due to the `n = len(qv)` call.
```python
def dicke_state(qv, k, num_qubits: int = 0):
"""
...
"""
    n = len(qv)
    for index2 in reversed(range(k + 1, n + 1)):
        split_cycle_shift(qv, index2, k)

    for index in reversed(range(2, k + 1)):
        split_cycle_shift(qv, index, index - 1)
```

### Desired Behavior

Suggestion is to add `num_qubits` parameter or something alike, to allow passing the number of qubits manually.
If specified, the parameter value is trusted and used as a length of quantum variable. If left unspecified or 0, (default behaviour), the function will behave as before the change - using `len(qv)`
```python
def dicke_state(qv, k, num_qubits: int = 0):
"""
...
    Parameters
    ----------
    ...
    num_qubits: Int
        Number of passed qubits, used instead of `len(qv)` call if specified to any value other than 0. Default is 0
...
"""
    if num_qubits == 0:
        n = len(qv)
    else:
        n = num_qubits

    for index2 in reversed(range(k + 1, n + 1)):
        split_cycle_shift(qv, index2, k)

    for index in reversed(range(2, k + 1)):
        split_cycle_shift(qv, index, index - 1)

```

### Rationale / Justification

Non-compact FOQCS-LCU implementation from #563 requires using Dicke state preparation under JASP to be compatible with BlockEncoding class functionality. The suggested change will not affect the existing code and will allow to use the `dicke_state()` function without relying on its internal `len(qv)` call.

### Migration Path

N/A

### Affected Components

- `qrisp.alg_primitives.dicke_state_prep`

### Acceptance Criteria

- [x] Unmodified function calls behaviour did not change
- [x] Dicke state can be used under JASP with passed `num_qubits` parameter.
- [x] Change is discussed and any required modifications made
- [ ] Change is accepted
- [ ] Peer-reviewed

### Means of Verification

Tested in context of FOQCS-LCU #563.
Special custom unit tests can be created if necessary.

### Dependencies

Was needed for non-compact implementation of Heisenberg model from #563, currently is not necessary.

### Definition of Done

- [x] Code implemented and peer-reviewed
- [x] All existing tests updated / still passing
- [x] New regression tests added
- [x] Migration path documented (if breaking)
- [x] Deprecation warnings added (if applicable)
- [x] Documentation updated
- [ ] Changelog entry added

### LINKED ISSUE #567: Unbalanced W state preparation
### Priority

Medium

### Responsible

@Redperij

### Short Description

Qrisp lacks an unbalanced version of `dicke_state()` preparation routine. Unbalanced W state allows to prepare unbalanced Dicke state with one excitation.

### Long Description

Linear chain unbalanced W state preparation algorithm version provided by @positr0nium , and modified by @Redperij is given below:

```python
def unbalanced_W_state(
    qv: QuantumVariable | Sequence[Qubit],
    amplitudes: list,
    num_qubits: int = 0,
    reversed: bool = False
) -> None:
    r"""
    Prepare a generalized W state (unbalanced Dicke state of Hamming weight 1)
    on the given :ref:`QuantumVariable`.

    The resulting quantum state is

    .. math::

        |\psi\rangle \;=\; \sum_{i=0}^{n-1} a_i \,|e_i\rangle

    where :math:`|e_i\rangle` is the computational basis state with a single
    ``1`` at position `i`, and :math:`a_i` are the (possibly complex)
    amplitudes given by ``amplitudes``. The input array is automatically
    normalized so that :math:`\sum_i |a_i|^2 = 1`.

    Parameters
    ----------
    qv : QuantumVariable
        A freshly allocated :ref:`QuantumVariable` in the
        :math:`|0\dots0\rangle` state whose size matches ``len(amplitudes)``.
    amplitudes : array_like
        A 1-D sequence of complex (or real) target amplitudes, one per qubit.
        Its length must equal ``qv.size``.
    num_qubits : int
        Number of passed qubits, used instead of `len(qv)` call if specified to any value other than 0. Default is 0
    reversed : bool, optional
        If ``True``, reverse the order of the received amplitudes before
        preparing the state. Default is ``False``

    Raises
    ------
    ValueError
        If ``len(amplitudes) != qv.size`` or if the amplitude vector is zero.

    Notes
    -----
    **Algorithm.**
    The circuit distributes a single excitation across all qubits using a
    linear chain of ``XXYY`` gates:

    1. Apply ``X`` to qubit 0, producing :math:`|10\dots0\rangle`.
    2. For each qubit :math:`i = 0, \dots, n{-}2`:

       a. Compute :math:`\theta = 2\arccos(|a_i|\,/\,r_i)` where :math:`r_i`
          is the remaining (undistributed) amplitude magnitude.
       b. Apply ``XXYY(θ, π/2)`` on qubits :math:`(i,\, i{+}1)`.  In the
          single-excitation subspace this acts as a parametrized partial swap,
          leaving magnitude :math:`|a_i|` on qubit :math:`i` and passing the
          rest to qubit :math:`i{+}1`.
       c. Apply a phase gate :math:`P(\arg a_i)` on qubit :math:`i` to imprint
          the correct complex phase.

    3. Apply :math:`P(\arg a_{n-1})` on the last qubit.

    **Resources.**
    The circuit uses :math:`n{-}1` ``XXYY`` gates (each decomposable into
    2 CNOTs + single-qubit rotations) and :math:`n` phase gates, yielding
    :math:`\mathcal{O}(n)` depth and gate count.

    Examples
    --------
    >>> import numpy as np
    >>> from qrisp import QuantumVariable
    >>> a = np.array([1j, 2, 3, 4])
    >>> qv = QuantumVariable(4)
    >>> unbalanced_W_state(qv, a)
    >>> print(qv.qs.statevector())
    """
    if num_qubits == 0:
        n = len(qv)
    else:
        n = num_qubits

    a = np.asarray(amplitudes, dtype=complex)

    if reversed:
        a = a[::-1]

    if len(a) != n:
        raise ValueError(
            f"Length of amplitudes ({len(a)}) must match qv.size ({n})."
        )

    # Normalize so that <a|a> = 1
    norm = np.sqrt(np.vdot(a, a).real)
    if norm < 1e-15:
        raise ValueError("Amplitude vector must be non-zero.")
    a = a / norm
    abs_a = np.abs(a)

    # --- Step 1: place the single excitation on qubit 0  --->  |10...0>
    x(qv[0])

    # --- Step 2: redistribute amplitude along the qubit chain
    # `remaining` tracks the magnitude still carried by the "active" qubit
    # (the one that has not yet been peeled off).
    remaining = 1.0
    for i in range(n - 1):
        # Choose θ so that cos(θ/2) = |a_i| / remaining,
        # i.e. qubit i retains exactly magnitude |a_i|.
        theta = 2 * np.arccos(np.clip(abs_a[i] / remaining, -1.0, 1.0))

        # XXYY(θ, π/2) performs a parametrized partial swap in the
        # single-excitation subspace {|01>, |10>}:
        #   |10> -> cos(θ/2)|10> - sin(θ/2)|01>
        xxyy(theta, np.pi / 2, qv[i], qv[i + 1])

        # Update the undistributed amplitude magnitude for the next step
        remaining *= np.sin(theta / 2)

        # Imprint the complex phase of a_i onto qubit i
        p(np.angle(a[i]), qv[i])

    # --- Step 3: imprint the phase on the last qubit
    p(np.angle(a[-1]), qv[-1])
```

As was pointed out, there is a log depth version possible through distributing the excitation through a tree structure. It requires additional work to implement and verify and may not be a suitable modification before v0.9 milestone.

### Means of Verification

Extensively tested in context of FOQCS-LCU #563 .
Special custom unit tests can be created if necessary.

Redperij:
## Description

This PR enables dicke_state() function compatibility with jasp and introduces unbalanced W state preparation routine, (unbalanced Dicke state with one excitation).

## Related Issues

Closes #565 
Closes #567 
Related to #563 

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

N/A

## What was changed?

- Added `qrisp.alg_primitives.unbalanced_w_state.py` with jasp-compatible unbalanced_W_state() implementation.
- Reworked dicke_state() function from `qrisp.alg_primitives.dicke_state_prep.py` to be jasp-compatible
- Added unit tests for unbalanced_W_state() and dicke_state() in `algorithms_tests.test_dicke_states.py`.

### dicke_state()
| Test-ID | Status | Description |
| ------------- | ------------- | ------------- |
| test_dicke_state_balanced  | ✅ | Tests function call produced state vectors against the manually prepared ones. |
| test_dicke_state_balanced_jasp  | ✅ | Tests that dicke_state call works under jasp and produces the expected distribution. |
| test_dicke_state_balanced_jasp_inverse | ✅ | Tests that dicke_state works under inverse and jasp environments simultaneously. |
| test_dicke_state_k| ✅ | Tests that dicke_state produces correct measurements with k > 1 number of excitations. |

### unbalanced_W_state()
| Test-ID | Status | Description |
| ------------- | ------------- | ------------- |
| test_unbalanced_W_state | ✅ | Tests function call produced state vectors against the manually prepared ones. Failed the compilation due to the bug fixed in #564. |
| test_unbalanced_W_state_measurement | ✅ | Tests that unbalanced_W_state produces correct measurements. |
| test_unbalanced_W_state_jasp | ✅ | Verifies that unbalanced_W_state functions under jasp, checks the resulting measurements. |
| test_unbalanced_W_state_jasp_inverse | ✅ | Tests that unbalanced_W_state works under inverse and jasp environments simultaneously. |
| test_unbalanced_W_state_fail_len_check | ✅ | Tests that in static environment amplitudes vector sanity check is executed. |
| test_unbalanced_W_state_fail_zero_vector | ✅ | Tests that in static environment amplitudes vector sanity check is executed. |
| test_unbalanced_W_state_one_qubit_jasp | ✅ | Tests that a single qubit unbalanced_W_state preparation produces single X and phase gate under jasp environment. |

## Checklist


src/qrisp/alg_primitives/dicke_state_prep.py
```diff
@@ -21,7 +21,7 @@
 from qrisp.core import cx, ry
```
renezander90, line 25:
```suggestion
from qrisp.jasp import jlen, jrange
```

src/qrisp/alg_primitives/dicke_state_prep.py
renezander90: - Remove unnecessary ``index_range`` list.
- Remove if/else distinction. (highIndex case can be treated outside of the loop)
- Use jax.numpy instead of numpy (index will be Tracer in Jasp)

src/qrisp/alg_primitives/dicke_state_prep.py
```diff
@@ -31,7 +31,9 @@ def dicke_state(qv, k):
     qv : QuantumVariable
         Initial quantum variable to be prepared. Has to be in target subspace.
     k : Int
-        The Hamming weight (i.e. number of "ones") for the desired dicke state
+        The Hamming weight (i.e. number of "ones") for the desired dicke state.
+    num_qubits: Int
+        Number of passed qubits, used instead of `len(qv)` call if specified to any value other than 0. Default is 0
```
renezander90: Thank you, while this solution works;
- it seems unnecessary to expose ``num_qubits`` to the user
- if ``num_qubits`` is passed through several function calls it might still become a Tracer in Jasp
- it is not optimal from a compilation performance perspective, since the loops will be unrolled in Jasp

Please note that ``jlen(qv)`` will work in both, static and dynamic (Jasp) modes. However, in Jasp, the result will be a Tracer, hence note work with standard Python loops (``range`` -> ``jrange``). 
``jrange`` will not work with ``reversed`` and the (unnecessary) static list ``index_range = [highIndex - i for i in range(lowIndex)]``. Hence, some minor rewriting of the code would be necessary (see hints below :)).

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Parameters
+    ----------
+    qv : QuantumVariable
+        A freshly allocated :ref:`QuantumVariable` in the
+        :math:`|0\dots0\rangle` state whose size matches ``len(amplitudes)``.
+    amplitudes : array_like
+        A 1-D sequence of complex (or real) target amplitudes, one per qubit.
+        Its length must equal ``qv.size``.
+    num_qubits : int
```
renezander90: See comments for dicke state preparation. A ``num_qubits`` parameter would only be need if you want to allow partial W-states as well, i.e., only the first k<n qubits are considered. If this is the intention, it should be noted in the documentation.

renezander90 (changes requested):

Redperij:
Thank you.
Rewritten both functions to use jax.numpy. It is a pity that I had to remove sanity checks from unbalanced_W_state, since it is nice to know if passed amplitude array length matches the number of qubits. Both functions (now only unbalanced_W_state) in context of FOQCS-LCU are often used on certain parts of ancillary qubits, so it is easy to make a mistake.

renezander90:
> Thank you. Rewritten both functions to use jax.numpy. It is a pity that I had to remove sanity checks from unbalanced_W_state, since it is nice to know if passed amplitude array length matches the number of qubits. Both functions (now only unbalanced_W_state) in context of FOQCS-LCU are often used on certain parts of ancillary qubits, so it is easy to make a mistake.

Thank you:) You can guard the checks with ``if check_for_tracing_mode() and a!=b: raise ...`` (use: ``from qrisp.jasp import check_for_tracing_mode``. Then they will still work in the static pipeline.

Redperij:
Thanks, I will add the check and test for it shortly.

Redperij:
`unbalanced_W_state` method requires a fix from #564 , specifically, from commit [cccfe04](https://github.com/eclipse-qrisp/Qrisp/commit/cccfe04ea509f0e08245142df29f783809e458b5)

Tests `test_unbalanced_W_state` and `test_unbalanced_W_state_trailing_zeroes` assume the fix is applied and fail on current branch.

renezander90:
Please note that in order to be shown in the documentation, you need to create a corresponding `.rst` file here: https://github.com/eclipse-qrisp/Qrisp/tree/main/documentation/source/reference/Primitives and add a reference in the `index.rst` here: https://github.com/eclipse-qrisp/Qrisp/blob/main/documentation/source/reference/Primitives/index.rst

Also see: https://github.com/eclipse-qrisp/Qrisp/blob/main/documentation/source/reference/Development%20Guide/documentation.rst

Best practice is to build the documentation locally to ensure that it is displayed as intended and does not introduce any new warnings when building the documentation.

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,159 @@
...
+
+import numpy as np
+import jax.numpy as jnp
+from qrisp import QuantumVariable, Qubit, x, xxyy, p
+from qrisp.jasp import jrange, check_for_tracing_mode
+from collections.abc import Sequence
+
+def unbalanced_W_state(
+    qv: QuantumVariable | Sequence[Qubit],
+    amplitudes: list,
```
renezander90: Should be `NDArrayLike`: https://github.com/eclipse-qrisp/Qrisp/blob/69cb731bd8ae1f2f963939b2819ebf779219b724/src/qrisp/typing.py#L72

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,159 @@
...
+    reversed: bool = False
```
renezander90: Do we need this `reversed` argument? It could also be delegated to the user to simply provide the reversed array.

tests/algorithms_tests/test_dicke_states.py
```diff
@@ -0,0 +1,302 @@
...
+    prepared_sv = qv.qs.compile().statevector_array()
+
+    # Manual expected state:
+    # |D^3_1> = (|001> + |010> + |100>) / sqrt(3)
+    expected_sv = np.zeros(2 ** n, dtype=complex)
+    amp = 1 / np.sqrt(n)
+    for i in range(n):
+        expected_sv[2 ** i] = amp
+
+    print(f"Prepared statevector:\n{prepared_sv}")
```
renezander90: Test should not contain `print` statements

tests/algorithms_tests/test_dicke_states.py
```diff
@@ -0,0 +1,302 @@
...
+        qv = QuantumVariable(n)
+        x(qv[n - 1])
+        dicke_state(qv, k)
+        with invert():
+            x(qv[n - 1])
+            dicke_state(qv, k)
+        return qv
+    
+    result = main()
+    #prepared_sv = result.qs.compile().statevector_array()
```
renezander90: Remove

tests/algorithms_tests/test_dicke_states.py
```diff
@@ -0,0 +1,302 @@
...
+    # Prepare balanced Dicke state
+    @terminal_sampling
+    def main():
+        qv = QuantumVariable(n)
+        x(qv[n - 1])
+        dicke_state(qv, k)
+        return qv
+    
+    result = main()
+    #prepared_sv = result.qs.compile().statevector_array()
```
renezander90: Remove

src/qrisp/alg_primitives/unbalanced_w_state.py
Redperij: I found it useful to have a quick switch for reversing the amplitudes array. Amplitude ordering commonly causes confusion due to the different qubit ordering conventions across frameworks / algorithms.

So, `reversed` is not necessarily needed, but proved to be convenient. It's up to preference.

Redperij:
I noticed that in https://github.com/eclipse-qrisp/Qrisp/blob/main/documentation/source/reference/Primitives/index.rst
`cyclic_shift` and `iterable_permutation` are present in the navigation tree but do not have entries in the primitives table. Is this intentional?

renezander90:
> I noticed that in https://github.com/eclipse-qrisp/Qrisp/blob/main/documentation/source/reference/Primitives/index.rst `cyclic_shift` and `iterable_permutation` are present in the navigation tree but do not have entries in the primitives table. Is this intentional?

Bildschirmfoto 2026-06-07 um 21 44 23

They are all bundled in one entry.

src/qrisp/alg_primitives/unbalanced_w_state.py
renezander90: I think there was a misunderstanding. I meant `amplitudes` should be of type `NDArrayLike`.

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,159 @@
...
+
+    **Resources.**
+    The circuit uses :math:`n{-}1` ``XXYY`` gates (each decomposable into
+    2 CNOTs + single-qubit rotations) and :math:`n` phase gates, yielding
+    :math:`\mathcal{O}(n)` depth and gate count.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from qrisp import QuantumVariable
```
renezander90:
```suggestion
    >>> from qrisp import QuantumVariable, unbalanced_W_state
```

src/qrisp/alg_primitives/dicke_state_prep.py
```diff
@@ -16,9 +16,10 @@
...
 """
 
-import numpy as np
+import jax.numpy as jnp
 
 from qrisp.core import cx, ry
+from qrisp.jasp import jlen, jrange
 
 
 def dicke_state(qv, k):
```
renezander90:
```suggestion
def dicke_state(qv: QuantumVariable | Sequence[Qubit], k: int) -> None:
```

We should fix the typing here:) This would require the corresponding imports.

src/qrisp/alg_primitives/dicke_state_prep.py
```diff
@@ -71,21 +74,21 @@ def split_cycle_shift(qv, highIndex, lowIndex):
     lowIndex : Int
```
renezander90:
```suggestion
    lowIndex : int
```

src/qrisp/alg_primitives/unbalanced_w_state.py
renezander90: After consulting with AI: this argument should be removed.

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,161 @@
...
+    >>> print(qv.qs.statevector())
+    """
+    a = jnp.asarray(amplitudes, dtype=complex)
+
+    if reversed:
+        a = a[::-1]
+
+    n = a.shape[0] # Use the static shape of amplitudes
+
+    if not check_for_tracing_mode():
```
renezander90, line 101: Needs tests that errors are raised; use `pytest.raises`

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Notes
+    -----
+    **Algorithm.**
+    The circuit distributes a single excitation across all qubits using a
+    linear chain of ``XXYY`` gates:
+
+    0. Precompute all required :math:`\theta` angles using
+       :math:`r_i = \sqrt{ \sum_{ j = i }^{ n - 1 }{ |a_j| ^ 2 } }`
+       and :math:`\theta = 2\arccos(|a_i|\,/\,r_i)`.
```
renezander90: Should this be theta_i ?

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,161 @@
...
+    1. Apply ``X`` to qubit 0, producing :math:`|10\dots0\rangle`.
+    2. For each qubit :math:`i = 0, \dots, n{-}2`:
+
+       a. Compute :math:`\theta = 2\arccos(|a_i|\,/\,r_i)` where :math:`r_i`
```
renezander90: I think we already do this in step 0?

tests/algorithms_tests/test_dicke_states.py
```diff
@@ -0,0 +1,269 @@
...
+from qrisp.alg_primitives.unbalanced_w_state import unbalanced_W_state
+from qrisp.alg_primitives.dicke_state_prep import dicke_state
+from qrisp.jasp import terminal_sampling
+from qrisp.environments import control, invert
+
+#############################################################
+##################### Dicke state tests #####################
+#############################################################
+
+def test_dicke_state_balanced():
```
renezander90: There should be tests also for k>1 to ensure there is no regression. Possibly use ` pytest.mark.parametrize`. It could be challenging to initialise the amp vector. One could also use `.get_measurement()` and check that all results have the correct number of ones in binary + that the number of distinct measurement results is n choose k.

renezander90 (changes requested):
Thanks for the contribution! We're making great progress, just a few more items to wrap up. :)

src/qrisp/alg_primitives/unbalanced_w_state.py
Redperij: I think that 2 existing tests in `test_dicke_states.py` already cover that: `test_unbalanced_W_state_fail_len_check` and `test_unbalanced_W_state_fail_zero_vector`.
```python
def test_unbalanced_W_state_fail_len_check():
    n = 1 # Number of qubits
    amps = np.array([0.25 + 0.2j, 2, 3, 4, 5, 6, 7], dtype=complex)

    # Prepare unbalanced Dicke state
    qv = QuantumVariable(n)
    with pytest.raises(ValueError) as exc_info:
        unbalanced_W_state(qv, amps)

    assert f"Length of amplitudes" in str(exc_info.value)

def test_unbalanced_W_state_fail_zero_vector():
    n = 3 # Number of qubits
    amps = np.array([0, 0, 0], dtype=complex)

    # Prepare unbalanced Dicke state
    qv = QuantumVariable(n)
    with pytest.raises(ValueError) as exc_info:
        unbalanced_W_state(qv, amps)

    assert f"Amplitude vector must be non-zero." in str(exc_info.value)
```

renezander90: Yes, indeed, I missed them.

renezander90:
The `test_dicke_state.py` should be moved to the `primitives_tests` folder.

src/qrisp/alg_primitives/unbalanced_w_state.py
```diff
@@ -0,0 +1,154 @@
...
+
+import numpy as np
+import jax.numpy as jnp
+from qrisp.core import QuantumVariable, x, xxyy, p
+from qrisp.circuit import Qubit
+from qrisp.typing import NDArrayLike
+from qrisp.jasp import jrange, check_for_tracing_mode
+from collections.abc import Sequence
+
+def unbalanced_W_state(
```
renezander90: Function name should be `unbalanced_w_state` (lowercase)

renezander90 (approved):

PR #587, opened 2026-05-23
Enhancing Grover

### LINKED ISSUE #586: Unjustified Exception in grovers_alg
### Priority

High

### Responsible

_No response_

### Qrisp Version

0.8.2

### Python Version

3.11.15

### Operating System

macOS

### Short Description

`grovers_alg` throws an unjustified `Exception`, when the oracle acts on a QuantumVariable that is not part of the `qv_list`.

### Steps to Reproduce

```
from qrisp import *
from qrisp.algorithms.grover import grovers_alg
import numpy as np


def prepare_oracle(qv, forbidden_value):

    @auto_uncompute
    def oracle(qv, phase=np.pi):

        flag = qv != forbidden_value
        p(phase, flag)

    grovers_alg(qv, oracle, winner_state_amount=2**qv.size - 1, exact=True)


a = QuantumFloat(2)
h(a)
b = QuantumFloat(2)
prepare_oracle(b, forbidden_value=a)

multi_measurement([a, b])
```

> Exception: Applied oracle introducing new QuantumVariables without uncomputing/deleting

### Expected Behavior

> {(0, 1): 0.08333333333333333,
 (0, 2): 0.08333333333333333,
 (0, 3): 0.08333333333333333,
 (1, 0): 0.08333333333333333,
 (1, 2): 0.08333333333333333,
 (1, 3): 0.08333333333333333,
 (2, 0): 0.08333333333333333,
 (2, 1): 0.08333333333333333,
 (2, 3): 0.08333333333333333,
 (3, 0): 0.08333333333333333,
 (3, 1): 0.08333333333333333,
 (3, 2): 0.08333333333333333}

### Actual Behavior

> Exception: Applied oracle introducing new QuantumVariables without uncomputing/deleting

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive |             |                 |
| T-002   | Negative |             |                 |
| T-003   | Boundary |             |                 |
| T-004   | Boundary |             |                 |

### Additional Context

Root cause: https://github.com/eclipse-qrisp/Qrisp/blob/8fed21134f626a2090fe22f975bddb202005de3e/src/qrisp/algorithms/grover/grover_tools.py#L529

### Definition of Done

- [x] Root cause identified
- [ ] Regression test added
- [ ] Changelog entry added

renezander90:
## Description

Refactors `grovers_alg` and `diffuser` with type annotations, a mutable-default-argument fix (`kwargs={}`→`None`), a simplified iteration calculation for exact Grover's, and improved docstrings. Renames the `qv_list` parameter to `args`. Replaces the `Exception` for missing `winner_state_amount` with a proper `ValueError`. Expands the single monolithic test into a focused, well-named test suite with two additional cases (all-states-tagged and oracle with side effects). Deactivates the oracle-uncompute check (`if qv_amount != len(qs.qv_list)`) related to #586.

## Related Issues

Closes #586
Related to

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **grover_tools.py**:
  - Renamed `qv_list` parameter to `args` in `grovers_alg` to avoid implying that input must be a list, and for consistency with amplitude amplification.
  - Added type annotations to `grovers_alg`, `diffuser`, and `tag_state`
  - Fixed mutable default argument: `kwargs={}` → `kwargs=None` (handled safely inside)
  - Changed `iterations=0` sentinel to `iterations: int | None = None`; iteration detection now uses `elif iterations is None` instead of `== 0`
  - Simplified exact Grover iteration count: replaced the iterative `while_loop`/`while` approach with the closed-form formula $\lceil \pi/(4\theta) - 0.5 \rceil$ where $\theta = \arcsin(\sqrt{M/N})$
  - Replaced `raise Exception(...)` for missing `winner_state_amount` in exact mode with `raise ValueError(...)`
  - Non-Jasp path now uses `elif iterations > 0` (skips `IterationEnvironment` entirely for 0 iterations)
  - Updated `diffuser` and `tag_state` docstrings: `float or sympy.Symbol` → `qrisp.Param`
  - Removed inlined diffuser implementation that had been commented out behind a stray `"""` block
- **test_grovers_algorithm.py**:
  - Refactored monolithic `test_grovers_algorithm` into 6 focused tests: `_basic_oracle`, `_equation_oracle`, `_exact_mode`, `_quantum_array`, `_all_states_tagged`, `_exact_missing_winner_amount`
  - Added `test_grovers_oracle_with_side_effects` (covers [#586](https://github.com/eclipse-qrisp/Qrisp/issues/586)): Grover with an oracle that acts on variables not belonging to the input `args`
  - Replaced fragile exact-dict equality assertions with `np.isclose`-based checks and `set`-based winner-state verification
  - Replaced `sbp_mult` uncompute pattern with injection operator `<<`
  - Moved imports to module level; removed stray `print` statements and `import time`

## How was it tested?

| Test-ID | Status |
|---------|--------|
| `test_grovers_basic_oracle` | ✅ |
| `test_grovers_equation_oracle` | ✅ |
| `test_grovers_exact_mode` | ✅ |
| `test_grovers_quantum_array` | ✅ |
| `test_grovers_list_quantum_array` | ✅ |
| `test_grovers_all_states_tagged` (exact=True / exact=False) | ✅ |
| `test_grovers_exact_missing_winner_amount` | ✅ |
| `test_grovers_oracle_with_side_effects` | ✅ |

## Checklist

## Reviewer Notes

- Commented-out oracle-uncompute check (`if qv_amount != len(qs.qv_list)`) which caused an Exception if the oracle acts on variables not belonging to the input `args`


src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,18 +39,25 @@
...
+):
     r"""
     Applies the Grover diffuser onto (multiple) QuantumVariables.
 
     Parameters
     ----------
     input_object : QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray]
         The (list of) QuantumVariables to apply the Grover diffuser on.
-    phase : float or sympy.Symbol, optional
+    phase : qrisp.Param, optional
```
positr0nium: This gives the impression that we have some exotic system of describing parameters that users need to get used to when in reality it is much rather very established ways of representing parameters (float, sympy, etc.)

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -430,105 +390,83 @@ def counting_oracle(qv, phase = np.pi, k = 1):
...
+    elif iterations > 0:
 
-        merge(qv_list)
-        qs = recursive_qs_search(qv_list)[0]
-        qv_amount = len(qs.qv_list)
+        merge(args)
+        qs = recursive_qs_search(args)[0]
+        # qv_amount = len(qs.qv_list)
 
         with IterationEnvironment(qs, iterations):
```
positr0nium, line 468: The original purpose of the iteration environment was to bridge Pythonic compilation speed bottlenecks. Now that we have Jasp, we don't need it anymore.

renezander90: Yes, I left it because we have not deprecated the `IterationEnvironment` for far, and it still gives some Python compilation speed advantages in static mode. But, we could also already remove it here if you prefer.
We should also consider deprecating the IterationEnvironment in the future because it could cause confusion regarding migration to Jasp.

positr0nium: The issue is now backlogged [here](https://github.com/eclipse-qrisp/Qrisp/issues/606), so I guess we can go ahead and merge this PR.

positr0nium: (maybe after addressing the typing thing in the docstring 😇)

src/qrisp/algorithms/grover/grover_tools.py
renezander90: So you would revert back to `float or sympy.Symbol`? Should we then also add, e.g., `np.number`, `jax.core.Tracer` here? This situation appears in different places. What would be the general strategy? Also looping in @PietropaoloFrisoni

positr0nium: Maybe a good compromise would be to have do something like
``float or floatlike``
and linking to the typing docs to explain what floatlike means?

PietropaoloFrisoni: For context, this came up because some parameters were only annotated as `int`, `float`, etc., and Pylance was complaining because it could not find some attributes that we were using (as a parameter can also be `sympy.Symbol`, `Tracer`, etc., depending on the context). Hence, the need to use a common and consistent type alias for all the parameters.

I was not sure about the name, though. If `Param` looks too structured to you, I agree that something like `floatlike`, `FloatLike`, `Scalarlight`, etc. could be better. You can just change `Param` into something else all at once in one single command 👍

src/qrisp/algorithms/grover/grover_tools.py
PietropaoloFrisoni, line 1: The `gphase` and `invert` functions are imported in this file but not used (this was pre-existing issue, not related to this PR). We might also kill two birds with one stone and run `isort` to sort all the imports. I did it on my branch with `isort src/qrisp/algorithms/grover/grover_tools.py` and it showed some changes.

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,18 +39,25 @@
...
+from typing import Callable, Any
 
 
 # Applies the grover diffuser onto the (list of) quantum variable input_object
-def diffuser(input_object, phase=np.pi, state_function=None, reflection_indices=None):
+def diffuser(
+    input_object: QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray],
+    phase: Param = np.pi,
+    state_function: Callable = None,
+    reflection_indices: list[int] = None,
```
PietropaoloFrisoni:
```suggestion
    state_function: Callable | None = None,
    reflection_indices: list[int] | None = None,
```

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,18 +39,25 @@
...
+from qrisp.typing import Param
+from typing import Callable, Any
 
 
 # Applies the grover diffuser onto the (list of) quantum variable input_object
-def diffuser(input_object, phase=np.pi, state_function=None, reflection_indices=None):
+def diffuser(
+    input_object: QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray],
+    phase: Param = np.pi,
+    state_function: Callable = None,
```
PietropaoloFrisoni: `Pylance` (but I believe you can also check this with other code quality checkers), warns that: 

```
Parameter declaration "state_function" is obscured by a declaration of the same name. 
```

It should not make any difference here, but for consistency, we could just replace the `state_function` defined within this function with a different name (it is not changing anything)

src/qrisp/algorithms/grover/grover_tools.py
PietropaoloFrisoni: `refection_indices` should be `reflection_indices`

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -430,105 +390,83 @@ def counting_oracle(qv, phase = np.pi, k = 1):
...
+                diffuser(args)
 
-        if qv_amount != len(qs.qv_list):
-            raise Exception(
-                "Applied oracle introducing new QuantumVariables without uncomputing/deleting"
-            )
+        # if qv_amount != len(qs.qv_list):
+        #    raise Exception(
+        #        "Applied oracle introducing new QuantumVariables without uncomputing/deleting"
+        #    )
```
PietropaoloFrisoni, line 481: Curiosity: why do you leave commented code like this in the source code? I have seen this in other contexts as well but I never figured it out. I have nothing against it if it is necessary/useful, but maybe it would be helpful to add a TODO, NOTE, or a brief explanation : )

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -282,47 +233,56 @@ def conjugator(qv_list, temp_qf):
 
 
 def grovers_alg(
-    qv_list,
```
PietropaoloFrisoni, line 285: I like the new `args` name more, but isn't this a breaking change in case someone was calling `grovers_alg` by using `qv_list` as keyword argument? That is, `grovers_alg(qv_list=...`?

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -282,47 +233,56 @@ def conjugator(qv_list, temp_qf):
...
-    oracle_function,
-    kwargs={},
-    iterations=0,
-    winner_state_amount=None,
-    exact=False,
+    args: QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray],
```
PietropaoloFrisoni: In case `args` (ex `qv_list`) can also be a tuple or something else, `Sequence` would be more general/appropriate. But even `list` is clear enough from my perspective and this is an optional improvement

PietropaoloFrisoni (commented):
Thanks @renezander90 , this PR is extremely encouraging to see and, from my perspective,  it goes exactly in the right direction!

I just have some minor points/questions : )

Image

src/qrisp/algorithms/grover/grover_tools.py
renezander90: We already have `ScalarLike`: 
Bildschirmfoto 2026-06-02 um 20 00 13

This is almost the same as `Param`:
Bildschirmfoto 2026-06-02 um 20 01 11

Maybe we should consolidate in `ScalarLike`. And in the documentation, we write `float or ScalarLike`? 

Calling a type alias `FloatLike` if it covers also `complex` is not ideal. Or we introduce a `FloatLike` that does not contain `complex`. (In this particular case, the `phase` also should not be complex.)

PietropaoloFrisoni: `ScalarLike` was originally brought up to distinguish JAX scalar arrays from multidimensional arrays (since they have different attributes, as we previously discussed in other PRs). If it weren't for `sympy.Expr`, we could also combine `ScalarLike` with `Param`. 

I don't remember why I inserted `complex` in `Params`, but it seems natural to me to consider only unitary gates by convention (having real parameters only).

What about removing `complex` from `Param` and renaming `Param` into `FloatLike` (or some other name that you prefer)? If we ever need to consider a parameter that can also be complex, we can always use `FloatLike | complex` in the type hint (or introduce another `ComplexLike` alias in the future)

renezander90: This could work. We also would have to further specify `np.number` for the `FloatLike` type alias.

src/qrisp/algorithms/grover/grover_tools.py
renezander90: Yes, if someone uses `qv_list` as keyword argument, it would be a breaking change.

src/qrisp/algorithms/grover/grover_tools.py
renezander90: I leave it if I think it could be used again in the future. But, this can be confusing; I will add a note or remove it here.

src/qrisp/algorithms/grover/grover_tools.py
renezander90: I'm a fan of sorting :). We can do this next week.

src/qrisp/algorithms/grover/grover_tools.py
renezander90: Providing a tuple causes an error. We should keep the `list` for now.

src/qrisp/algorithms/grover/grover_tools.py
PietropaoloFrisoni: Marking this as resolved since the PR replacing `Param` has been created, reviewed, and merged by all the present

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,24 +39,31 @@
 )
 from qrisp.alg_primitives.reflection import reflection
 from qrisp.jasp import check_for_tracing_mode, jrange
+from qrisp.typing import FloatLike
+from typing import Callable, Any, Sequence
```
PietropaoloFrisoni: Sequence should come from collections.abc since Python 3.9: https://docs.python.org/3/library/typing.html#typing.Sequence

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,24 +39,31 @@
...
     r"""
     Applies the Grover diffuser onto (multiple) QuantumVariables.
 
     Parameters
     ----------
-    input_object : QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray]
+    input_object : QuantumVariable | QuantumArray | Sequence[QuantumVariable | QuantumArray]
         The (list of) QuantumVariables to apply the Grover diffuser on.
-    phase : float or sympy.Symbol, optional
+    phase : float | FloatLike, optional
```
PietropaoloFrisoni:
```suggestion
    phase : FloatLike, optional
```

`float` is already included in `FloatLike` : )

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -39,24 +39,31 @@
 )
 from qrisp.alg_primitives.reflection import reflection
 from qrisp.jasp import check_for_tracing_mode, jrange
+from qrisp.typing import FloatLike
+from typing import Callable, Any, Sequence
 
 
 # Applies the grover diffuser onto the (list of) quantum variable input_object
```
PietropaoloFrisoni:
```suggestion
# Applies the grover diffuser onto the (sequence of) quantum variable input_object
```

src/qrisp/algorithms/grover/grover_tools.py
PietropaoloFrisoni: Seems reasonable to me

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -109,98 +116,44 @@ def diffuser(input_object, phase=np.pi, state_function=None, reflection_indices=
...
     Parameters
     ----------
     tag_specificator : dict
         A dictionary specifying which state should be tagged.
     binary_values : bool, optional
         If set to True, the values in the tag_specificator dict have to be bitstrings
         instead of labels. The default is False.
-    phase : float or sympy.Symbol, optional
-        Specify the phase shift the tag should perform. The default is pi, i.e. a
+    phase : float | FloatLike, optional
```
PietropaoloFrisoni:
```suggestion
    phase : FloatLike, optional
```

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -430,105 +392,94 @@ def counting_oracle(qv, phase = np.pi, k = 1):
...
+    if isinstance(args, Sequence):
+        for qv in args:
+                h(qv)
     else:
-        h(qv_list)
+        h(args)
 
     if check_for_tracing_mode():
 
         for i in jrange(iterations):
```
PietropaoloFrisoni:
```suggestion
        for _ in jrange(iterations):
```

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -109,98 +116,44 @@ def diffuser(input_object, phase=np.pi, state_function=None, reflection_indices=
 
     if state_function == None:
```
PietropaoloFrisoni:
```suggestion
    if state_function is None:
```

Since there is just one `None` object in every program (it's a singleton), the general rule is to use `is None` (more efficient and Pythonic) : )

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -282,47 +235,56 @@ def conjugator(qv_list, temp_qf):
...
+        keyword argument ``phase`` to apply the calculated fractional phase shift.
 
     Raises
     ------
-
-    Exception
-        Applied oracle introducing new QuantumVariables without uncomputing/deleting
+    ValueError
+        If ``exact`` is set to True but ``winner_state_amount`` is not specified.
+    ZeroDivisionError
```
PietropaoloFrisoni, line 277: [Optional suggestion] Should we add an explicit guard at the entry point instead?

```
if winner_state_amount is not None and winner_state_amount <= 0:
    raise ValueError("winner_state_amount must be a positive integer")
```

PietropaoloFrisoni (approved):
Good stuff!

PR #593, opened 2026-05-26
Composite gate interpreter

### LINKED ISSUE #550: "Missing" definition of state_init gate in jaspr/mlir
### Priority

Critical

### Responsible

_No response_

### Qrisp Version

0.8.2

### Python Version

3.11.14

### Operating System

macOS

### Short Description

1) When using the qiskit-based ``prepare`` method, the jaspr (and hence the mlir) contains a `` jasp.quantum_gate[gate=state_init]``. The definition of such a gate depends on the particular ``target_array``and is not part of the (printed)  jaspr/mlir. Hence, it cannot be processed starting from the generated mlir.

2) When using ``prepare`` repeatedly (with different target arrays), the name ``gate=state_init`` is ambiguous.

### Steps to Reproduce

```
from qrisp import *

def main():
    qv = QuantumFloat(2)
    prepare(qv, target_array=np.array([1.0, 2.0, 3.0, 4.0]))
    qv2 = QuantumFloat(2)
    prepare(qv2, target_array=np.array([1.0, 5.0, 3.0, 4.0]))
    return measure(qv[0])

jaspr = make_jaspr(main)()
mlir = jaspr.to_mlir(lower_stablehlo=True)

print(jaspr)
#print(mlir)
```

{ lambda ; a:QuantumState. let                  
    b:QubitArray c:QuantumState = jasp.create_qubits 2:i64[] a
    d:Qubit = jasp.get_qubit b 0:i64[]
    e:Qubit = jasp.get_qubit b 1:i64[]
    f:QuantumState = jasp.quantum_gate[gate=state_init] d e c
    g:QubitArray h:QuantumState = jasp.create_qubits 2:i64[] f
    i:Qubit = jasp.get_qubit g 0:i64[]
    j:Qubit = jasp.get_qubit g 1:i64[]
    k:QuantumState = jasp.quantum_gate[gate=state_init] i j h
    l:bool[] m:QuantumState = jasp.measure d k
  in (l, m) }

### Expected Behavior

Good question.

### Actual Behavior

n/a

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive |             |                 |
| T-002   | Negative |             |                 |
| T-003   | Boundary |             |                 |
| T-004   | Boundary |             |                 |

### Additional Context

_No response_

### Definition of Done

- [ ] Root cause identified
- [ ] Regression test added
- [ ] Changelog entry added

renezander90:
## Description

Introduces a new Jasp interpreter pass `decompose_composite_gates` that recursively inlines all composite (non-primitive) quantum gates in a `Jaspr` into their primitive constituents. A composite gate is any `Operation` whose `.definition` attribute holds a sub-circuit. The pass traverses not only top-level equations but also sub-jaxprs nested inside `jit`, `while`, `cond`, and `scan` control-flow primitives, making it applicable to complex programs.

Call-graph compression is preserved: if multiple `jit` equations in the original jaspr reference the same `ClosedJaxpr` object (as produced by `@qache`), the decomposed jaspr ensures they still reference the same transformed object — not independent copies — by caching results via `@lru_cache`.

## Related Issues

Closes #550
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Added `composite_gate_interpreter.py` — new interpreter module containing:
  - `_apply_op(op, qubit_tracers, abs_qst)`: recursively inlines a gate by walking its `.definition` circuit and rebinding each sub-instruction via `quantum_gate_p`
  - `decompose_eqn_evaluator(eqn, context_dic)`: `reinterpret`-compatible evaluator that decomposes composite `quantum_gate_p` equations and recursively applies the transformation to `jit`/`while`/`cond`/`scan` sub-jaxprs
  - `_decompose_sub_jaxpr(jaxpr)`: cached helper (`@lru_cache`) to apply the pass to both `Jaspr` and `ClosedJaxpr` sub-expressions; caching ensures shared sub-jaxpr objects remain shared after transformation
  - `decompose_composite_gates(jaspr)`: public API (`@lru_cache`) — returns a new `Jaspr` with all composite gates inlined
- Updated `__init__.py` to export the new module
- Added test_composite_gate_interpreter.py with a full test suite

## How was it tested?

| Test-ID | Description | Status |
|---------|-------------|--------|
| T-001 | `test_decompose_state_preparation` — `prepare()` introduces a `state_init` composite gate; verifies it is absent after decomposition and that exact output probabilities match the original via `to_qc().run()` + `np.allclose` | ✅ |
| T-002 | `test_decompose_mcx` — multi-controlled X is a composite gate; verifies decomposition and identical deterministic output distribution | ✅ |
| T-003 | `test_decompose_cx_is_noop` — `cx` is already primitive; verifies the pass is a no-op and the Bell-state distribution is preserved | ✅ |
| T-004 | `test_decompose_nested_jaspr` — `BlockEncoding` with deeply nested `jit`/`while`/`cond` sub-jaxprs; verifies full recursive decomposition and that the resulting jaspr remains executable | ✅ |
| T-005 | `test_call_graph_compression_preserved` — two calls to the same `@qache` function share one `ClosedJaxpr` object in the original; verifies the decomposed jaspr maintains that identity (`is` check), confirming call-graph compression is not destroyed | ✅ |
| T-006 | `test_decompose_scan_body` — a `@quantum_kernel` using `prepare()` is called inside a `jax.lax.scan` body; verifies the composite gate inside the scan body jaxpr is decomposed and the result remains executable | ✅ |

## Checklist

## Reviewer Notes

The key design decision is that `decompose_eqn_evaluator` explicitly handles `jit`, `while`, `cond`, and `scan` by patching their embedded sub-jaxprs before re-executing — following the same pattern used in `environment_flattening.py`. Without this, composite gates inside control-flow bodies would be silently left undecomposed.

Call-graph compression is preserved via `@lru_cache` on both `_decompose_sub_jaxpr` and `decompose_composite_gates`, keyed by object identity — matching the pattern used in `get_traced_fun` in catalyst_interpreter.py. Without caching, two `jit` equations referencing the same `ClosedJaxpr` would each receive a distinct copy after transformation, defeating the structural sharing that downstream passes (e.g. XLA's `flatten-call-graph`) rely on to avoid compilation explosion.

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -46,7 +46,7 @@ def _apply_op(op, qubit_tracers, abs_qst):
     if op.definition is None:
         return quantum_gate_p.bind(*qubit_tracers, abs_qst, gate=op)
 
-    defn = op.definition
+    defn = op.definition.transpile()
     for instr in defn.data:
         qubit_indices = [defn.qubits.index(qb) for qb in instr.qubits]
         sub_qubits = [qubit_tracers[i] for i in qubit_indices]
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -98,6 +98,12 @@ def decompose_eqn_evaluator(eqn, context_dic):
         exec_eqn(new_eqn, context_dic)
         return False
 
+    elif eqn.primitive.name == "scan":
+        new_eqn = _copy_eqn(eqn)
+        new_eqn.params["jaxpr"] = _decompose_sub_jaxpr(eqn.params["jaxpr"])
+        exec_eqn(new_eqn, context_dic)
+        return False
+
     return True  # fall back to default execution
 
 
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -1,0 +1,119 @@
+"""
+********************************************************************************
+* Copyright (c) 2026 the Qrisp authors
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+from jax.extend.core import JaxprEqn
+
+from qrisp.jasp.interpreter_tools import exec_eqn, reinterpret, extract_invalues, insert_outvalues
+from qrisp.jasp.primitives import quantum_gate_p
+
+
+def _copy_eqn(eqn):
+    return JaxprEqn(
+        primitive=eqn.primitive,
+        invars=list(eqn.invars),
+        outvars=list(eqn.outvars),
+        params=dict(eqn.params),
+        source_info=eqn.source_info,
+        effects=eqn.effects,
+        ctx=eqn.ctx,
+    )
+
+
+def _apply_op(op, qubit_tracers, abs_qst):
+    """Recursively inline a gate.
+
+    - Primitive gates (op.definition is None) are bound directly via quantum_gate_p.
+    - Composite gates are expanded by iterating their definition circuit and
+      recursively calling _apply_op for each sub-instruction.
+    """
+    if op.definition is None:
+        return quantum_gate_p.bind(*qubit_tracers, abs_qst, gate=op)
+
+    defn = op.definition
+    for instr in defn.data:
+        qubit_indices = [defn.qubits.index(qb) for qb in instr.qubits]
+        sub_qubits = [qubit_tracers[i] for i in qubit_indices]
+        abs_qst = _apply_op(instr.op, sub_qubits, abs_qst)
+
+    return abs_qst
+
+
+def decompose_eqn_evaluator(eqn, context_dic):
+    """Equation evaluator that recursively decomposes composite quantum gates.
+
+    Whenever a quantum_gate_p equation carries a composite gate (one that has
+    a circuit definition), the gate is expanded in-place.
+
+    For jit/while/cond equations whose sub-jaxprs may themselves contain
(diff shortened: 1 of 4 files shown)


src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -0,0 +1,119 @@
...
+    """Recursively inline a gate.
+
+    - Primitive gates (op.definition is None) are bound directly via quantum_gate_p.
+    - Composite gates are expanded by iterating their definition circuit and
+      recursively calling _apply_op for each sub-instruction.
+    """
+    if op.definition is None:
+        return quantum_gate_p.bind(*qubit_tracers, abs_qst, gate=op)
+
+    defn = op.definition
```
positr0nium: you should call op.definition.transpile here because the definition circuit can also contain composite gates

positr0nium: Ah this is fixed because apply_op is called recursively. Might still be faster/less convoluted

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -0,0 +1,119 @@
...
+        gate = eqn.params["gate"]
+        invalues = extract_invalues(eqn, context_dic)
+        qubit_tracers = invalues[: gate.num_qubits]
+        abs_qst = invalues[-1]
+
+        abs_qst = _apply_op(gate, qubit_tracers, abs_qst)
+        insert_outvalues(eqn, context_dic, abs_qst)
+        return False  # equation handled; skip default exec_eqn
+
+    elif eqn.primitive.name == "jit":
```
positr0nium, line 80: The way this is currently implemented destroys call-graph compression. Call graph compression means that jit primitives contain the exact same ClosedJaxpr object when referring to the same function instead of a copy. In other words, there is only one ClosedJaxpr, which is referenced in every jit call. Your current implementation produces new copies for each ClosedJaxpr even if the "compilation source" is the same ClosedJaxpr. Other interpreters solve this through an LRU Cache based mechanism. You could look this up for instance in the catalyst conversion interpreter.

positr0nium (changes requested):

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -0,0 +1,119 @@
...
+        new_eqn = _copy_eqn(eqn)
+        new_eqn.params["jaxpr"] = _decompose_sub_jaxpr(eqn.params["jaxpr"])
+        exec_eqn(new_eqn, context_dic)
+        return False
+
+    elif eqn.primitive.name == "while":
```
positr0nium, line 86: There is another primitive that is not covered here: "scan". Scan describes a loop with a static bound. While there is no way to have Qrisp code to produce scan, Qrisp code can be called inside scan, which would be silently ignored with the current implementation.

positr0nium (approved):

PR #598, opened 2026-05-29
Docs: Static Vs. Dynamic mode

purva-thakre:
## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -73,6 +73,8 @@ copybutton_prompt_is_regexp = True
 # Only copy lines that have prompts (input lines), automatically excluding output
 copybutton_only_copy_prompt_lines = True
 
+nbsphinx_allow_errors = True
+
 # Use Python for default highlighting:
 highlight_language = "python"
 
--- a/documentation/source/general/tutorial/index.rst
+++ b/documentation/source/general/tutorial/index.rst
@@ -267,7 +267,6 @@ Explore the Jasp pipeline to build highly performant algorithms, and ensure your
         +++
         **Using BigInteger to compile Shor's at 2048 bit**
 
-
 Scientific Computing and Quantum Numerical Linear Algebra
 ---------------------------------------------------------
 Unlock the potential of quantum computers for scientific discovery and advanced mathematics.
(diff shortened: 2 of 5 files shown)


purva-thakre:
@positr0nium @renezander90 @PietropaoloFrisoni For a page demonstrating the differences of programming in static and dynamic mode, what would you expect to be mentioned?

My initial idea for adding the page was to show why common patterns used in static mode break when used in dynamic mode followed by a solution to how it works in dynamic mode. Now that I have started working on it, I think it is turning out more about how to program in JAX even though the focus exclusively is on Jasp. Here is [a link](https://drive.google.com/file/d/16ZIlkrpAibhoLxTJIyU35OmeoaoPCntM/view?usp=sharing) to how the current version looks like.

renezander90:
Looks very nice!

- I would also mention that it is useful to replace a larger static loop with a dynamic loop to speed up the compilation (because they do not get unrolled in the IR). 

- I think the most common patters that cause problems when migrating to Jasp are the `TracerBoolConversionError` and `range` vs `jrange` which you already covered :). 

- We should mention that the decorator `jaspify` or `terminal_sampling` (should be mentioned too), must only be apply to the "main", i.e., outer function. Negative example:

```
@jaspify
def test(qv):
    h(qv)

@jaspify
def main():
   qv = QuantumVariable(2)
   test(qv)
   return measure(qv)
```

renezander90:
Not sure if this is already mentioned: 
- `QuantumVariable` decodes to a bit string in static mode, but to an integer in dynamic mode
- `QuantumChar` and `QuantumString` are not compatible with dynamic mode

PR #601, opened 2026-05-29
Fix composite gate interpreter and MLIR lowering

renezander90:
## Description

Composite gates with parametrized expressions (e.g. `rxx(phi)`, which decomposes into `gphase(-phi/2)` and `p(phi)`) were silently broken in two related places.

1. **composite_gate_interpreter.py** — when inlining a composite gate, the sub-gate parameter tracers were not forwarded to the primitive `quantum_gate_p` bind call. Sub-gates whose symbolic expressions referenced a subset of the parent symbols (e.g. only `beta` out of `alpha, beta`) had their tracers dropped entirely, leaving them with zero dynamic parameter invars.

2. **jasp_lowering_rules.py** — the `quantum_gate_p` MLIR lowering rule forwarded raw parameter operands unchanged, ignoring the gate's symbolic `params` expressions entirely. A gate like `gphase(-alpha/2)` was emitted to MLIR as `gphase(alpha)`, losing the `-1/2` factor.

Both bugs are fixed on this branch.

## Related Issues

Closes #
Related to #

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **composite_gate_interpreter.py**
  - Added `_greek_letter_order` lookup to sort a gate's free symbols in canonical `greek_letters` order before binding parameter tracers.
  - Added `_apply_op` helper that recursively inlines composite gates while correctly threading the parent gate's `param_dict` (symbol → tracer) into all primitive sub-gates.
  - `decompose_eqn_evaluator` now builds `param_dict` from the parent gate's `params` list and passes it through `_apply_op` so sparse sub-gate symbol subsets (e.g. only `beta`) are normalized and bound correctly.
  - Added `_decompose_sub_jaxpr` to propagate decomposition into `jit`, `while`, `cond`, and `scan` sub-jaxprs.

- **jasp_lowering_rules.py**
  - Added `_lower_gate_param_expr` — a recursive Sympy-to-StableHLO scalar lowerer handling `Add`, `Mul`, `Pow`, constants, and symbol lookups.
  - Added `_sorted_param_symbols` and `_lower_f64_constant` helpers.
  - `quantum_gate_lowering` now reconstructs each gate parameter operand from `gate.params` symbolic expressions via `_lower_gate_param_expr`, emitting the required StableHLO arithmetic (e.g. `stablehlo.constant dense<-5.0e-01>` + `stablehlo.multiply`) before constructing `jasp.quantum_gate`.

- **test_composite_gate_interpreter.py**
  - Added `test_decompose_parametrized_rxx`: structural check that every parametrized primitive gate in the decomposed jaspr has the correct number of parameter invars; numerical check that the decomposed jaspr yields the same measurement distribution as the non-decomposed one.

- **test_mlir.py**
  - Added `test_mlir_symbolic_gate_parameter_expression_lowering`: asserts that lowering a dynamic `rxx(phi)` circuit to MLIR produces a `gphase` gate with a materialized `-0.5 * phi` expression (`stablehlo.multiply` with `dense<-5.000000e-01>`).

## How was it tested?

| Test-ID | Description | Status |
|---------|-------------|--------|
| T-001 | `test_decompose_parametrized_rxx` — structural + numerical check on `decompose_composite_gates` for `rxx(phi)` | ✅ |
| T-002 | `test_mlir_symbolic_gate_parameter_expression_lowering` — MLIR contains `stablehlo.multiply` with `-0.5` for `gphase(-phi/2)` | ✅ |
| T-003 | `test_mlir_basic_dialect_operations` — existing MLIR lowering of non-parametrized gates unaffected | ✅ |

## Reviewer Notes

The two bugs are logically independent but always co-occur for composite parametrized gates: the interpreter bug loses the tracer, and the MLIR lowering bug loses the expression. Both need to be fixed together for `rxx`, `rzz`, `xxyy`, and any other composite gate that uses abstract parameters in its decomposition.

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -22,6 +22,14 @@ from jax.extend.core import JaxprEqn
 
 from qrisp.jasp.interpreter_tools import exec_eqn, reinterpret, extract_invalues, insert_outvalues
 from qrisp.jasp.primitives import quantum_gate_p
+from qrisp.jasp.primitives.operation_primitive import greek_letters as _greek_letters
+
+# quantum_gate_p.append_impl binds runtime parameter slot i to greek_letters[i].
+# When a decomposed primitive gate uses only a subset of the parent symbols
+# (for example a sub-gate depending only on beta), we must first recover the
+# symbols in canonical greek_letters order and then remap them to a dense
+# prefix alpha, beta, ... before binding the tracers positionally.
+_greek_letter_order = {sym: i for i, sym in enumerate(_greek_letters)}
 
 
 def _copy_eqn(eqn):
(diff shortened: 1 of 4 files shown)


renezander90:
Fix at Jaspr level in #602

PR #602, opened 2026-05-29
Fix composite gate interpreter

renezander90:
## Description

`decompose_composite_gates` had two related bugs in `_apply_op`, both in the branch that handles primitive sub-gates (i.e. gates with `definition is None`):

**Bug 1 — symbolic params dropped (original fix)**  
When a parametrized composite gate (e.g. `rxx(phi)`) was decomposed, the symbolic expressions in its primitive sub-gates (e.g. `gphase(-phi/2)`) were silently dropped. The decomposed JASPR contained those sub-gates with no dynamic parameter invars, so execution with any backend — simulator or MLIR — would produce wrong results.

The root cause was that `_apply_op` passed primitive sub-gates straight through to `quantum_gate_p.bind` without (a) evaluating their symbolic parameter expressions against the parent gate's symbol→tracer bindings, or (b) normalising the gate to identity params so downstream consumers do not re-apply the expression a second time.

**Bug 2 — constant params embedded in gate object instead of becoming invars (new fix)**  
Primitive sub-gates with *concrete* (non-symbolic) constant params — such as `rz(-π/2)` inside the `xxyy` decomposition — also fell through to the bare `quantum_gate_p.bind` call with no param invars. The constant angle was embedded in the gate object rather than passed as an explicit invar. This caused the MLIR lowering to emit a gate call with the angle input variable missing entirely, producing malformed output.

## Related Issues

Closes #  
Related to #

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

**`composite_gate_interpreter.py`**

- Added `_U3_IDENTITY_FACTORIES` and `_make_identity_param_op(op)` — returns a version of a gate whose `params` are exactly `[alpha, beta, ...]` (the dense greek-letter prefix). For standard single-param `U3Gate` types (`gphase`, `rx`, `ry`, `rz`, `p`, `u1`) and the three-param `u3` gate, a fresh gate is created via the existing factory function. This is necessary because `U3Gate.bind_parameters` evaluates the gate's internal `theta/phi/lam/global_phase` fields, not `params` directly — so setting `params` alone is insufficient. For generic `Operation` types the copy's `params` and `abstract_params` are patched directly.

- Fixed `_apply_op` for sub-gates with **symbolic** params: when a primitive sub-gate has `abstract_params`, it now (1) evaluates each `op.params` expression via `lambdify` against the parent gate's symbol→tracer bindings, and (2) calls `_make_identity_param_op` to obtain a gate that downstream consumers can bind correctly. The result is a self-consistent JASPR equation: identity params in the gate object, pre-computed value as the dynamic invar.

- Fixed `_apply_op` for sub-gates with **constant** params: added an `elif op.params:` branch (guarded by `not op.abstract_params`) that promotes each concrete constant to a `jnp.float64` JAX literal tracer and calls `_make_identity_param_op` in the same way as the symbolic path. This ensures every parametrised gate in the decomposed JASPR has its angle(s) as explicit input variables, as required by the MLIR lowering.

**test_composite_gate_interpreter.py**

- Added `test_decompose_parametrized_rxx` — structural check (every primitive gate with `abstract_params` in the decomposed jaspr has exactly `len(abstract_params)` parameter invars) and numerical check (same unitary as the non-decomposed reference).

- Added `test_decompose_parametrized_xxyy` — same two invariants for the two-parameter case.

- Added `test_decompose_constant_param_subgates` — specifically targets Bug 2: verifies that after decomposition every gate's `params` list contains only `sympy.Basic` expressions (never raw Python floats/ints). A raw float in `gate.params` is the observable signature of the constant-param bug. Includes a unitary check on a fixed-param version of the circuit.

## How was it tested?

| Test-ID | Status |
|---------|--------|
| `test_decompose_state_preparation` | ✅ |
| `test_decompose_mcx` | ✅ |
| `test_decompose_cx_is_noop` | ✅ |
| `test_decompose_nested_jaspr` | ✅ |
| `test_call_graph_compression_preserved` | ✅ |
| `test_decompose_scan_body` | ✅ |
| `test_decompose_constant_param_subgates` (new) | ✅ |
| `test_decompose_parametrized_rxx` (new) | ✅ |
| `test_decompose_parametrized_xxyy` (new) | ✅ |

## Reviewer Notes

The key invariant this PR establishes: after `decompose_composite_gates`, every `quantum_gate_p` equation in the resulting JASPR is self-consistent — the gate object has identity params (`params[j] == greek_letters[j]`) and its dynamic invars carry the fully evaluated concrete values, whether those values were computed from symbolic expressions or were plain constants in the original composite definition. No backend needs to interpret the gate's original symbolic expression or read embedded numeric values from the gate object.

The fix deliberately does not touch `jasp_lowering_rules.py`. Making the JASPR semantically correct at the IR level is preferable to compensating for a broken IR inside a single backend lowering rule, because it also fixes correctness for the simulator path (`append_impl`) and the inversion transform (`inv_transform.py`).

The `_U3_IDENTITY_FACTORIES` dictionary must be extended whenever a new named `U3Gate` variant is introduced that has non-zero params. Currently registered: `gphase`, `rx`, `ry`, `rz`, `p`, `u1`, `u3`.

### DIFF
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -118,7 +118,7 @@ def _apply_op(op, qubit_tracers, abs_qst, param_dict=None):
             # Evaluate each symbolic parameter expression (e.g. -alpha/2) against
             # the parent gate's symbol->tracer bindings to produce a concrete JAX
             # tracer for each parameter slot.
-            computed_tracers = [_lambdify(sorted_syms, expr)(*sorted_tracers) for expr in op.params]
+            computed_tracers = [_lambdify(sorted_syms, expr, modules="jax")(*sorted_tracers) for expr in op.params]
             # Emit an identity-param version of the gate (params = [alpha, beta, ...])
             # together with the pre-computed tracers.  This is required because
             # append_impl later calls gate.bind_parameters({alpha: val}), which
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -17,9 +17,8 @@
 """
 
 from functools import lru_cache
-
 from jax.extend.core import JaxprEqn
-
+import jax.numpy as jnp
 from sympy import lambdify as _lambdify
 
 from qrisp.jasp.interpreter_tools import exec_eqn, reinterpret, extract_invalues, insert_outvalues
@@ -39,6 +38,8 @@ _U3_IDENTITY_FACTORIES = {
     "ry":     lambda: _std_ops.RYGate(_greek_letters[0]),
     "rz":     lambda: _std_ops.RZGate(_greek_letters[0]),
     "p":      lambda: _std_ops.PGate(_greek_letters[0]),
+    "u1":     lambda: _std_ops.U1Gate(_greek_letters[0]),
+    "u3":     lambda: _U3Gate(_greek_letters[0], _greek_letters[1], _greek_letters[2]),
 }
 
 
@@ -126,6 +127,18 @@ def _apply_op(op, qubit_tracers, abs_qst, param_dict=None):
             # time: bind_parameters({alpha: val}) would yield -val/2 instead of val.
             identity_op = _make_identity_param_op(op)
             return quantum_gate_p.bind(*qubit_tracers, *computed_tracers, abs_qst, gate=identity_op)
+        elif op.params:
+            # Gate has only constant (non-symbolic) parameters — e.g. rz(-π/2)
+            # emitted as an internal fixed rotation inside a composite gate
+            # definition.  Without this branch the parameter value stays embedded
+            # in the gate object and no param invar is emitted.  The MLIR lowering
+            # expects every parametrised gate to have its angle(s) as explicit
+            # input variables, so we promote each constant to a JAX float64 literal
+            # tracer and emit an identity-param gate, exactly as in the symbolic
+            # branch above.
+            const_tracers = [jnp.float64(float(p)) for p in op.params]
+            identity_op = _make_identity_param_op(op)
+            return quantum_gate_p.bind(*qubit_tracers, *const_tracers, abs_qst, gate=identity_op)
         else:
             return quantum_gate_p.bind(*qubit_tracers, abs_qst, gate=op)
 
--- a/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
+++ b/src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
@@ -106,16 +106,24 @@ def _apply_op(op, qubit_tracers, abs_qst, param_dict=None):
 
     if op.definition is None:
         if op.abstract_params and param_dict:
+            # op.abstract_params is a set, so iteration order is non-deterministic.
+            # lambdify maps positional arguments to symbols in the order given, so
+            # the symbol list passed to lambdify and the tracer list must agree on
+            # the same ordering.  We sort by _greek_letter_order (the canonical
+            # index of each symbol in the greek_letters sequence) to get a stable,
+            # deterministic ordering regardless of Python's set hash randomisation.
+            sorted_syms = sorted(op.abstract_params, key=lambda s: _greek_letter_order[s])
+            sorted_tracers = [param_dict[sym] for sym in sorted_syms]
             # Evaluate each symbolic parameter expression (e.g. -alpha/2) against
             # the parent gate's symbol->tracer bindings to produce a concrete JAX
             # tracer for each parameter slot.
-            sorted_syms = sorted(op.abstract_params, key=lambda s: _greek_letter_order[s])
-            sorted_tracers = [param_dict[sym] for sym in sorted_syms]
             computed_tracers = [_lambdify(sorted_syms, expr)(*sorted_tracers) for expr in op.params]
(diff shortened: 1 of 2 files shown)


src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -17,11 +17,67 @@
...
+
+def _make_identity_param_op(op):
+    """Return a version of op whose params are [alpha, beta, ...] (identity expressions).
+
+    Two cases require different treatment:
+
+    U3Gate instances (gphase, rx, ry, rz, p, ...):
+        U3Gate.bind_parameters evaluates its internal fields (theta, phi, lam,
+        global_phase) directly — it does NOT read self.params.  Patching params
+        on a copy therefore has no effect.  Instead, a fresh gate must be
```
positr0nium, line 54: Generating a new copy of the gate here is the right call but we should change (at some point) the behavior of U3Gate attribute (lambda, phi, theta etc.) not being patched when ``.params`` is being updated. This could be achieved by making it a property. (tagging @PietropaoloFrisoni )

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -36,21 +92,61 @@ def _copy_eqn(eqn):
...
+            # the symbol list passed to lambdify and the tracer list must agree on
+            # the same ordering.  We sort by _greek_letter_order (the canonical
+            # index of each symbol in the greek_letters sequence) to get a stable,
+            # deterministic ordering regardless of Python's set hash randomisation.
+            sorted_syms = sorted(op.abstract_params, key=lambda s: _greek_letter_order[s])
+            sorted_tracers = [param_dict[sym] for sym in sorted_syms]
+            # Evaluate each symbolic parameter expression (e.g. -alpha/2) against
+            # the parent gate's symbol->tracer bindings to produce a concrete JAX
+            # tracer for each parameter slot.
+            computed_tracers = [_lambdify(sorted_syms, expr)(*sorted_tracers) for expr in op.params]
```
positr0nium:
```suggestion
            computed_tracers = [_lambdify(sorted_syms, expr, modules="jax")(*sorted_tracers) for expr in op.params]
```

positr0nium (approved):

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
PietropaoloFrisoni: Yes, a `setter` should work here

tests/jax_tests/test_composite_gate_interpreter.py
```diff
@@ -264,3 +274,154 @@ def scan_body(carry, x):
...
+        qv = QuantumVariable(2)
+        xxyy(0.5, 0.4, qv[0], qv[1])
+        return qv
+
+    jaspr_fixed = make_jaspr(circuit_fixed)()
+    decomposed_fixed = decompose_composite_gates(jaspr_fixed)
+    assert_same_unitary(jaspr_fixed, decomposed_fixed)
+
+
+def test_decompose_parametrized_rxx():
```
PietropaoloFrisoni: Can we please add another test for `rzz`, since it has the same parametrised sub-gate structure as `rxx`?

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -17,11 +17,67 @@
...
+        constructed via its factory function so that the internal fields carry
+        the identity symbolic structure (e.g. GPhaseGate(alpha) has
+        global_phase=alpha, not -alpha/2).
+
+    Generic Operation instances:
+        Operation.bind_parameters lambdifies over self.params, so patching
+        params=[alpha, beta, ...] and clearing the lambdify cache is sufficient.
+    """
+    n = len(op.params)
+    if n == 0:
```
PietropaoloFrisoni:
```suggestion
    if not n:
```

https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-comparison-to-zero.html

src/qrisp/jasp/interpreter_tools/interpreters/composite_gate_interpreter.py
```diff
@@ -36,21 +92,61 @@ def _copy_eqn(eqn):
...
+            computed_tracers = [_lambdify(sorted_syms, expr, modules="jax")(*sorted_tracers) for expr in op.params]
+            # Emit an identity-param version of the gate (params = [alpha, beta, ...])
+            # together with the pre-computed tracers.  This is required because
+            # append_impl later calls gate.bind_parameters({alpha: val}), which
+            # must map val -> val (identity).  Using the original gate (e.g.
+            # gphase(-alpha/2)) would cause the expression to be applied a second
+            # time: bind_parameters({alpha: val}) would yield -val/2 instead of val.
+            identity_op = _make_identity_param_op(op)
+            return quantum_gate_p.bind(*qubit_tracers, *computed_tracers, abs_qst, gate=identity_op)
+        elif op.params:
```
PietropaoloFrisoni:
```suggestion
        if op.params:
```


https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/no-else-return.html

PietropaoloFrisoni (commented):
Thanks @renezander90, looks great!

PietropaoloFrisoni (approved):
🚀

PR #607, opened 2026-06-02
QSVT

MatP1337:
## Description

Introduces `QSVT` (Quantum Singular Value Transformation) as a first-class algorithm and uses it to replace or extend the existing Hermitian-only approaches in the block encoding framework. This PR also applies the same Base Class + Facade Pattern from `refactor_block_encoding` to the transformation methods (`inv`, `sim`, `poly`, `pseudo_inv`): these are extracted from the monolithic block_encoding_base.py into a dedicated `transformations/` subpackage. Additionally, a `dagger()` method and a new `svt()` transformation method are added to `BlockEncoding`.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **New `QSVT` algorithm** (qsvt.py): implements Quantum Singular Value Transformation for block-encoded operators via `qsvt_angles`. Accepts `BlockEncoding`, `QubitOperator`, and `FermionicOperator`. Supports both `"odd"` and `"even"` parity and `"Polynomial"` / `"Chebyshev"` bases.
- **New `BlockEncoding.svt()` method**: applies QSVT directly on a `BlockEncoding` instance for singular value transformation.
- **New `BlockEncoding.dagger()` method**: returns a `BlockEncoding` representing the Hermitian conjugate $A^\dagger$ of the encoded operator.
- **New `BlockEncoding.pseudo_inv()` method**: computes the pseudo-inverse of the encoded operator via QSVT, with a configurable threshold $\theta$.
- **`inversion` updated**: added `"QSVT"` as a supported (and now default) method alongside `"QET"` and `"GQSVT"`.
- **`transformations/` subpackage** (transformations): extracted `inv`, `sim`, `poly`, `pseudo_inv`, and new `svt` from block_encoding_base.py into individual modules, following the same facade pattern as the `constructors/` subpackage. The facade block_encoding.py now attaches them as instance methods.
- **`gqsp_angles.py`**: added inline documentation for the QSVT angle computation procedure.
- **`gqsvt.py`**: minor fixes and updated cross-reference to `QSVT`.
- **`dalzell_inversion.py`**: minor updates.
- **Documentation**: new `.rst` files for `QSVT`, `svt`, `dagger`, and `pseudo_inv`; updated `BlockEncoding.rst` and `applications.rst`.
- **CI**: updated `qrisp_test.yml`.

## How was it tested?

- `pytest -n 4` across the full test suite passes without regressions.
- New dedicated test files:

| Test file | Coverage |
|---|---|
| test_qsvt.py | QSVT polynomial transformation with 4 parametrized cases (odd/even parity, varying degrees) |
| test_block_encoding_svt.py | `BlockEncoding.svt()` scaling factor correctness |
| test_block_encoding_pseudo_inv.py | `BlockEncoding.pseudo_inv()` correctness vs classical pseudo-inverse |
- Existing tests for `inv`, `poly`, `chebyshev`, `sim`, `arithmetic`, and `gqsvt` all pass after the extraction refactoring.

## Checklist


MatP1337:
@renezander90 should we distinguish (in a folder) between the generalized approaches and the vanilla approaches?

qet and qsvt being in the gqsp folder seems like a stretch. 

Would be a small change (:

MatP1337:
Suggested reviewers @renezander90 & @positr0nium 

One could also add the algorithm schematics for QSVT, QSP, and GQSP within this PR. Upon at least one thumbs up, affirmative sentiment to this I'll do it.

renezander90 (approved):

documentation/source/reference/Block Encodings/applications.rst
```diff
@@ -18,6 +18,10 @@ Key applications of block-encodings are:
 
 - **Matrix Inversion:** Solves the Quantum Linear Systems Problem (QLSP) by applying a polynomial transformation approximating :math:`1/x` to the eigenvalues of an encoded matrix [Childs2017]_.
 
+- **Threshold Matrix Pseudoinverse:** : Approximates the pseudoinverse of a matrix by ignoring singular values below a specified threshold.
```
positr0nium, line 21: Cool!

src/qrisp/algorithms/gqsp/inversion.py
```diff
@@ -21,11 +21,12 @@
 from qrisp.algorithms.cks import cks_coeffs, cks_params
 from qrisp.algorithms.gqsp.gqsvt import GQSVT
 from qrisp.algorithms.gqsp.qet import QET
+from qrisp.algorithms.gqsp.qsvt import QSVT
 from qrisp.block_encodings import BlockEncoding
 from typing import Literal
 
 
-def inversion(A: BlockEncoding, eps: float, kappa: float, method: Literal["QET", "GQSVT"] = "QET") -> BlockEncoding:
+def inversion(A: BlockEncoding, eps: float, kappa: float, method: Literal["QET", "QSVT", "GQSVT"] = "QSVT") -> BlockEncoding:
```
positr0nium, line 30: I'm not sure if it is the best idea to name this function simply "inversion". There are so many different kind of inversion in this field (arithmetic inversion, unitary inversion, modular inversion). Yes, it is imported from a location that makes the context obvious but seeing it in code could make it difficult to read and reason especially if several topics that have their own inversion come together.

src/qrisp/algorithms/gqsp/pseudo_inversion.py
```diff
@@ -189,7 +190,7 @@ def threshold_pseudoinverse(A, threshold):
     p = _pseudo_inversion_cheb(theta, delta, eps)
 
     # Set _rescale=False to apply p(A/α) instead of p(A).
-    A_pseudo_inv = GQSVT(A, p, kind="Chebyshev", rescale=False)
+    A_pseudo_inv = QSVT(A, p, kind="Chebyshev", rescale=False)
```
positr0nium, line 193: Is it more efficient to do it like this?

src/qrisp/algorithms/gqsp/qsvt.py
```diff
@@ -0,0 +1,225 @@
...
+    if kind not in ALLOWED_KINDS:
+        raise ValueError(
+            f"Invalid kind specified: '{kind}'. "
+            f"Allowed kinds are: {', '.join(ALLOWED_KINDS)}"
+        )
+
+    if isinstance(A, (QubitOperator, FermionicOperator)):
+        A = BlockEncoding.from_operator(A)
+
+    # Rescaling of the polynomial to account for scaling factor alpha of block-encoding
```
positr0nium, line 182: It would be nice to have some more comments here

src/qrisp/block_encodings/transformations/sim.py
```diff
@@ -0,0 +1,140 @@
...
+        c = np.abs(psi_(0.5))
+        print(c)
+        #[0.88288217 0.22468197 0.22269638 0.05723056 0.22269638 0.05669446
+        # 0.05705877 0.01457772 0.22468197 0.0571786  0.05669446 0.01456988
+        # 0.05723056 0.01456988 0.01457772 0.00372439]
+
+    """
+    from qrisp.algorithms.gqsp import hamiltonian_simulation
+
+    return hamiltonian_simulation(self, t, N)
```
positr0nium, line 140: Same comment as above with inversion. But these functions are already part of the public interface, or? So it is too late anyway..

tests/block_encodings_tests/test_block_encoding_inv.py
```diff
@@ -48,9 +50,6 @@ def main():
         return operand
 
     res_dict = main()
-    for k, v in res_dict.items():
-        res_dict[k] = v**0.5
-    q = np.array([res_dict.get(key, 0) for key in range(len(b))])
-    
-    c = (np.linalg.inv(A) @ b) / np.linalg.norm(np.linalg.inv(A) @ b)
-    assert np.linalg.norm(np.abs(c) - q) < 1e-2
+    amps = np.sqrt([res_dict.get(key, 0) for key in range(len(b))])
```
positr0nium, line 53: This operation seems to appear a lot. Maybe we should introduce some convenience wrapper for it.

positr0nium (approved):
Amazing work!

src/qrisp/algorithms/gqsp/inversion.py
renezander90: The `inversion` was introduced in 0.8. Changing the name would be out of scope for this PR, and require some careful consideration. A topic for another release :) 
Most users will likely interact with the inversion via `BlockEncoding.inv`.

src/qrisp/algorithms/gqsp/pseudo_inversion.py
renezander90: Yes, with (our implementation of) GQSVT we require two times the number of calls to the BlockEncoding oracle. We essentially pay for a benefit (complex polynomials) we don't need (pseudo inverse uses a real polynomial).

src/qrisp/block_encodings/transformations/sim.py
renezander90: Yes, they were introduced in 0.8. I would not want to change the name on short notice. This should require some careful consideration. And it would definitively require a proper deprecation cycle.

PR #610, opened 2026-06-03
`convert_from_cirq` function

### LINKED ISSUE #498: Add `convert_from_cirq`
`convert_to_cirq` was added in #307. 

We also want to add a converter function the other way. This will also make https://github.com/unitaryfoundation/mitiq/pull/2894#issuecomment-4073545419 easier as one can avoid the convoluted Cirq -> Qiskit -> Qrisp conversion path. 

Use this issue to also take a look at the changes in a9f4f967a1fc501344273705bdf72425eeed8a1f as this should have lead to a unit test failure.

purva-thakre:
## Description

Fixes #498 

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist


tests/interface_tests/test_cirq_converter.py
```diff
@@ -201,3 +171,384 @@ def test_recursive_conversion():
...
+
+
+def _build_multi_valued_circ():
+    q0, q1, q2 = cirq.LineQubit.range(3)
+    g = cirq.ControlledGate(cirq.X, control_values=[(0, 1), 1])
+    return cirq.Circuit([g(q0, q1, q2)])
+
+
+def _build_global_phase_only_circ():
+    return cirq.Circuit([cirq.GlobalPhaseGate(1j).on()])
```
PietropaoloFrisoni: Is this used somewhere? if not, can we just remove it?

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -30,196 +58,409 @@ def convert_to_cirq(qrisp_circuit, cirq_qubits=None):
...
-    # get data from Qrisp circuit
-    qrisp_circ_num_qubits = qrisp_circuit.num_qubits()
-    qrisp_circ_ops_data = qrisp_circuit.data
+    # repeatedly transpile unknown gates until only known ones remain
+    def _unknown_names(circ):
+        return {
+            instr.op.name
+            for instr in circ.data
+            if instr.op.name not in gate_map
+            and instr.op.name not in ("gphase", "qb_alloc", "qb_dealloc")
```
PietropaoloFrisoni:
```suggestion
```
`"gphase"`, `"qb_alloc"`, and `"qb_dealloc"` are already in `gate_map`, so this line seems redundant

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -30,196 +58,409 @@ def convert_to_cirq(qrisp_circuit, cirq_qubits=None):
...
+    during conversion.  Cirq's automatic key generation is used instead, so
+    a round-trip (Qrisp -> Cirq -> Qrisp) will lose the original
+    classical-bit associations.
+    """
+    try:
+        import cirq
+    except (ModuleNotFoundError, ImportError):
+        raise ImportError(
+            "Cirq must be installed to be able to use the Cirq to Qrisp converter."
+        )
```
PietropaoloFrisoni:
```suggestion
    except (ModuleNotFoundError, ImportError) as exc:
        raise ImportError(
            "Cirq must be installed to be able to use the Cirq to Qrisp converter."
        ) from exc
```

https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/raise-missing-from.html

Same for the other `except` below (4 in total)

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -30,196 +58,409 @@ def convert_to_cirq(qrisp_circuit, cirq_qubits=None):
...
+            continue
+
+        # controlled operations (multi-qubit)
+        if isinstance(instr.op, ControlledOperation):
+            cs = instr.op.ctrl_state
+            # ctrl_state is a string ("11") or int (-1 for all-ones);
+            # derive the number of control qubits from whichever form we got
+            n_ctrl = len(cs) if isinstance(cs, str) else (
+                instr.op.num_qubits - instr.op.base_operation.num_qubits
+            )
```
PietropaoloFrisoni:
```suggestion
            n_ctrl = len(cs)
```

Looking at the code, it seems to me that `ControlledOperation.ctrl_state` is always a `str` at this point (that is, when we reach this line of code). If this the case, we should also change the comment above.

PietropaoloFrisoni (commented):
Thanks @purva-thakre ! 🚀

tests/interface_tests/test_cirq_converter.py
purva-thakre: Good catch! I had a unit test using this but I decided to remove it as the behavior was tested elsewhere. Forgot to remove this.

src/qrisp/interface/converter/cirq_converter.py
```diff
@@ -21,205 +21,442 @@
...
+            # Convert ctrl_state for Cirq's .controlled()
+            if isinstance(cs, str):
+                ctrl_vals = [int(c) for c in cs]
+            elif isinstance(cs, int) and cs != -1:
+                # integer ctrl_state like 0b101 → [1, 0, 1]
+                ctrl_vals = [int(b) for b in bin(cs)[2:].zfill(n_ctrl)]
+            else:
+                # cs == -1 means all controls on; None tells Cirq the same
+                ctrl_vals = None
+
```
PietropaoloFrisoni:
```suggestion
            ctrl_vals = [int(c) for c in cs]

```

Same as above since we know that `cs` is a string : )

PietropaoloFrisoni (commented):
I used code coverage directly and there are a few lines of source code which are not tested (`222`, `190-191`, and `427-458` according to my tool). We don't have the requirement of testing each line of source code (as other libraries), so it is an optional suggestion whether to cover such lines as well.

purva-thakre:
> We don't have the requirement of testing each line of source code (as other libraries), so it is an optional suggestion whether to cover such lines as well.

my issue is that I am not sure which unit tests are important when the unit tests are getting excessive.

src/qrisp/interface/converter/cirq_converter.py
purva-thakre: yes, I did a similar refactor locally just a few minutes ago.

PietropaoloFrisoni:
> my issue is that I am not sure which unit tests are important when the unit tests are getting excessive.

I think that if tests are properly written and they are concise, we should add as many as possible of them when we are in doubt (although this is a simple example where JAX is not involved, there are 300+ tests in `test_quantum_circuit.py`, but it only takes about 4 seconds to run them sequentially). 

But I agree, this is also one of the reasons why we should have a proper CI-based splitting and parallelization of tests, so that we are not slowed down by them. Of course it does not make sense to have redundant tests, but if we start removing tests because they take take too long  (even when they might be useful), in my opinion we are digging our own grave

PietropaoloFrisoni (approved):

purva-thakre:
@positr0nium Do you want to take a look at this? Or can I go ahead and merge it?

src/qrisp/interface/converter/cirq_converter.py
positr0nium, line 1: Maybe also add a "roundtrip" to the grover test?

positr0nium (approved):
Thanks :)

Issue #570, opened 2026-05-13
Stabilize Stochastic Tests in CI


purva-thakre:
Currently, CI failures related to stochastic unit tests means we have to rerun the CI pipeline until all unit tests pass.  This is mostly because failing unit tests show up in the readme button. 

<img width="457" height="248" alt="Image" src="https://github.com/user-attachments/assets/f9c1a0c7-9f23-4779-a185-0e94c550d1e3" />

This issue aims to reduce CI flakiness by ensuring test outcomes are driven by code changes, not random sampling variance. The first step is to identify and tag tests that are stochastic and intermittently failing. I think some of these unit tests are also set up to skip via pytest. 

Wherever sampling is not the behavior being tested, those tests should be converted to deterministic assertions against exact simulator probabilities. A reproducible baseline should be established by applying consistent seeding for Python, NumPy, and JAX across the test environment.

- For tests that still need randomness, don’t check for exact same numbers every run. Instead, check that results are “close enough” using one common helper so all tests do this the same way.
- if possible: In CI, keep deterministic tests as the main required checks, and run heavier random/stress tests separately.
- Add a way to track flaky failures over time, and remove temporary retries once tests are stable.

PR #615, opened 2026-06-03
Add optional `carry_in` to the Cuccaro Adder

purva-thakre:
## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist

### DIFF
--- a/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
+++ b/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
@@ -30,7 +30,7 @@ def cuccaro_adder(
     a: int | QuantumVariable,
     b: QuantumVariable,
     c_in: QuantumBool | Qubit | None = None,
-    c_out: QuantumVariable | None = None,
+    c_out: QuantumBool | Qubit | None = None,
     ctrl: QuantumBool | None = None,
 ) -> None:
     """In-place adder as introduced in https://arxiv.org/abs/quant-ph/0410184
@@ -56,13 +56,13 @@ def cuccaro_adder(
         The value that should be modified in the in-place addition.
     c_in : QuantumBool or Qubit, optional
         An optional carry in value. The default is None.
-    c_out : QuantumVariable, optional
+    c_out : QuantumBool or Qubit, optional
         An optional carry out value. The default is None.
 
     Raises
     ------
     TypeError
-        If carry in is not of type QuantumBool or Qubit in static mode.
+        If carry in or carry out is not of type QuantumBool or Qubit in static mode.
     ValueError
         If the inputs are not valid quantum or classical types.
 
@@ -133,7 +133,14 @@ def cuccaro_adder(
         cx(c_in, ancilla[0])
 
     if c_out is not None:
-        ancilla2 = c_out
+        if isinstance(c_out, QuantumBool):
+            ancilla2 = c_out[0]
+        elif not check_for_tracing_mode() and not isinstance(c_out, Qubit):
+            raise TypeError(
+                f"c_out must be of type QuantumBool or Qubit, not {type(c_out)}"
+            )
+        else:
+            ancilla2 = c_out
 
     # first maj gate application
     cx(a[0], b[0])
@@ -149,7 +156,7 @@ def cuccaro_adder(
 
     # cnot
     if c_out is not None:
-        cx(a[-1], ancilla2[0])
+        cx(a[-1], ancilla2)
 
     if ctrl is None:
         # iterator uma gate application
--- a/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
+++ b/src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
@@ -15,15 +15,24 @@
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 ********************************************************************************
 """
+
 from qrisp.core import QuantumVariable, x, cx, mcx
+from qrisp.circuit import Qubit
 from qrisp.qtypes import QuantumFloat, QuantumBool
 from qrisp.environments import conjugate, custom_control
 from qrisp.misc import int_encoder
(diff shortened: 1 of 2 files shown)


purva-thakre:
This is a minor fix as I previously failed to account for an optional carry in to the Cuccaro adder.

positr0nium:
I changed the tests for this a bit. The reason for that is that we already have a 70 minute pipeline and we need to be a bit economic what we test with these simulation heavy PRs. The strategy here is to have a few smoke tests in static mode and very exhaustive loops using ``boolean_simulation``. Let me know what you think about this @PietropaoloFrisoni @purva-thakre @renezander90

purva-thakre:
I agree on the unit tests taking too long. If we could agree on which cases require full simulation pipeline and which ones don't, I think this will make it easier for everyone to add unit tests. 

I created #570  for this. I am hoping we get to this for the next milestone.

purva-thakre:
The failures in this PR are related to #621

purva-thakre:
What if we split the unit tests into 2 different workflows: the ones that rely on the full simulation pipeline and the ones that don't. 

The simulation pipeline ones could be run on a nightly basis but not each PR. The ones that don't require simulation could be run each time a change in made in a PR etc. This should hopefully reduce the CI time for us. 

The only thing worth considering is to test for the same type of behavior in static mode and dynamic mode which Raphael is referring to as smoke test, I think.

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -45,6 +45,8 @@ def cuccaro_adder(a, b, c_out=None, ctrl = None):
         The value that should be added.
     b : QuantumVariable or list[Qubit]
         The value that should be modified in the in-place addition.
+    c_in : QuantumVariable, optional
```
positr0nium: The input type doesn't look right (I would assume its QuantumBool || Qubit)

purva-thakre: I was assuming this is a generic type that is checked for in the adder class.

positr0nium: The generic version can also have more than one Qubit. With ``QuantumBool`` and ``Qubit`` as input types, we can be sure that only one qubit is provided.

purva-thakre: Good point.

src/qrisp/alg_primitives/arithmetic/adders/cuccaro_adder.py
```diff
@@ -119,6 +121,15 @@ def cuccaro_adder(a, b, c_out=None, ctrl = None):
 
     ancilla = QuantumFloat(max_size)
 
+    if c_in is not None:
+        if isinstance(c_in, QuantumBool):
+            c_in = c_in[0]
+        elif not check_for_tracing_mode() and not isinstance(c_in, Qubit):
+            raise TypeError(
+                f"c_in must be of type QuantumBool or Qubit, not {type(c_in)}"
+            )
```
purva-thakre, line 132: @positr0nium I am unsure if there is a way to use `Qubit` in dynamic mode. Feel free to suggest a change here. 

I think I am correct that the raw `Qubit` type can only be used in static mode.

positr0nium: This is how it works in principle, but I'm not sure if this is a good long term solution. Maybe some typing magic could improve this? @PietropaoloFrisoni @renezander90 
```python
from qrisp import *
@make_jaspr
def main():
    qv = QuantumVariable(5)
    print(isinstance(qv[0].aval, AbstractQubit))
main()
```

purva-thakre: But it is not a `Qubit` instance here. It is an `AbstractQubit` obtained from a dynamic array. This is already possible when I index a QuantumVariable in dynamic mode (which is what your code block is also doing).

purva-thakre:
```mermaid
flowchart TD
    A["You interact with a QuantumVar<br/>(QuantumFloat, QuantumBool, etc)"]
    B{"Is JAX tracing?<br/><code>check_for_tracing_mode()</code>"}
    C["STATIC MODE"]
    D["DYNAMIC MODE"]
    E["QuantumSession"]
    F["TracingQuantumSession"]
    G["<code>qv.reg = List[Qubit]</code><br/>Concrete Python objects"]
    H["<code>qv.reg = DynamicQubitArray</code><br/>Wraps a JAX tracer array"]
    I["<code>qv[i]</code> returns a <b>Qubit</b> object<br/>Has .qs(), .hash, .identifier, etc."]
    J["<code>qv[i]</code> returns an <b>AbstractQubit</b> tracer<br/>JAX tracer with aval=AbstractQubit()"]

    A --> B
    B -- No --> C
    B -- Yes --> D
    C --> E
    D --> F
    E --> G
    F --> H
    G --> I
    H --> J

    style C fill:#90caf9,stroke:#0d47a1,color:#0d47a1
    style D fill:#ffcc80,stroke:#e65100,color:#e65100
    style I fill:#90caf9,stroke:#0d47a1,color:#0d47a1
    style J fill:#ffcc80,stroke:#e65100,color:#e65100
    style G fill:#90caf9,stroke:#0d47a1,color:#0d47a1
    style H fill:#ffcc80,stroke:#e65100,color:#e65100
```
This is my understanding of when I figured out a `Qubit` is a static object.

positr0nium (approved):

PR #619, opened 2026-06-05
Remove `Job.submit()` from the public interface

PietropaoloFrisoni:
## Description

This PR removes `Job.submit()` from the abstract interface of the `Job` base class. 

This enables more freedom for the vendor(s) implementing their own `Backend` subclasses.
Concrete subclasses may still define their own `submit()` as an internal implementation detail, and existing backends are left unchanged.

## Related Issues

None.

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

> Existing `Job` subclasses that implement `submit()` continue to work without modification. The only scenario that would break is calling `super().submit()` from a subclass (which previously raised `NotImplementedError` anyway).

## What was changed?

- Removed `@abstractmethod submit()` from `Job`. We updated class docstring as well.

- The rest is a consequence of this (see changed files).

## How was it tested?

| Test-ID | Status |
|---------|--------|
| `TestJobAbstractInterface::test_job_cannot_be_instantiated_directly` | ✅ |
| `TestJobAbstractInterface::test_result_is_abstract` | ✅ |
| `TestJobAbstractInterface::test_cancel_is_abstract` | ✅ |
| `TestJobAbstractInterface::test_status_is_abstract` | ✅ |
| `JobTestLifecycle::test_job_starts_in_initializing_state` | ✅ |
| `JobTestLifecycle::test_run_async_does_not_leave_job_in_initializing` | ✅ |
| `JobTestLifecycle::test_full_lifecycle_order_via_set_status` | ✅ |
| Full interface test suite (170 tests) | ✅ |

## Checklist

## Reviewer Notes

### DIFF
--- a/documentation/source/reference/Backend Interface/index.rst
+++ b/documentation/source/reference/Backend Interface/index.rst
@@ -380,7 +380,7 @@ the execution is synchronous or asynchronous.
 This follows the *Future* (or *Promise*) pattern: execution happens independently of the
 caller, and the caller decides when to block for the result.
 
-The base ``Job`` class defines *only the observable contract*: four abstract methods that
+The base ``Job`` class defines *only the observable contract*: three abstract methods that
 every concrete implementation must provide.
 It deliberately prescribes no internal synchronisation mechanism.
 A synchronous simulator may resolve the job inline; an asynchronous hardware backend
@@ -400,9 +400,8 @@ From the caller's perspective, both are used identically:
    # Attempt cancellation (best-effort, may return False if already done).
    accepted = job.cancel()
 
-Concrete subclasses must implement the four abstract methods:
+Concrete subclasses must implement three abstract methods:
 
-- :meth:`~qrisp.interface.Job.submit`: trigger the actual execution.
 - :meth:`~qrisp.interface.Job.result`: block until the result is available and return it.
 - :meth:`~qrisp.interface.Job.cancel`: attempt to cancel the job.
 - :meth:`~qrisp.interface.Job.status`: return the current :ref:`JobStatus` without blocking.
@@ -431,7 +430,7 @@ during its lifecycle.
 
 The six states are:
 
-- ``INITIALIZING``: the job has been created but not yet submitted to the backend.
+- ``INITIALIZING``: the job has been created and execution has not yet been handed off to the backend.
 - ``QUEUED``: the job has been submitted and is waiting for execution resources.
 - ``RUNNING``: the job is currently being executed.
 - ``DONE``: the job completed successfully. Results are available via :meth:`~qrisp.interface.Job.result`.
--- a/documentation/source/reference/Backend Interface/index.rst
+++ b/documentation/source/reference/Backend Interface/index.rst
@@ -406,6 +406,20 @@ Concrete subclasses must implement three abstract methods:
 - :meth:`~qrisp.interface.Job.cancel`: attempt to cancel the job.
 - :meth:`~qrisp.interface.Job.status`: return the current :ref:`JobStatus` without blocking.
 
+.. note::
+
+   *Submission responsibility:* ``Job`` does not mandate how or when a job transitions
+   out of :attr:`~qrisp.interface.JobStatus.INITIALIZING`. That responsibility belongs
+   entirely to the concrete backend author. 
+   
+   The only hard contract is that the job returned by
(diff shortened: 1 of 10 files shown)


PietropaoloFrisoni:
@smite I cannot request your review for some reason (probably because you are not one of the Eclipse commiters), but I tag you explicitly here :+1:

src/qrisp/interface/backend.py
```diff
@@ -56,11 +56,21 @@ def name(self) -> str:
...
+    def run(
+        self, circuits: Sequence[QuantumCircuit], shots: int | None = None
+    ) -> list[MeasurementResult]: ...
+
     def run(
         self,
         circuits: QuantumCircuit | Sequence[QuantumCircuit],
         shots: int | None = None,
-    ):
+    ) -> MeasurementResult | list[MeasurementResult]:
```
PietropaoloFrisoni, line 73: This is a change unrelated with the purpose of this PR, but it is necessary for other reasons (see this discussion https://github.com/eclipse-qrisp/Qrisp/pull/564#discussion_r3348456794)

documentation/source/reference/Backend Interface/index.rst
```diff
@@ -400,13 +400,26 @@ From the caller's perspective, both are used identically:
...
 
+.. note::
+
+   *Submission responsibility:* ``Job`` does not mandate how or when a job transitions
+   out of :attr:`~qrisp.interface.JobStatus.INITIALIZING`. That responsibility belongs
+   entirely to the concrete backend author. 
+   
+   The only hard contract is that the job returned by
+   :meth:`~qrisp.interface.Backend.run_async` must not still be in
+   :attr:`~qrisp.interface.JobStatus.INITIALIZING` state.
```
positr0nium, line 417: Do we still need this JobStatus? What could it be useful for?

PietropaoloFrisoni: `JobStatus` is the data structure that we are using to manage every lifecycle state a Job can be in (it is not the only possible way, but I think this is perfectly reasonable).
 
I originally modelled it after larger frameworks like Qiskit, which also define a `JobStatus` `enum`, both for familiarity and because I think that an `enum` is the most self-documenting way to communicate to backend implementors what states are valid.

In practice, as of today, it is the return type of `Job.status()`. Any caller that polls a job, logs its state, or drives a monitoring loop observes those values directly. Using a typed `enum` rather than raw strings means a typo like `"runing"` or `"Done"` is caught by the type system at development time, not silently at runtime on real hardware. 
The enum is also what makes `JOB_FINAL_STATES` and `in_final_state()` reliable: terminal states are defined once and reused consistently everywhere, rather than scattered string literals that each caller must maintain independently.

We are deliberately more minimal than Qiskit, which includes additional states such as `VALIDATING` (a server-side validation step before execution begins) that we chose not to expose, as it seemed redundant to me : )

positr0nium: I'm not questioning the ``JobStatus`` itself - this is a very good idea and well implemented! I'm questioning if ``JobStatus.INITIALIZING`` is still required because there is no way for users to get this result from a correctly implemented backend.

PietropaoloFrisoni: Yes, I asked myself the same question, but I think we should still have it here. 

It is true that a correctly implemented backend will never expose `INITIALIZING` to the caller (that is precisely the contract). But, on one hand. it is the only semantically correct initial value for `_last_known_status` when the `Job` object is constructed: at that moment the job has not been submitted yet, so `QUEUED`, `RUNNING`, and every other state would be factually wrong. 

On the other hand, the lifecycle invariant:

```
`run_async` must not return while the job is still in `INITIALIZING` state"`
```

can only be expressed and tested because the state exists. It gives backend authors and our own test suite something concrete to assert against, and it makes buggy backends (ones that return a job before it has been handed off) immediately detectable rather than silently producing undefined behaviour. 

I think the fact that callers never see it in production is a feature, not a reason to remove it.

positr0nium: Alright :)

positr0nium (approved):

Issue #616, opened 2026-06-04
Integrate Kipu Quantum's Hub API as a Qrisp backend


iolostanel:
### Priority

Medium

### Responsible

Fraunhofer FOKUS

### Short Description

Integrating Kipu Quantum’s Hub into Eclipse Qrisp will allow Qrisp users to execute their circuits against the available backends.
An integration of the quantum cloud API with the qrisp backend in needed to enable this feature.

### Long Description

The Kipu Quantum cloud is in continuous expansion in terms of available backends (https://hub.kipu-quantum.com/quantum-backends). In order to make the integration clean, each backend provider will be integrated separately.

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive | Circuits are compiled and sent for execution| Satisfactory test results are returned|

PR #620, opened 2026-06-05
feature: qrisp integration with Kipu Quantum backends including testing

iolostanel:
Kipu Quantum Cloud backends integration into Qrisp, ongoing development as more backends become available

## Description

This PR provides the needed interface code to be able to call on Kipu Quantum provided quantum backends directly from qrisp.

## Related Issues

Related to #616

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- added new backend references into /src/qrisp/interface/provider_backends/__init__.py
- added kipu_ibm_backend -ibm qiskit specific backend
- added kipu_qiskit_backend - kipu backends supporting qiskit apart from the one above
- added kipu_braket_backend - kipu backends supporting braket

## How was it tested?
- added test_kipu_ibm_backend -ibm qiskit specific backend tests
- added test_kipu_qiskit_backend - qiskit supporting kipu backends tests
- added test_kipu_braket_backend - kipu braket supporting backends tests

## Checklist


src/qrisp/interface/provider_backends/__init__.py
```diff
@@ -19,3 +19,80 @@
...
+    "list_backends",
+    "create_simulator_backend",
+    # Braket
+    "KipuBraketBackend",
+    "DeviceMode",
+    "create_ionq_forte_backend",
+    "create_aquila_ahs_backend",
+]
+ 
+__version__ = "0.1.0"
```
renezander90: The import errors are caused by this block. Removing this block resolves the errors and the pipeline passes.

src/qrisp/interface/provider_backends/kipu_ibm_backend.py
```diff
@@ -0,0 +1,449 @@
...
+        if hasattr(evs, "tolist"):
+            evs = evs.tolist()
+        return {"expectation_values": evs}
+
+
+# ---------------------------------------------------------------------------
+# Main class
+# ---------------------------------------------------------------------------
+
+class KipuIBMBackend(BatchedBackend):
```
PietropaoloFrisoni: Please notice that `Backend` is the class from which you should inherit to create a backend. 

In the new design, `BatchedBackend` is just a proxy for every backend that you can access with [this method](https://github.com/eclipse-qrisp/Qrisp/blob/615b591d672f5409253f29b804360bfed7b71c5d/src/qrisp/interface/backend.py#L347), which enables batched execution for multiple circuits. 

Things changed (for the better) drastically compared to the previous interface, in which `BatchedBackend` was indeed a class to inherit from.

We can discuss next week (I can help). Thanks for your interest and contribution @iolostanel !

PietropaoloFrisoni:
@iolostanel My understanding is that this feature is not part of the next release (this PR and the related issue are not part of the 0.9 Milestone), so it's not urgent to merge this by the end of the week. 

Please let me know whether you need help, or simply want to chat about the implementation of such backends following the new interface : )

PietropaoloFrisoni (commented):
Thanks again for this contribution @iolostanel !

I think this is a very nice integration. The main issue is that the implementation still follows the old Backend infrastructure and, therefore, does not follow the `Backend` protocol introduced in the current codebase and needs to be reworked before it can be merged (the issues are architectural, not cosmetic or related to the functionality itself).

I used Copilot to organize my main comments below. These are the main architectural changes that we require to be compatible with the new `Backend` protocol:

---

## Critical issues

### 1. Wrong base class: `BatchedBackend` must not be subclassed

All three backends do:

```python
class KipuQiskitBackend(BatchedBackend): ...
class KipuIBMBackend(BatchedBackend): ...
class KipuBraketBackend(BatchedBackend): ...
```

[`BatchedBackend`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/batched_backend.py#L35-L40) is explicitly *not* a base class for backends. Its source comment reads:

> `BatchedBackend` intentionally does not inherit from `Backend`. [...] Inheriting from
> `Backend` would therefore violate the Liskov Substitution Principle.

`BatchedBackend` is a *wrapper* around a `Backend` that clients obtain via `backend.batched()`. The correct pattern is to subclass [`Backend`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/backend.py) and implement `run_async`. Users who want batched execution then call `backend.batched()` themselves. You can find several examples in the `BatchedBackend` documentation : )

Please note that `IQMBackend` in the codebase also used `BatchedBackend` directly, but it is already marked as `deprecated` and is being removed precisely because of this pattern (there will be a new `IQMBackend` in another repo).

---

### 2. No `Job` subclasses: `MockJob` / `MockJobResult` break the protocol

[`Backend.run_async`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/backend.py#L225-L275) is contractually required to return a [`Job`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/job.py#L207) instance (so we can support asynch. execution, as we discussed extensively [in the ADR](https://github.com/eclipse-qrisp/Qrisp/discussions/472#discussioncomment-16070622)). 

All three backends instead define anonymous inner classes:

```python
class MockJobResult:
    all_counts = evaluated_results   # class-level attribute!

class MockJob:
    def result(self, timeout=None):
        return MockJobResult()

return MockJob()
```

Problems:

- `MockJob` does not extend `Job` and does not implement the two required abstract   methods `cancel()` and `status()`. Any caller that polls `job.status()`,  calls `job.done()`, or checks `isinstance(job, Job)` will fail.

- `MockJobResult.all_counts` is a **class-level** attribute set at class definition  time. It is not a property and not a `JobResult` instance. 

- The status lifecycle (`INITIALIZING → QUEUED → RUNNING → DONE`) is never  advanced, so `job.last_known_status` is always `INITIALIZING`.

The recommended pattern for a *synchronous* backend is shown in [`QrispSimulatorJob`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/simulators/qrisp_simulator_backend.py#L29) and for an *asynchronous* one in [`QiskitJob`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/provider_backends/qiskit_backend.py#L73). The only thing we should really try to follow is the `run_async -> Job` protocol, everything else is up to us and obviously we can be very flexible.

---

## 3. Function-as-backend duck-typing hack

This is just a consequence of the first issue (wrong subclassing). To satisfy `BatchedBackend.__init__(backend)`, the current implementation patches a plain function with instance attributes.

This function is stored as `self._backend`. Calling `self._backend.update_options(**kwargs)` (done by the `update_options` method) will raise `AttributeError` because plain functions do not have that method. 

The `options` returned is a raw `dict`, not the `MappingProxyType` that `Backend.options` guarantees, so callers that assume immutability can silently mutate it. Input validation (`_validate_shots`, `_check_circuit_limit`) is bypassed entirely.

---

### 4. `retrieve_job` returns SDK-native objects, not `Job`

As before, this is also a consequence of the first and second issues.

Both `KipuQiskitBackend.retrieve_job` and `KipuIBMBackend.retrieve_job` return the underlying SDK job directly.

The [`Backend.retrieve_job`](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/interface/backend.py#L362-L398) contract requires a Qrisp `Job` subclass to be returned. The SDK job should be wrapped in the backend's `Job` subclass once one is defined (see issue 2 above).

---

### 5. `_counts_to_qrisp` returns probabilities, not raw counts

All three `_counts_to_qrisp` helpers normalise to floats:

```python
out[key] = v / total if total else 0.0
```

Every other backend in the codebase (`QiskitJob`, `QrispSimulatorJob`, `AQTBackend`) returns raw integer counts and lets Qrisp's own measurement pipeline handle normalisation. 

Passing floats here is inconsistent with the rest of the framework and will produce incorrect results in any downstream code that sums or re-normalises (we spent a significant amount of time designing the measurement pipeline as we needed to be compatible with `BatchedBackend` and avoid overhead in post-processing when dealing with a batch of circuits).

---


These are the main/minimal things we should fix at this stage : )

You can find a lot of examples and explanations about the new `Backend` interface in the codebase, documentation, and in the ADR that I linked above. I am happy to support, so please let me know if you need help or you simply want to discuss about this.

Thanks again!

iolostanel:
Thank you! I'm working on the changes and the new code should be ready by eow.

documentation/source/reference/Backend Interface/Hardware Backends/KipuBackends.rst
```diff
@@ -0,0 +1,56 @@
+Kipu Quantum Hub Backends
```
PietropaoloFrisoni:
```suggestion
.. _KipuBackends:

Kipu Quantum Hub Backends
```

documentation/source/reference/Backend Interface/Hardware Backends/KipuBackends.rst
```diff
@@ -0,0 +1,56 @@
...
+=========================
+
+The Kipu Quantum Hub interface module provides high-performance, unified, and cloud-compatible wrappers enabling the execution of Qrisp computational algorithms on multi-tenant hardware clusters. 
+
+Instead of maintaining separate class structures per provider, the refactored architecture relies on a unified ``KipuBackend``. This backend dynamically maps the specified ``backend_id`` to its respective structural schemas (such as AWS, IBM, IQM, Azure, Quandela, or Qudora) and loads the matching job request models on demand.
+
+All classes subclass Qrisp's ``Backend`` infrastructure to optimize throughput, orchestrate execution blocks, and natively parse outcomes back into standard Qrisp frequency metrics.
```
PietropaoloFrisoni:
```suggestion
``KipuBackend`` subclasses Qrisp's :class:`~qrisp.interface.Backend` and ``KipuJob`` subclasses :class:`~qrisp.interface.Job`, following the standard Qrisp backend protocol.
```

It is not clear to me which classes we are referring to. I suggest an alternative version which uses `sphinx`-compatible links

documentation/source/reference/Backend Interface/Hardware Backends/KipuBackends.rst
```diff
@@ -0,0 +1,56 @@
...
+   :depth: 3
+   :local:
+   :backlinks: none
+
+---
+
+Unified Dynamic Backend
+----------------------
+
+The ``KipuBackend`` serves as the singular integration framework routing all general-purpose gate-based backends and simulators through the Kipu Hub. When initialized without arguments, it gracefully defaults to the Azure IonQ Simulator infrastructure.
```
PietropaoloFrisoni: The `api_key` argument is mandatory, so I wouldn't say that it can be be initialized without arguments

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+
+    _MAP = {
+        "PENDING": JobStatus.QUEUED,
+        "RUNNING": JobStatus.RUNNING,
+        "SUCCEEDED": JobStatus.DONE,
+        "COMPLETED": JobStatus.DONE,
+        "CANCELLED": JobStatus.CANCELLED,
+        "FAILED": JobStatus.ERROR,
+        "ERROR": JobStatus.ERROR,
+    }
```
PietropaoloFrisoni, line 69: I think this should be a module-level constant. This would also avoid recreating this dict on every call to `_map_kipu_status`

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+            while True:
+                if timeout is not None and (time.time() - start_time) > timeout:
+                    raise TimeoutError(f"Polling timed out for Kipu job {job_id!r}")
+
+                status = _map_kipu_status(self._client, job_id)
+                self._last_known_status = status
+
+                if status == JobStatus.DONE:
+                    break
+                elif status == JobStatus.CANCELLED:
```
PietropaoloFrisoni, line 105:
```suggestion
                if status == JobStatus.CANCELLED:
```

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+                    raise JobCancelledError(f"Kipu job {job_id!r} was cancelled.")
+                elif status == JobStatus.ERROR:
```
PietropaoloFrisoni, line 107:
```suggestion
                if status == JobStatus.ERROR:
```

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+    def submit(self) -> None:
+        """Mark the job as QUEUED. The jobs are already submitted during run_async()."""
+        self._last_known_status = JobStatus.QUEUED
+
+    def result(self, timeout: float | None = None) -> JobResult:
+        """Block until all associated Kipu jobs complete and return a :class:`~qrisp.interface.JobResult`."""
+        start_time = time.time()
+        poll_interval = 2
+
+        for job_id in self._job_ids:
```
PietropaoloFrisoni, line 95: It seems to me that this polls jobs sequentially (if job N+1 completes before job N, the caller waits for job N anyway before moving on). Is this necessary or is there a way to avoid this?

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+                raw_results = self._client.jobs.get_job_result(job_id)
+            except Exception as exc:
+                raise JobFailureError(f"Failed to fetch results for job {job_id!r}: {exc}") from exc
+            
+            result_dicts.append(self._post_process_to_qrisp(raw_results))
+
+        return JobResult(result_dicts)
+
+    def cancel(self) -> bool:
+        """Attempt to cancel all active jobs. Returns True if all actions completed."""
```
PietropaoloFrisoni, line 126:
```suggestion
        """Attempt to cancel all active jobs. Returns True if all actions completed."""
        if self._last_known_status in (JobStatus.DONE, JobStatus.ERROR, JobStatus.CANCELLED):
            return False
```

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+            counts_dict = raw_results["histogram"]
+        elif isinstance(raw_results, list) and len(raw_results) > 0:
+            counts_dict = raw_results[0]
+        elif hasattr(raw_results, "results") and len(raw_results.results) > 0:
+            counts_dict = raw_results.results[0]
+        else:
+            counts_dict = raw_results
+
+        if not isinstance(counts_dict, dict):
+            counts_dict = getattr(counts_dict, "counts", {}) or counts_dict.get("counts", {}) or {}
```
PietropaoloFrisoni, line 168:
```suggestion
            counts_dict = getattr(counts_dict, "counts", None) or {}
```

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+
+        processed_results = {}
+        for bitstring, val in counts_dict.items():
+            if isinstance(val, float):
+                processed_results[str(bitstring)] = round(val, 4)
+            else:
+                processed_results[str(bitstring)] = round(val / self._shots, 4)
```
PietropaoloFrisoni, line 175: Every other backend in the codebase (`QiskitJob`, `QrispSimulatorJob`, `AQTBackend`) returns raw integer counts since Qrisp's measurement pipeline normalises them internally (@renezander90  can confirm or deny this), so we can just use `int` here

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+        self._check_circuit_limit(circuits)
+        job_ids = []
+
+        for qc in circuits:
+            native_items = self._pre_process_circuit(qc)
+            
+            job_input = self._input_cls(
+                circuit=native_items,
+                gateset="qis",
+                qubits=qc.num_qubits() if hasattr(qc, "num_qubits") else 2,
```
PietropaoloFrisoni, line 264: `QuantumCircuit` always has `num_qubits`, so the `else 2` branch should be never reached. Furthermore, isn't this hiding a wrong circuit width if a non-`QuantumCircuit` object is passed?

src/qrisp/interface/provider_backends/kipu_qhub_api_backend.py
```diff
@@ -0,0 +1,307 @@
...
+                    self._item_cls(gate=gate_name, targets=[qubit_indices[0]])
+                )
+            elif len(qubit_indices) == 2:
+                native_items.append(
+                    self._item_cls(
+                        gate=gate_name, 
+                        controls=[qubit_indices[0]], 
+                        targets=[qubit_indices[1]]
+                    )
+                )
```
PietropaoloFrisoni, line 306: If 3+ qubit gates cannot be handled, I would suggest raising a `ValueError` here.

Why are we not taking them into account?

tests/interface_tests/test_kipu_qhub_api_backend.py
PietropaoloFrisoni, line 1: I would definitely add several tests here.

I run the coverage test locally and most of the source code lines are not tested. I would suggest testing all source code lines. There are, in my opinion, the most important starting points:

- add a test for `KipuJob.cancel()` — both the happy path (returns `True`, status transitions to `CANCELLED`) and the terminal-state guard (returns `False` when job is already `DONE/ERROR/CANCELLED`).

- add a test for `KipuJob.cancel()` when the client has no `cancel_job` attribute.

- add a test for `KipuJob.status()` with multiple job IDs that have mixed statuses (e.g. one `RUNNING`, one `DONE`) to verify the aggregation logic.

- add a test for `result()` with a timeout that expires before the job completes, asserting `TimeoutError` is raised.

- add a test for `run_async()` dispatching multiple circuits, asserting that one `create_job` call is made per circuit and the returned `KipuJob` holds all job IDs.

- add a test for `run_async()` where create_job raises, asserting `RuntimeError`.

- add a test for _pre_process_circuit() with a 3+-qubit gate (if they can be handled)

PietropaoloFrisoni (commented):
Thanks @iolostanel !

PR #622, opened 2026-06-06
PassManager in QrispSimulatorBackend

positr0nium:
## Description

Add a `pm` (PassManager) keyword argument to `QrispSimulatorBackend`, allowing users to inject custom circuit transformation passes that run on every circuit before it reaches the simulator. This is useful for debugging (e.g. inserting `visualize` to inspect the circuits the backend receives) and for applying transpilation/optimisation passes transparently during simulation.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Added `pm` keyword argument to `QrispSimulatorBackend.__init__`, accepting a `PassManager` or `None`, with runtime type validation
- The pass manager is applied in `run_async` to every circuit before simulation
- Added a `.. code-block::` example in the `QrispSimulatorBackend` docstring showing how to use `visualize` in a `PassManager` to inspect expectation value circuits
- Added a matching example in `QubitOperator.expectation_value`'s docstring explaining how `visualize` reveals the change-of-basis circuits generated for each commuting group
- Added 11 pytest tests covering: construction, pass invocation counting, gate-level circuit modification, multiple-circuit batches, pass chaining, shot-based and analytic execution, batched backend integration, and invalid-type rejection

## How was it tested?

11 new pytest tests in `tests/interface_tests/test_default_backend.py::TestQrispSimulatorBackendPassManager`. All pass locally.

| Test-ID | Status |
|---------|--------|
| test_construct_with_pm_none | ✅ |
| test_construct_with_pm | ✅ |
| test_pm_pass_is_called | ✅ |
| test_pm_modifies_circuit_before_simulation | ✅ |
| test_pm_is_applied_to_every_circuit | ✅ |
| test_empty_pm_is_identity | ✅ |
| test_pm_chain_executes_in_order | ✅ |
| test_pm_works_with_shot_based_execution | ✅ |
| test_pm_works_with_batched_backend | ✅ |
| test_pm_works_with_analytic_execution | ✅ |
| test_pm_rejects_non_passmanager | ✅ |

Example usage:

```python
from qrisp import PassManager, visualize
from qrisp.interface import QrispSimulatorBackend

pm = PassManager()
pm += visualize
backend = QrispSimulatorBackend(pm=pm)

# Each circuit sent to the simulator is printed to stdout before execution
```

## Checklist

## Reviewer Notes

- Tests use `qc.data.insert(0, ...)` to prepend gates before measurements, since pass-manager-added gates would otherwise land after existing measure instructions and have no effect.

### DIFF
--- a/src/qrisp/interface/simulators/qrisp_simulator_backend.py
+++ b/src/qrisp/interface/simulators/qrisp_simulator_backend.py
@@ -20,8 +20,8 @@
 
 from __future__ import annotations
 
-from collections.abc import Mapping
-from typing import Sequence, cast
+from collections.abc Sequence
+from typing import cast
 
 from qrisp.circuit.pass_management.pass_manager import PassManager
 from qrisp.circuit.quantum_circuit import QuantumCircuit
--- a/src/qrisp/interface/simulators/qrisp_simulator_backend.py
+++ b/src/qrisp/interface/simulators/qrisp_simulator_backend.py
@@ -20,7 +20,7 @@
 
 from __future__ import annotations
 
-from collections.abc Sequence
+from collections.abc import Sequence
 from typing import cast
 
 from qrisp.circuit.pass_management.pass_manager import PassManager
--- a/src/qrisp/interface/simulators/qrisp_simulator_backend.py
+++ b/src/qrisp/interface/simulators/qrisp_simulator_backend.py
@@ -134,8 +134,6 @@ class QrispSimulatorBackend(Backend):
 
     Parameters
     ----------
-    options : Mapping or None, optional
-        Runtime options.  Defaults to ``{"shots": None, "token": ""}``.
     pm : PassManager or None, optional
         An optional :class:`~qrisp.circuit.pass_management.PassManager` that
         is applied to every circuit before it is submitted to the simulator.
@@ -277,31 +275,19 @@ class QrispSimulatorBackend(Backend):
 
     def __init__(
         self,
-        name: str | None = None,
-        options: Mapping | None = None,
         pm: PassManager | None = None,
-        **kwargs,
     ) -> None:
         """
         Initialize the QrispSimulatorBackend.
 
         Parameters
         ----------
-        name : str or None, optional
-            Optional user-defined name for the backend.
-            Defaults to the class name.
-        options : Mapping or None, optional
-            Runtime execution options for the backend.
-            If omitted, :meth:`_default_options` is used.
         pm : PassManager or None, optional
             An optional :class:`~qrisp.circuit.pass_management.PassManager`
             that is applied to every circuit before it is submitted to the
             simulator. Defaults to ``None``.
-        **kwargs :
-            Additional backend-specific parameters forwarded to the base
-            :class:`~qrisp.interface.Backend` initialiser.
         """
-        super().__init__(name=name, options=options, **kwargs)
+        super().__init__(name="QrispSimulator", options=None)
         if pm is not None and not isinstance(pm, PassManager):
             raise TypeError(
                 f"Expected a PassManager instance for 'pm', "
--- a/src/qrisp/interface/simulators/qrisp_simulator_backend.py
+++ b/src/qrisp/interface/simulators/qrisp_simulator_backend.py
@@ -21,12 +21,9 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Sequence, cast
+from typing import Sequence, cast
 
-if TYPE_CHECKING:
-    from qrisp.circuit.pass_management.pass_manager import PassManager
-
-from qrisp.circuit.pass_management.pass_manager import PassManager as _PassManager
+from qrisp.circuit.pass_management.pass_manager import PassManager
 from qrisp.circuit.quantum_circuit import QuantumCircuit
 from qrisp.interface.backend import Backend
 from qrisp.interface.job import Job, JobResult, JobStatus
@@ -305,7 +302,7 @@ class QrispSimulatorBackend(Backend):
             :class:`~qrisp.interface.Backend` initialiser.
         """
         super().__init__(name=name, options=options, **kwargs)
-        if pm is not None and not isinstance(pm, _PassManager):
+        if pm is not None and not isinstance(pm, PassManager):
             raise TypeError(
                 f"Expected a PassManager instance for 'pm', "
                 f"got {type(pm).__name__}."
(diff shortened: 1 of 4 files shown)


src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -132,6 +136,11 @@ class QrispSimulatorBackend(Backend):
     ----------
```
renezander90, line 136: Needs documentation of the `name` keyword argument of the `init` functions.

src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -130,8 +134,11 @@ class QrispSimulatorBackend(Backend):
 
     Parameters
     ----------
-    options : Mapping or None, optional
```
renezander90, line 133: Not sure if removing `options` is a good idea. `shots` is relevant also for the simulator backend.

positr0nium: The user can still specify the amount of shots manually. This is the default amount of shots. Making the default user-specifiable is contradictory in my opinion

src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -1,25 +1,29 @@
...
+# * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+# ********************************************************************************
+# """
 
 """This module defines :class:`QrispSimulatorBackend` and its associated :class:`QrispSimulatorJob`."""
 
+from __future__ import annotations
+
+from collections.abc import Mapping
 from typing import Sequence, cast
```
PietropaoloFrisoni:
```suggestion
from collections.abc Sequence
from typing import cast
```

I missed this in the first PR, but `Sequence` should come from `collections.abc` since Python 3.9: https://docs.python.org/3/library/typing.html#typing.Sequence

Also, it looks like `Mapping` is not used, so I would just remove it

src/qrisp/interface/simulators/qrisp_simulator_backend.py
PietropaoloFrisoni, line 1: The more I think about it, the more I convince myself that we should have a reference to `PassManager` in the abstract Backend class as well. What do you think @positr0nium @renezander90 ?

Maybe an `apply_passes` utility method which receives an empty (or default) pipeline by default, and the user has the chance to provide a custom passes pipeline when he/she instantiate the Backend (similar to how it is done in this PR for `QrispSimulatorBackend`, or how a default `CompilePipeline` is used [here](https://github.com/PennyLaneAI/pennylane/blob/3959a061960044e054b7d99dfa625b1d506758cd/pennylane/devices/device_api.py#L388)).

If we agree, we can still add it in the next release (it would most probably be a non-breaking change)

src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -230,6 +338,11 @@ def run_async(
             circuits = [circuits]
         else:
             circuits = list(circuits)
+
+        # Apply the pass manager (if any) to each circuit before simulation
+        if self._pm is not None:
+            circuits = [self._pm.run(c) for c in circuits]
```
PietropaoloFrisoni, line 344: My brain associates the `run` method to something that returns a `dict`-like `MeasurementResult` type, instead of a `QuantumCircuit`. I would have used `PassManager.apply_pass` instead of `PassManager.run` (just a preference, but nevermind at this point)

src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -159,6 +166,81 @@ class QrispSimulatorBackend(Backend):
...
+
+        ev_function = H.expectation_value(state_prep, backend=backend)
+        result = ev_function(np.pi/2)
+
+    .. code-block:: none
+
+               ┌─────────┐┌───┐┌─┐
+         qv.0: ┤ Ry(π/2) ├┤ H ├┤M├
+               ├─────────┤└┬─┬┘└╥┘
+         qv.1: ┤ Ry(π/2) ├─┤M├──╫─
```
PietropaoloFrisoni, line 220: If I execute this, I see `qf_0.X` instead of `qv.X` as we are using a `QuantumFloat`. I would suggest a fix in this tab, but this would mess up all the docstrings since we have a different number of characters. It's up to you whether you want to fix this as this is honestly very minor

PietropaoloFrisoni (commented):
🚀

src/qrisp/interface/simulators/qrisp_simulator_backend.py
positr0nium: Yes, I also thought about having this somehow in there. It would however represent a deviation from the principle that the backend architecture is minimally opinionated and leaves the backend provider the most freedom. Maybe we can mention that we recommend having a passmanager registered under the ``.pm`` attribute.

src/qrisp/interface/simulators/qrisp_simulator_backend.py
renezander90: Yes, `apply_pass` would be better. Gemini also prefers this :).

Bildschirmfoto 2026-06-09 um 13 25 35

positr0nium: okay, I will open a separate PR for this.

src/qrisp/interface/simulators/qrisp_simulator_backend.py
```diff
@@ -1,25 +1,29 @@
...
+# * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+# ********************************************************************************
+# """
 
 """This module defines :class:`QrispSimulatorBackend` and its associated :class:`QrispSimulatorJob`."""
 
-from typing import Sequence, cast
+from __future__ import annotations
 
+from collections.abc Sequence
```
PietropaoloFrisoni:
```suggestion
from collections.abc import Sequence
```

I forgot `import` 🤦 
Sorry about that

src/qrisp/interface/simulators/qrisp_simulator_backend.py
positr0nium: For me I get ``qv.0`` when I execute this 🤔
In static mode, Qrisp tries to infer the name of the Qubits from the name of the Python variable.

PietropaoloFrisoni (approved):
🚀

src/qrisp/interface/simulators/qrisp_simulator_backend.py
PietropaoloFrisoni: Yeah, there are many options. We can revise this in the next dev cycle

renezander90 (approved):

PR #627, opened 2026-06-08
Enhanced reflection

renezander90:
**Description:**
This PR refines the reflection operator primitive in `qrisp`. It updates the function signature with more precise type hinting and enhances its documentation for better clarity on how it operates across different quantum variable types (single variables, arrays, lists, tuples).

**Changes:**
*   **reflection.py**:
    *   Updated parameter type hints in the `reflection` function to support modern Python typing and handle various input structures (`QuantumVariable`, `QuantumArray`, `Sequence`).
    *   Improved docstrings for better readability and corrected a typo ("refection" -> "reflection").

**Tests:**
The following new tests have been added to test_reflection.py to ensure the reflection primitive works correctly across different input types:
*   `test_reflection_quantum_variable`: Validates reflection with a single `QuantumVariable`.
*   `test_reflection_quantum_array`: Validates reflection when the input is a `QuantumArray`.
*   `test_reflection_list_quantum_variable`: Tests reflection using a list of `QuantumVariables`.
*   `test_reflection_tuple_quantum_variable`: Tests reflection using a tuple of `QuantumVariables`.
*   `test_reflection_list_quantum_varaible_quantum_array`: Validates mixed input types (list containing both `QuantumVariable` and `QuantumArray`).
*   `test_jasp_reflection`: Verifies the primitive's functionality within the JASP environment.

### DIFF
--- a/src/qrisp/alg_primitives/reflection.py
+++ b/src/qrisp/alg_primitives/reflection.py
@@ -17,8 +17,8 @@
 """
 
 from collections.abc import Callable, Sequence
-import numpy as np
 from typing import Any
+import numpy as np
 
 from qrisp import (
     QuantumArray,
@@ -64,7 +64,7 @@ def reflection(
     state_function : Callable, optional
         A Python function ``state_function(*qargs, *args, **kwargs)`` preparing the state $\ket{\psi}$ in variables ``qargs`` around which to reflect.
         By default, the reflection is performed around the $\ket{0}$ state.
-    args : Sequence, optional
+    args : Sequence[Any], optional
         Additional arguments for the state function.
     kwargs : dict, optional
         Keyword arguments for the state function.
@@ -191,7 +191,8 @@ def reflection(
     if reflection_indices is None:
         reflection_indices = range(len(flattened_qargs))
 
-    qubits = sum([flattened_qargs[i].reg for i in reflection_indices], [])
+    indices = reflection_indices if reflection_indices is not None else range(len(flattened_qargs))
+    qubits = [q for i in indices for q in flattened_qargs[i].reg]
 
     # Reflection around |0> state
     def inner_reflection(qubits, phase):
--- a/src/qrisp/alg_primitives/reflection.py
+++ b/src/qrisp/alg_primitives/reflection.py
@@ -188,9 +188,6 @@ def reflection(
         else:
             raise TypeError("Arguments must be of type QuantumVariable or QuantumArray")
 
-    if reflection_indices is None:
-        reflection_indices = range(len(flattened_qargs))
-
     indices = reflection_indices if reflection_indices is not None else range(len(flattened_qargs))
     qubits = [q for i in indices for q in flattened_qargs[i].reg]
 
--- a/src/qrisp/alg_primitives/reflection.py
+++ b/src/qrisp/alg_primitives/reflection.py
@@ -189,7 +189,7 @@ def reflection(
             raise TypeError("Arguments must be of type QuantumVariable or QuantumArray")
 
     indices = reflection_indices if reflection_indices is not None else range(len(flattened_qargs))
-    qubits = [q for i in indices for q in flattened_qargs[i].reg]
+    qubits = sum([flattened_qargs[i].reg for i in indices], [])
 
     # Reflection around |0> state
     def inner_reflection(qubits, phase):
--- a/src/qrisp/alg_primitives/reflection.py
+++ b/src/qrisp/alg_primitives/reflection.py
@@ -16,7 +16,10 @@
 ********************************************************************************
 """
 
+from collections.abc import Callable, Sequence
 import numpy as np
+from typing import Any
+
 from qrisp import (
     QuantumArray,
     QuantumVariable,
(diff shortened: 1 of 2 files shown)


src/qrisp/alg_primitives/reflection.py
```diff
@@ -16,7 +16,10 @@
 ********************************************************************************
 """
 
+from collections.abc import Callable, Sequence
 import numpy as np
+from typing import Any
```
PietropaoloFrisoni, line 20:
```suggestion
from collections.abc import Callable, Sequence
from typing import Any
import numpy as np
```

src/qrisp/alg_primitives/reflection.py
```diff
@@ -51,18 +59,18 @@ def reflection(
...
-    qargs : QuantumVariable | QuantumArray | list[QuantumVariable | QuantumArray]
-        The (list of) QuantumVariables representing the state to apply the reflection on.
-    state_function : function, optional
+    qargs : QuantumVariable | QuantumArray | Sequence[QuantumVariable | QuantumArray]
+        The quantum variable, array, or collection thereof on which the reflection is performed.
+    state_function : Callable, optional
         A Python function ``state_function(*qargs, *args, **kwargs)`` preparing the state $\ket{\psi}$ in variables ``qargs`` around which to reflect.
         By default, the reflection is performed around the $\ket{0}$ state.
-    args : tuple, optional
+    args : Sequence, optional
```
PietropaoloFrisoni:
```suggestion
    args : Sequence[Any], optional
```

Optional suggestion

src/qrisp/alg_primitives/reflection.py
PietropaoloFrisoni:
```suggestion
    indices = reflection_indices if reflection_indices is not None else range(len(flattened_qargs))
    qubits = [q for i in indices for q in flattened_qargs[i].reg]
```

Instead of building the qubit list by repeatedly concatenating lists (O(n²) in the number of qubits), this flat comprehension is O(n).

PietropaoloFrisoni (commented):
🚀

src/qrisp/alg_primitives/reflection.py
```diff
@@ -177,7 +191,8 @@ def perturbed_ghz(qv, a, b):
     if reflection_indices is None:
         reflection_indices = range(len(flattened_qargs))
```
renezander90:
```suggestion
```

src/qrisp/alg_primitives/reflection.py
renezander90: I reverted to `qubits = sum([flattened_qargs[i].reg for i in indices], [])`. The testing pipeline otherwise got stuck in the `qsp_tests`. Claude's explanation is:

In JASP tracing mode, `QuantumVariable.reg` is a `DynamicQubitArray`, not a plain Python `list`. `DynamicQubitArray` wraps a JAX tracer and implements `__add__` / `__radd__` to **fuse the underlying tracer** (via `fuse_qb_array`), returning another `DynamicQubitArray`.

### Why `sum(...)` is fast

```python
sum([flattened_qargs[i].reg for i in indices], [])
```

The `sum` starts with `[]` (Python list) and calls `DynamicQubitArray.__radd__([])` for each register. This calls `fuse_qb_array` at the JAX primitive level and returns a single `DynamicQubitArray` wrapping a fused abstract qubit array. The result is one `DynamicQubitArray` object with a `.size` property backed by a `get_size` JAX primitive.

When `jlen` sees a `DynamicQubitArray` (not a `list`), it calls `x.size` → `get_size(self.tracer)` — which returns a **static integer** derived from the tracer's abstract value type (known at trace time from the `QuantumVariable`'s declared size).

### Why the list comprehension is slow

```python
[q for i in indices for q in flattened_qargs[i].reg]
```

`DynamicQubitArray.__getitem__` is called for each integer index (via Python's iteration protocol). The result is a **plain Python `list`** of `AbstractQubit` objects. When `jlen` sees a plain `list`, it returns `len(list)` — a static Python int — but this is only as fast as the construction of that list.

The real problem is what happens downstream: `qubits[:-1]` on a plain Python list returns a plain Python list, and `qubits[-1]` returns a single `AbstractQubit`. These are passed to `mcx`. But more critically, the construction of the list by calling `DynamicQubitArray.__getitem__(0)`, `__getitem__(1)`, ... up to `n` times calls `get_qubit(self.tracer, key)` **n individual times**. Each `get_qubit` creates a new JAX primitive binding in the trace. For QSP, `reflection` is called inside `jrange` loops (e.g., `chebyshev` calls it `k//2` times), so `reflection` is unrolled by JAX. If the target quantum variable has `n` qubits, each unrolled call materializes `n` individual `get_qubit` primitives. With high polynomial orders (large `k`) and multi-qubit variables, this creates **O(k × n) individual JAX primitives** per trace, vs **O(1) fused array primitives** with `sum`. The expression graph grows quadratically or worse, consuming memory and time.

### Fix

The `sum` approach is not just stylistically different — it is semantically the correct way to combine `DynamicQubitArray` objects in JASP because it uses their `__radd__` and keeps the result as a single fused abstract array. The list comprehension bypasses this and degenerates into per-qubit indexing operations.

positr0nium: Nice Job Rene & Claude 😊 One minor comment here: It would be even faster to concatenate the DynamicQubitArrays in a tree-like structure. This would reduce the amount of concatenations to log(N) instead of N

PietropaoloFrisoni: Thanks @renezander90 ! Very useful to know indeed

P.S. I admit that this time I didn't try the suggestion locally, but I limited myself to the changed files

PietropaoloFrisoni (approved):
Great stuff

positr0nium:
Ahh one more comment: In the case of a Quantum array, you directly access the flattened DynamicQubitArray through the corresponding attribute 

Edit: I'm wrong, the attribute of the QuantumArray, doesn't always contain the corresponding dynamic qubit array. If the underlying QuantumArray is the result of a slicing, this doesn't hold anymore

PR #639, opened 2026-06-10
Implementing `reviewdog` in `qrisp`

PietropaoloFrisoni:
## Description

Introduces [reviewdog](https://github.com/reviewdog/reviewdog) as a non-blocking CI check on pull requests. When a PR is opened, ruff runs over `src/` and reviewdog posts inline review comments directly on the changed lines in the "Files changed" tab.

The key design constraint is that **only lines introduced in the PR are checked** (`-filter-mode=added`). Pre-existing issues in the codebase are ignored entirely. The check is also **non-blocking** (`-fail-level=none`): it annotates but never prevents a merge.

The motivation is to reduce reviewer time spent on style issues and coding standards that can be caught automatically and consistently by tooling.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [x] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Added `.github/workflows/reviewdog.yml`: triggers on PRs to `main`, installs ruff and reviewdog, runs `ruff check src/ --output-format=rdjson` piped to reviewdog with `filter-mode=added` and `fail-level=none`
- The changed in `src/qrisp/algorithms/grover/grover_tools.py` are just a consequence of formatting.

## How was it tested?

Validated via a draft PR. The workflow ran successfully and produced no false positives (the PR itself only modifies a YAML file, so no Python lines were in scope). A dedicated end-to-end test with a deliberate ruff violation can be done on the next PR that touches Python source files.

## Checklist

## Reviewer Notes

- `filter_mode=added` is the critical guard: reviewdog discards any finding not on a newly added line, so the existing backlog of issues is invisible to this check
- `fail_level=none` means this will never block a merge — it only annotates; the intention is to tighten this once the codebase is in better shape
- Ruff runs with implicit defaults for now; a `[tool.ruff]` config in `pyproject.toml` will be added as part of the upcoming dedicated ruff PR

## Security Review (I asked Claude to do this)

This workflow was reviewed for supply-chain and CI-privilege risks. Summary: **no security concerns identified.**

- **Uses `pull_request`, not `pull_request_target`.** This is the critical safety property: for PRs from forks, GitHub automatically downgrades `GITHUB_TOKEN` to read-only at runtime, regardless of the `permissions:` block declared in the workflow. A fork PR therefore cannot use this token to write anything, even though `pull-requests: write` is requested for same-repo PRs. The dangerous pattern this avoids is `pull_request_target` + checkout of untrusted fork code, which runs attacker-controlled code with a privileged, secret-bearing token.
- **`secrets.GITHUB_TOKEN` scope is minimal.** It's the automatic per-run token (not a personal/org secret), used only so reviewdog can post inline review comments.
- **No execution of untrusted PR content.** `ruff check` only lints the repository's own source; nothing in the pipeline evals or shells out to PR-supplied input.
- **Third-party action pinning.** `reviewdog/action-setup` is pinned to a full commit SHA (not a mutable tag), specifically because its `v1` tag was hijacked in a real supply-chain attack in March 2025 that exfiltrated CI secrets. `actions/checkout` and `actions/setup-python` are pinned to major version tags (`@v7`), consistent with the rest of the repo's workflows.
- **Blast radius is capped even in a worst case.** The check is `continue-on-error: true` and non-blocking (`-fail-level=none`), and now scoped via `paths` to `src/**` and the workflow file itself — it can't silently affect unrelated CI behavior or run on every PR regardless of content.

**Conclusion:** this PR does not weaken CI security posture; it follows the secure `pull_request`-only pattern and pins the one dependency with a known supply-chain incident.


.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,39 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  ruff:
+    name: ruff
```
purva-thakre, line 20: you can directly add this as an action: https://github.com/astral-sh/ruff-action

I am planning to add all the ruff rules next week. you can keep it to the basics for now otherwise the other PRs will have merge conflicts.

PietropaoloFrisoni: Thanks, `astral-sh/ruff-action` is simpler, but it has no `diff-filtering` (or, at least, so it seems to me). 

Implementing it as an action would report all violations across the entire codebase, including pre-existing ones. The `reviewdog` pipe is intentional and `-filter-mode=added` ensures only lines introduced in a PR are checked, which is a hard requirement given the current state of the codebase.

purva-thakre: Ah, gotcha. Then feel free to make the changes you want. I will have to go over the review-dog configuration settings. 

BTW if you have the time, type checking could also be enabled: https://github.com/astral-sh/ty

src/qrisp/algorithms/grover/grover_tools.py
```diff
@@ -16,6 +16,9 @@
 ********************************************************************************
 """
 
+# Test for reviewdog
+import os
+
```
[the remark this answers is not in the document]
PietropaoloFrisoni: Good job!

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+
+      - uses: reviewdog/action-setup@v1.5.0
+
+      - name: Install ruff
+        run: pip install ruff
```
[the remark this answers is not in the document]
PietropaoloFrisoni: I think it is reasonable to use the latest `ruff`. My understanding is that a dedicated `ruff` integration PR is planned that will establish the full rule set and configuration. Pinning at that point makes more sense, and I would not postpone this change for this reason

PietropaoloFrisoni:
@renezander90 Thanks for the security update, now we are using the commit SHA instead of the version

purva-thakre:
@PietropaoloFrisoni @positr0nium @renezander90 What is the plan for this tool? Do we want to add this? The PR has been open for quite sometime now....

I wonder if it might be helpful to also add type checking for the changes in a PR without worrying about the entire codebase i.e. I want the workflow to fail if the PR adds new functions that do not have any type hints. The same comment has been made on multiple PRs so far that states 'yes, qrisp lacks type hinting but we want to make sure we add type hints for the new features we add'

positr0nium:
If I recall correctly we benched it because of the risk that the tool could inject malicious suggestions that get approved by a careless reviewer? Couldn't we somehow configure it to only give comments but no suggestions?

PietropaoloFrisoni:
Thanks for bringing this PR back from the dead!

This was discussed (and informally accepted, since no one objected) during one of the Qrisp meetings a few months ago. It was also brought up on Discord, and as far as I remember, we all agreed to proceed with this change (the security aspect that was raised at some point was addressed in [this commit](https://github.com/eclipse-qrisp/Qrisp/pull/639/changes/e271aea02b23301b3e61b85daa22d0b73f0dabbc)). I have been waiting for reviews ever since : )

I do not know if there was further discussion on this while I was away, but as I have emphasized several times, I consider this aspect very important because we currently have no check that verifies the quality of new source code. From my perspective, this PR isn't sufficient at all (in most big libraries, the code must pass very strict quality/coverage tests before the CI turns green), but it is a minimal starting point.

@renezander90 @positr0nium Since we already had many discussions around this back in the day, can we please get this in and close this ticket? If we don't like this for some reason, we can always revert : )

positr0nium (approved):

purva-thakre:
When you were away, only `ruff` as a formatter was added for the entire CI. We are waiting for Niko's students to fix the issues flagged by the linter before we turn it on for the project. So, using reviewdog on ruff as a linter for just the changes in a PR will be helpful. 

Also, if it is not too much to ask, could you please add a type checker to reviewdog? I often have to make a comment about this when reviewing a PR.

PietropaoloFrisoni:
> When you were away, only ruff as a formatter was added for the entire CI. We are waiting for Niko's students to fix the issues flagged by the linter before we turn it on for the project. So, using reviewdog on ruff as a linter for just the changes in a PR will be helpful.

Good, that was exactly the purpose : )

> Also, if it is not too much to ask, could you please add a type checker to reviewdog? I often have to make a comment about this when reviewing a PR.

I can take a look as soon as I have time, but could we please merge this first and then add type hints in a second stage (hoping it does not take to much time)? This PR has been open for > 2 months and we had so many discussions around this :sweat_smile: 

@renezander90 Please let me know if you have further comments or concerns, otherwise I can merge this myself. Thanks!

renezander90 (approved):

PietropaoloFrisoni:
One final check for security before merging (I asked Claude to perform a careful investigation with max reasoning power):

___
This workflow was reviewed for supply-chain and CI-privilege risks. Summary: **no security concerns identified.**

- **Uses `pull_request`, not `pull_request_target`.** This is the critical safety property: for PRs from forks, GitHub automatically downgrades `GITHUB_TOKEN` to read-only at runtime, regardless of the `permissions:` block declared in the workflow. A fork PR therefore cannot use this token to write anything, even though `pull-requests: write` is requested for same-repo PRs. The dangerous pattern this avoids is `pull_request_target` + checkout of untrusted fork code, which runs attacker-controlled code with a privileged, secret-bearing token.
- **`secrets.GITHUB_TOKEN` scope is minimal.** It's the automatic per-run token (not a personal/org secret), used only so reviewdog can post inline review comments.
- **No execution of untrusted PR content.** `ruff check` only lints the repository's own source; nothing in the pipeline evals or shells out to PR-supplied input.
- **Third-party action pinning.** `reviewdog/action-setup` is pinned to a full commit SHA (not a mutable tag), specifically because its `v1` tag was hijacked in a real supply-chain attack in March 2025 that exfiltrated CI secrets. `actions/checkout` and `actions/setup-python` are pinned to major version tags (`@v7`), consistent with the rest of the repo's workflows.
- **Blast radius is capped even in a worst case.** The check is `continue-on-error: true` and non-blocking (`-fail-level=none`), and now scoped via `paths` to `src/**` and the workflow file itself — it can't silently affect unrelated CI behavior or run on every PR regardless of content.

**Conclusion:** this PR does not weaken CI security posture; it follows the secure `pull_request`-only pattern and pins the one dependency with a known supply-chain incident.
___

Issue #638, opened 2026-06-10
`dot()` fails with `'QuantumArrayIterator' object is not iterable`


purva-thakre:
Came across this while working on #637 

```python
from qrisp import QuantumFloat, QuantumArray, dot

qf = QuantumFloat(3)
a = QuantumArray(qf, shape=2)
b = QuantumArray(qf, shape=(2, 2))
a[:] = [2, 3]
b[:] = [[0, 2], [1, 0]]

dot(a, b)
# TypeError: 'QuantumArrayIterator' object is not iterable
```
Broken docs:

- documentation/source/reference/Examples/MatrixMultiplication.rst
- src/qrisp/alg_primitives/arithmetic/matrix_multiplication.py docstring

PR #641, opened 2026-06-11
fix: add __iter__ to QuantumArrayIterator to satisfy iterator protocol

Ishant5436:
hey, spotted issue #638 while browsing the repo and figured it was a small fix.

the root cause is that `QuantumArrayIterator` defines `__next__` but not `__iter__`, so python doesn't recognise it as a proper iterator. calling `list()` on it (or using it anywhere that expects an iterable) raises the `'QuantumArrayIterator' object is not iterable` error you see in the bug report.

the fix is just adding `__iter__` returning `self`, which is the standard pattern for iterator classes in python.

reproduced the original traceback with the snippet from the issue before and confirmed it passes after the fix. let me know if anything else needs adjusting!

### DIFF
--- a/src/qrisp/core/quantum_array.py
+++ b/src/qrisp/core/quantum_array.py
@@ -2280,6 +2280,9 @@ class QuantumArrayIterator:
         self.qa = qa
         self.counter = -1
 
+    def __iter__(self):
+        return self
+
     def __next__(self):
         self.counter += 1
         if self.counter >= self.qa.size:


purva-thakre:
Correct me if I am wrong but I think this PR can be closed as the changes are already in #642

Ishant5436:
ah, you're right. closing this in favor of #642. thanks!

PR #647, opened 2026-06-11
Enhanced from_lcu constructor

renezander90:
## Description

Extends `BlockEncoding.from_lcu` to support **complex and negative coefficients**. Previously, the LCU constructor rejected any non-positive coefficient with a `ValueError`. It now handles the general complex case via an asymmetric PREP_R–SEL–PREP_L† circuit, while preserving the original symmetric PREP–SEL–PREP† fast path for real non-negative coefficients. The normalization factor is updated to $\alpha = \sum_i |\alpha_i|$.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **`constructors/from_lcu.py`**:
  - Removed `ValueError` for negative coefficients; `from_lcu` now accepts any complex array.
  - Normalization updated from $\alpha = \sum_i \alpha_i$ to $\alpha = \sum_i |\alpha_i|$.
  - All inputs are internally cast to `complex` (`complex_coeffs = np.array(coeffs, dtype=complex)`).
  - Module-level constant `_TOLERANCE = 1e-12` introduced.
  - Added **fast path** for real non-negative coefficients: unchanged symmetric `PREP–SEL–PREP†` circuit.
  - Added **general path** for complex/negative coefficients: asymmetric `PREP_R–SEL–PREP_L†` circuit, where `PREP_R` prepares $\sqrt{\alpha_i/\alpha}$ and `PREP_L` prepares $\sqrt{\alpha_i/\alpha}^*$.
  - Single-unitary short-circuit (`m == 1`) now extracts the real part of the coefficient if its imaginary part is within tolerance, and raises `ValueError` for genuinely complex single coefficients.
  - Added internal helper `_is_real_non_negative_array` to select between the two circuit paths.
  - Moved `prepare` and `q_switch` imports from inside the function body to module level.
  - Updated docstring: coefficient type updated to complex, `Raises` section updated, normalization note updated, math extended to include the asymmetric PREP_R/PREP_L variant.
- **`constructors/from_projector.py`**: minor docstring fixes (`Defauts` → `Defaults`, clarified `Returns` to cover both `kernel=True` and `kernel=False`).
- **`constructors/from_eye.py`**: minor docstring wording fix.

## How was it tested?

New test file test_block_encoding_from_lcu.py. The main test uses a **swap test** to verify the block-encoded output matches the classical $Ab/\|Ab\|$ for a banded circulant matrix, parametrized over 5 coefficient types:

| Test-ID | Coefficient type | Status |
|---------|-----------------|--------|
| `test_block_encoding_from_lcu[coeffs0]` | Real non-negative `[1.4, 1.1, 0.5, 0.3]` | ✅ |
| `test_block_encoding_from_lcu[coeffs1]` | Mixed real with negative `[0.8, -0.6, 0.2, 0.1j]` | ✅ |
| `test_block_encoding_from_lcu[coeffs2]` | Complex `[1.0+0.5j, 0.9j, 0.4, 0.2]` | ✅ |
| `test_block_encoding_from_lcu[coeffs3]` | Purely imaginary `[-0.6j, 0.9j, 0.4, 0.7+0.2j]` | ✅ |
| `test_block_encoding_from_lcu[coeffs4]` | Mixed imaginary/negative real `[0.3j, 0.9j, -0.4, 0.2]` | ✅ |
| `test_block_encoding_from_lcu_single_unitary_raises_value_error` | `ValueError` for complex single-unitary coefficient | ✅ |

Full block-encodings test suite (`pytest -n 4`) passes without regressions.

## Checklist

## Reviewer Notes

The test uses a **swap test** rather than direct amplitude comparison, which correctly verifies the output state up to a global phase — appropriate for verifying quantum state equality after a RUS procedure.

### DIFF
--- a/src/qrisp/block_encodings/constructors/from_lcu.py
+++ b/src/qrisp/block_encodings/constructors/from_lcu.py
@@ -144,11 +144,17 @@ def build_from_lcu(
     if m == 1:
         if np.abs(complex_coeffs[0].imag) < _TOLERANCE:
             return cls(
-                complex_coeffs[0].real, [], unitaries[0], num_ops=num_ops, is_hermitian=is_hermitian
+                complex_coeffs[0].real,
+                [],
+                unitaries[0],
+                num_ops=num_ops,
+                is_hermitian=is_hermitian,
             )
-        
-        raise ValueError("For a single unitary, the coefficient must be real (up to numerical precision).")
-    
+
+        raise ValueError(
+            "For a single unitary, the coefficient must be real (up to numerical precision)."
+        )
+
     # Block encoding of a linear combination of unitaries via the LCU protocol
     # If all coefficients are real and non-negative: LCU = PREP SEL PREP_dg
     if _is_real_non_negative_array(complex_coeffs):
@@ -166,8 +172,8 @@ def build_from_lcu(
             num_ops=num_ops,
             is_hermitian=is_hermitian,
         )
-    
-     # If coefficients are complex or negative, we use a state preparation pair (PREP_R, PREP_L): LCU = PREP_R SEL PREP_L_dg
+
+    # If coefficients are complex or negative, we use a state preparation pair (PREP_R, PREP_L): LCU = PREP_R SEL PREP_L_dg
     complex_coeffs_r = np.sqrt(complex_coeffs / alpha)
     complex_coeffs_l = np.sqrt(complex_coeffs / alpha).conjugate()
 
@@ -188,13 +194,15 @@ def build_from_lcu(
     )
 
 
-def _is_real_non_negative_array(arr: npt.NDArray[np.number], tol: float=_TOLERANCE) -> bool:
+def _is_real_non_negative_array(
+    arr: npt.NDArray[np.number], tol: float = _TOLERANCE
+) -> bool:
     """Checks if all entries in an array are non-negative and have negligible imaginary parts."""
     # 1. Check if the array is a complex type
     if np.issubdtype(arr.dtype, np.complexfloating):
         abs_imag_part = np.abs(arr.imag)
         real_part = arr.real
-        
+
         # Check if imaginary parts are within tolerance AND real parts are non-negative
         return (abs_imag_part < tol).all() and (real_part >= 0).all()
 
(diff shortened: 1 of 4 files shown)


src/qrisp/block_encodings/constructors/from_lcu.py
```diff
@@ -42,57 +47,71 @@ def build_from_lcu(
 
         O = \sum_{i=0}^{M-1} \alpha_i U_i
 
-    where $\alpha_i$ are real non-negative coefficients such that $\sum_i \alpha_i = \alpha$,
+    where $\alpha_i$ are complex coefficients, $\alpha = \sum_i |\alpha_i|$,
```
positr0nium, line 50: It would be nice to have some reference here for both the strictly positive and the complex case.

src/qrisp/block_encodings/constructors/from_lcu.py
```diff
@@ -42,57 +47,71 @@ def build_from_lcu(
...
+
+        \text{PREP}_R \ket{0}_a = \sum_{i=0}^{M-1} \sqrt{\frac{\alpha_i}{\alpha}} \ket{i}_a, \qquad
+        \text{PREP}_L \ket{0}_a = \sum_{i=0}^{M-1} \sqrt{\frac{\alpha_i}{\alpha}}^* \ket{i}_a
+
     Parameters
     ----------
-    coeffs : ArrayLike
-        1-D array of non-negative coefficients $\alpha_i$.
+    coeffs : np.ndarray
+        1-D array containing the complex coefficients $\alpha_i$.
```
positr0nium: It might be worth considering having some ``ensure_positive`` keyword argument. With the current "automatic" decision, a user might intend to submit positive ``coeffs`` but due to a bug submits complex or negative coefficients. With the current interface, this would silently degrade performance.

renezander90: I would rather keep the API minimal. `ensure_positive` would have of be set to `False` by default, because complex inputs are valid. If a user inputs an array with data type complex, but all imaginary parts are close to zero, it gets treated as real. Hence, `ensure_positive` is equivalent of the user using an extra `np.abs`. But, we could note in the documentation that the performance will be better for real non-negative coefficients.

positr0nium (approved):
🎉

PR #656, opened 2026-06-15
Add `py.typed` marker

PietropaoloFrisoni:
## Description

Adds a `py.typed` marker file (PEP 561) so that type checkers such as `mypy` and `pyright` recognise that Qrisp ships inline type annotations and should analyse them, rather than treating the package as untyped. Also registers the marker in `setup.py` via `package_data` so it is included in the built wheel and installed distributions.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [x] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Added empty `src/qrisp/py.typed` marker file as required by PEP 561
- Added `package_data={"qrisp": ["py.typed"]}` to `setup.py` so the marker is included in the wheel

## How was it tested?

This change has no runtime behaviour. Correct inclusion in the wheel can be verified by building a distribution (`python -m build`) and inspecting its contents with `unzip -l dist/*.whl | grep py.typed`.

| Test-ID | Status |
|---------|--------|

N/A

## Checklist

## Reviewer Notes

The `py.typed` file must reach the installed package directory; without the `package_data` entry in `setup.py`, setuptools would silently omit it from the wheel even though the file exists in the source tree. No functional code was changed.


PietropaoloFrisoni:
Hi @smite , please let me know if I forgot something. Thanks!

setup.py
```diff
@@ -34,5 +34,6 @@
     ],
     package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
-    python_requires=">=3.11, <3.14",
+    package_data={"qrisp": ["py.typed"]},
+    python_requires=">=3.11", <3.14,
```
PietropaoloFrisoni:
```suggestion
    python_requires=">=3.11, <3.14",
```

purva-thakre:
@PietropaoloFrisoni Is this PR ready to be merged? just wanted to check because I want to test adding `ty` as a type checker to the testing pipeline.

PietropaoloFrisoni:
Yes, it is still waiting for review

setup.py
```diff
@@ -34,5 +34,6 @@
     ],
     package_dir={"": "src"},
     packages=setuptools.find_packages(where="src"),
+    package_data={"qrisp": ["py.typed"]},
```
purva-thakre, line 37: question: should this be moved to the `pyproject.toml` file? I am planning to get rid of the `setup.py` file soon. 

Plus, my local environment has `setuptools==82.0.1`. According to the `setuptools` docs, these files should have been included by default after `v69`. https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html

The question then becomes why do we need to manually add these?

purva-thakre: or maybe `setup.py` file is worth keeping for `setuptools`? I need to read over the docs. 😅 

https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#should-setup-py-be-deleted

Issue #632, opened 2026-06-08
Controlled-gate to pytket conversion has context-dependent qubit ordering


ryanhill1:
**Version:** 0.8.2

Converting a controlled gate to pytket (`to_pytket`) produces a **context-dependent qubit ordering**: the same gate converts correctly in isolation, but its qubit ordering changes after other controlled gates have been constructed in the same session.

Example with `MCRXGate` (3 controls): in isolation the RX target lands on the expected qubit, but after building other controlled gates the target is emitted on a different qubit index — a qubit **permutation** (not a reversal/endianness flip). The unitary is otherwise correct, so it's purely an ordering issue, but it makes cross-framework equivalence checks non-deterministic.

```python
# MCRXGate(control_amount=3) -> to_pytket
# isolation:     RX block at states {1110, 1111}
# after others:  RX block at states {1011, 1111}   # target moved to a different qubit
```

The conversion's qubit ordering should be deterministic / independent of prior construction (ideally label-preserving).

> Discovered during PR review integrating Qrisp as a supported framework in the qBraid-SDK: https://github.com/qBraid/qBraid/pull/1211

PR #662, opened 2026-06-16
Deprecate `pytket` converter

### LINKED ISSUE #630: to_pytket maps controlled-phase (cp) to CRz instead of controlled-phase
**Version:** 0.8.2

In `src/qrisp/interface/converter/pytket_converter.py`, the controlled-phase gate `cp` is mapped to `OpType.CRz`:

```python
elif op.name == "cp":
    if hasattr(op, "ctrl_state"):
        tket_ins = OpType.CRz   # <-- controlled-Rz, not controlled-phase
```

`CRz != CP`. Controlled-phase is `diag(1,1,1,e^{iλ})`, but the conversion yields `diag(1,1,e^{-iλ/2},e^{iλ/2})` — an **observable** error (not a global phase), inconsistent with qrisp's own `get_unitary()` for `cp`:

```python
import qrisp
from qrisp.circuit.library import CPGate
qc = qrisp.QuantumCircuit(2); qc.append(CPGate(0.7), [0, 1])
# diag(qc.get_unitary())        -> [1, 1, 1, e^{0.7i}]
# diag(qc.to_pytket unitary)    -> [1, 1, e^{-0.35i}, e^{0.35i}]
```

**Fix:** map `cp` to `OpType.CU1` (pytket's controlled-phase).

> Discovered during PR review integrating Qrisp as a supported framework in the qBraid-SDK: https://github.com/qBraid/qBraid/pull/1211

### LINKED ISSUE #631: to_pytket divides the u1 gate angle by pi twice
**Version:** 0.8.2

In `src/qrisp/interface/converter/pytket_converter.py`, the `u1` angle is divided by π twice. `"u1"` is already in the list that scales params by `1/π`:

```python
if op.name in ["cp","p","rx","rz","ry","rxx","rzz","ryy","u1","u3"]:
    params = [index / np.pi for index in params]
```

and then the `u1` branch divides again (and the source already flags it):

```python
elif op.name == "u1":
    params[0] = params[0] / np.pi
    # bugged
    tket_ins = OpType.Rz
```

So `U1(0.7)` converts to `Rz(0.7/π²)` ≈ `Rz(0.223 rad)` instead of the intended angle (a plain `RZGate(0.7)` correctly gives `0.7/π` half-turns = 0.7 rad).

**Fix:** remove the redundant `/ np.pi` in the `u1` branch (and ideally map to `OpType.U1`).

> Discovered during PR review integrating Qrisp as a supported framework in the qBraid-SDK: https://github.com/qBraid/qBraid/pull/1211

alighazi288:
## Description

Two `to_pytket` gate-mapping fixes in the same converter, with deterministic unitary regression tests.

## Related Issues

- Closes #631
- Closes #630 

## Type of Change

- [x] Bug Fix

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **#631**: removed the duplicate `params[0] / np.pi` for `u1` (it's already π-converted in the shared block).
- **#630**: `cp` now maps to `OpType.CU1` (controlled-phase) instead of `OpType.CRz`.
- Added `test_to_pytket_u1_angle` and `test_to_pytket_cp_controlled_phase`: convert the gate, take the pytket `get_unitary()`, compare to the expected diagonal up to global phase. Guarded with `pytest.importorskip("pytket")`, matching `test_to_pytket_basic`.

### Verification (local, pytket installed)
u1(0.7):  before diag(1, e^{i·0.223})  →  after diag(1, e^{i·0.7})  ✅
cp(0.7):  before diag(1,1, e^{∓i·0.35}) (CRz)  →  after diag(1,1,1, e^{i·0.7})  ✅
`pytest tests/interface_tests/test_tket_converter.py` → 2 passed; full interface suite 638 passed, 0 failed.

## Checklist

## Reviewer Notes

The new tests use `pytest.importorskip("pytket")`, so they **skip** on the GitHub Actions check and run on GitLab CI / locally, same as the existing `test_to_pytket_basic`. If you'd like the GitHub workflow to actually exercise the converter, adding `pip install pytket` to `.github/workflows/qrisp_test.yml` would do it (only `test_to_pytket_basic` + these two tests would newly run). Happy to add that if you want.

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -54,9 +54,11 @@ Other New Features
 Bug Fixes
 ---------
 
-* Fixed the ``to_pytket`` converter: ``u1`` angles are no longer divided by
-  ``pi`` twice, ``cp`` now maps to controlled-phase (``CU1``) instead of
-  ``CRz``, and global-phase (``gphase``) gates are now supported
+* Fixed and refactored the ``to_pytket`` converter: ``u1`` angles are no longer
+  divided by ``pi`` twice, ``cp`` now maps to controlled-phase (``CU1``) instead
+  of ``CRz``, and global-phase (``gphase``) gates are now supported. The gate
+  mapping was consolidated into a lookup table, dead code removed, and converter
+  unit-test coverage substantially increased
   (`PR #662 <https://github.com/eclipse-qrisp/Qrisp/pull/662>`_).
 
 * Fixed a bug where :func:`dot <qrisp.dot>` failed with a
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -56,9 +56,14 @@ Bug Fixes
 
 * Fixed and refactored the ``to_pytket`` converter: ``u1`` angles are no longer
   divided by ``pi`` twice, ``cp`` now maps to controlled-phase (``CU1``) instead
-  of ``CRz``, and global-phase (``gphase``) gates are now supported. The gate
-  mapping was consolidated into a lookup table, dead code removed, and converter
-  unit-test coverage substantially increased
+  of ``CRz``, ``p``/``u1`` map to the relative-phase ``U1`` (phase-exact, no
+  longer leaking a global phase into controlled boxes), a flipped ``ctrl_state``
+  on the elementary controlled gates (``cx``/``cy``/``cz``/``cp``) is now
+  respected, global-phase (``gphase``) gates are supported, and ``barrier``,
+  ``reset`` and classically-controlled (``c_if``) gates now convert instead of
+  raising. The gate mapping was consolidated into a lookup table, the CircBox
+  construction de-duplicated, dead code removed, and converter unit-test coverage
+  substantially increased
   (`PR #662 <https://github.com/eclipse-qrisp/Qrisp/pull/662>`_).
 
 * Fixed a bug where :func:`dot <qrisp.dot>` failed with a
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -54,6 +54,11 @@ Other New Features
 Bug Fixes
 ---------
 
+* Fixed the ``to_pytket`` converter: ``u1`` angles are no longer divided by
+  ``pi`` twice, ``cp`` now maps to controlled-phase (``CU1``) instead of
+  ``CRz``, and global-phase (``gphase``) gates are now supported
+  (`PR #662 <https://github.com/eclipse-qrisp/Qrisp/pull/662>`_).
+
 * Fixed a bug where :func:`dot <qrisp.dot>` failed with a
   ``TypeError: 'QuantumArrayIterator' object is not iterable``
   (`PR #642 <https://github.com/eclipse-qrisp/Qrisp/pull/642>`_).
--- a/src/qrisp/interface/converter/pytket_converter.py
+++ b/src/qrisp/interface/converter/pytket_converter.py
@@ -174,8 +174,7 @@ def pytket_converter(qc, boxFlag=False):
             tket_ins = OpType.SXdg
 
         elif op.name == "u1":
-            params[0] = params[0] / np.pi
-            # bugged
+            # angle already converted to pi-units in the block above (#631)
             tket_ins = OpType.Rz
         elif op.name == "id":
             params = []
--- a/src/qrisp/interface/converter/pytket_converter.py
+++ b/src/qrisp/interface/converter/pytket_converter.py
@@ -159,10 +159,9 @@ def pytket_converter(qc, boxFlag=False):
                 tket_ins = OpType.CZ
 
         elif op.name == "cp":
-            if hasattr(op, "ctrl_state"):
-                tket_ins = OpType.CRz
-            else:
-                tket_ins = OpType.Rz
+            # cp is the controlled-phase gate diag(1,1,1,e^{i*theta}); pytket's
+            # CU1 is exactly controlled-U1 (CRz is a different gate). (#630)
+            tket_ins = OpType.CU1
 
         elif op.name == "sx":
             # bugged -> params empty
--- a/src/qrisp/interface/converter/pytket_converter.py
+++ b/src/qrisp/interface/converter/pytket_converter.py
@@ -21,8 +21,11 @@ from qrisp import ControlledOperation
 
 
 def create_tket_instruction(op):
-    from pytket import OpType
-    from pytket.circuit import CircBox
+    try:
+        from pytket import OpType
+        from pytket.circuit import CircBox
+    except (ModuleNotFoundError, ImportError) as exc:
+        raise ImportError("PyTket must be installed to be able to use the Qrisp to PyTket converter.") from exc
 
     if op.name == "rxx":
         tket_ins = OpType.XXPhase
@@ -78,8 +81,11 @@ def create_tket_instruction(op):
 
 
 def pytket_converter(qc, boxFlag=False):
-    from pytket import Circuit, OpType, Qubit
-    from pytket.circuit import CircBox, Op, QControlBox
+    try:
+        from pytket import Circuit, OpType, Qubit
+        from pytket.circuit import CircBox, Op, QControlBox
+    except (ModuleNotFoundError, ImportError) as exc:
+        raise ImportError("PyTket must be installed to be able to use the Qrisp to PyTket converter.") from exc
 
     # This dic gives the qiskit qubits/clbits when presented with their identifier
     qubit_dic = {}
--- a/src/qrisp/interface/converter/pytket_converter.py
+++ b/src/qrisp/interface/converter/pytket_converter.py
@@ -15,82 +15,123 @@
 ********************************************************************************
 """
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 import numpy as np
 
 from qrisp import ControlledOperation
 
+if TYPE_CHECKING:
+    from pytket import Circuit, OpType
+    from pytket.circuit import CircBox
 
-def create_tket_instruction(op):
-    try:
-        from pytket import OpType
-        from pytket.circuit import CircBox
-    except (ModuleNotFoundError, ImportError) as exc:
-        raise ImportError("PyTket must be installed to be able to use the Qrisp to PyTket converter.") from exc
+    from qrisp.circuit import Operation, QuantumCircuit
 
-    if op.name == "rxx":
-        tket_ins = OpType.XXPhase
-    elif op.name == "rzz":
-        tket_ins = OpType.ZZPhase
-    elif op.name == "ryy":
-        tket_ins = OpType.YYPhase
-    elif op.name == "measure":
-        tket_ins = OpType.Measure
-    elif op.name == "swap":
-        tket_ins = OpType.SWAP
-    elif op.name == "h":
-        tket_ins = OpType.H
-    elif op.name == "p":
-        tket_ins = OpType.Rz
-    elif op.name == "x":
-        tket_ins = OpType.X
-    elif op.name == "y":
-        tket_ins = OpType.Y
-    elif op.name == "z":
-        tket_ins = OpType.Z
-    elif op.name == "rx":
-        tket_ins = OpType.Rx
-    elif op.name == "ry":
-        tket_ins = OpType.Ry
-    elif op.name == "rz":
-        tket_ins = OpType.Rz
-    elif op.name == "s":
-        tket_ins = OpType.S
-    elif op.name == "s_dg":
-        tket_ins = OpType.Sdg
-    elif op.name == "t":
-        tket_ins = OpType.T
-    elif op.name == "t_dg":
-        tket_ins = OpType.Tdg
-    elif op.name == "u3":
-        tket_ins = OpType.U3
+# Maps a Qrisp operation name to the corresponding pytket ``OpType`` attribute
+# name. Values are strings (resolved via ``getattr`` inside the converter) so this
+# table can live at module level without importing pytket eagerly.
+_GATE_OPTYPES = {
+    "h": "H",
+    "x": "X",
+    "y": "Y",
+    "z": "Z",
+    "s": "S",
+    "s_dg": "Sdg",
+    "t": "T",
+    "t_dg": "Tdg",
+    "sx": "SX",
+    "sx_dg": "SXdg",
+    "id": "noop",
+    "rx": "Rx",
+    "ry": "Ry",
+    "rz": "Rz",
+    "p": "Rz",
+    "u1": "Rz",
+    "u3": "U3",
+    "cx": "CX",
+    "cy": "CY",
+    "cz": "CZ",
+    "cp": "CU1",
+    "rxx": "XXPhase",
+    "rzz": "ZZPhase",
+    "ryy": "YYPhase",
+    "swap": "SWAP",
+    "measure": "Measure",
+}
 
-    elif op.definition:
-        # if complex definition we create an abstract circBox for the section
+
+def create_tket_instruction(op: Operation) -> OpType | CircBox:
+    """Map a single Qrisp operation to its pytket instruction.
+
+    Parameters
+    ----------
+    op : Operation
+        The Qrisp operation to convert.
+
+    Returns
+    -------
+    pytket.OpType or pytket.circuit.CircBox
+        An elementary ``OpType`` for a known gate, or a ``CircBox`` wrapping the
+        operation's definition for a composite gate.
+
+    Raises
+    ------
+    ValueError
+        If the operation is neither a known gate nor decomposable via a
+        ``definition``.
+
+    """
+    from pytket import OpType
+    from pytket.circuit import CircBox
+
+    if op.name in _GATE_OPTYPES:
+        return getattr(OpType, _GATE_OPTYPES[op.name])
+
+    if op.definition:
+        # Composite gate: wrap its definition as an abstract CircBox.
         tket_definition = pytket_converter(op.definition, boxFlag=True)
-        # Defensive check: the boxed definition should span the same number of
-        # qubits as the operation it represents.
         if tket_definition.n_qubits != op.num_qubits:  # pragma: no cover
-            raise Exception("Converted definition of '" + str(op.name) + "' has a mismatched qubit count")
+            raise ValueError("Converted definition of '" + str(op.name) + "' has a mismatched qubit count")
 
-        tket_ins = CircBox(tket_definition)
+        return CircBox(tket_definition)
 
-    else:
-        raise Exception("Could not convert operation " + str(op.name) + " to PyTket")
-
-    return tket_ins
+    raise ValueError("Could not convert operation " + str(op.name) + " to PyTket")
 
 
-def pytket_converter(qc, boxFlag=False):
+def pytket_converter(qc: QuantumCircuit, boxFlag: bool = False) -> Circuit:
+    """Convert a Qrisp QuantumCircuit to a pytket Circuit.
+
+    Parameters
+    ----------
+    qc : QuantumCircuit
+        The Qrisp circuit to convert.
+    boxFlag : bool, optional
+        If True, build the circuit for use as an abstract ``CircBox``: qubits are
+        indexed positionally rather than by named identifier. Used internally when
+        recursing into composite/controlled gate definitions. The default is False.
+
+    Returns
+    -------
+    pytket.Circuit
+        A pytket Circuit equivalent to the input Qrisp circuit.
+
+    Raises
+    ------
+    ImportError
+        If pytket is not installed.
+
+    """
     try:
-        from pytket import Circuit, OpType, Qubit
+        from pytket import Circuit, Qubit
         from pytket.circuit import CircBox
     except (ModuleNotFoundError, ImportError) as exc:
         raise ImportError("PyTket must be installed to be able to use the Qrisp to PyTket converter.") from exc
 
-    # This dic gives the qiskit qubits/clbits when presented with their identifier
+    # This dict gives the pytket qubits/clbits when presented with their identifier.
     qubit_dic = {}
     tket_qc = Circuit()
-    # stringListQubs = []
     tketQubits = []
     for i in range(len(qc.qubits)):
         # add a named qubit
@@ -107,16 +148,17 @@ def pytket_converter(qc, boxFlag=False):
 
     clbit_dic = {}
     # Add Clbits
-    tketClbits = []
     if len(qc.clbits):
         c_reg = tket_qc.add_c_register(name="creg_std", size=len(qc.clbits))
     for i in range(len(qc.clbits)):
         clbit_dic[qc.clbits[i].identifier] = c_reg[i]
-        # this will hopefully be used one day, when other simulators other than Aer are used with this backend, or... quantinuum decides to fix their backend integration
-        # will throw an error on Aer backend and QASM converter, since they apparently only supports a single classical register, which is a lie
-        """ tketClbits.append(Bit( name = str(qc.clbits[i].identifier)))
-        clbit_dic[qc.clbits[i].identifier] = tketClbits[-1]
-        tket_qc.add_bit(tketClbits[-1]) """
+        # NOTE: all clbits share a single register ("creg_std"). An older constraint
+        # (Aer / QASM only accepting one classical register) no longer applies --
+        # verified that multiple registers now run on AerBackend -- so this is just a
+        # simplification. A per-bit alternative (named Bits) is kept for reference:
+        # tketClbits.append(Bit(name=str(qc.clbits[i].identifier)))
+        # clbit_dic[qc.clbits[i].identifier] = tketClbits[-1]
+        # tket_qc.add_bit(tketClbits[-1])
 
     for i in range(len(qc.data)):
         op = qc.data[i].op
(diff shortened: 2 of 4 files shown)


alighazi288:
Hi @purva-thakre. Would love a review when you get the chance. Thanks!

purva-thakre (commented):
Thanks for the PR @alighazi288! 

In case you are interested, would you mind adding more unit tests for the pytket interface? I see the existing unit tests are not enough. No worries, if not. I can create another issue for someone else to pick it up. 

I also think `pytket_rand_test` should be renamed with a `test_` prefix as this is not caught by `pytest` (I think).

alighazi288:
Thanks @purva-thakre! Happy to add broader pytket-interface coverage, but since that's wider than these two mapping fixes, could we track it as the separate issue you mentioned? I'd be glad to pick it up.

On `pytket_rand_test`: You're right that it isn't collected. Heads-up though, it can't just take a `test_` prefix because it uses `AerBackend` from `pytket.extensions.qiskit` with no `importorskip` guard, and CI installs neither `pytket` nor `pytket-qiskit`, so it'd error there instead of run. Since it's pre-existing infra rather than part of these fixes, I think it fits best in that same follow-up issue, happy to handle it there!

purva-thakre:
@alighazi288 Is there a particular reason why you would prefer a new issue and an additional PR for the increased unit test coverage? My previous suggestion was to add an issue to have someone else to pick it up. If you add the unit tests in this PR, we will be able to check the pytket interface is working as expected. Given the lack of proper unit test coverage, I suspect there could be additional pain points beyond the two linked issues. I would prefer to keep all this conversation in one place. 

Yes, I know the CI does not install `pytket` or `pytket-qiskit` right now. We are working to clean up the CI. If you want to, you could add `pytket` to the `pyproject.toml` file temporarily. For the other interface unit tests that were added recently, we use `try except` at the module level imports to keep the interface optional. You might want to add this pattern to `src/qrisp/interface/converter/pytket_converter.py` as well.

alighazi288:
@purva-thakre I mainly suggested a new issue because I didn't want to overstep by unilaterally modifying `pyproject.toml` or expanding the scope of a targeted bug fix. But since you're giving the green light to update the CI config temporarily, your point makes complete sense.

I've added the coverage directly to this PR:

* Added parametrized unitary-equivalence tests over 21 single- and two-qubit gate mappings (`h`, `x`, `y`, `z`, `s`, `s_dg`, `t`, `t_dg`, `sx`, `rx`, `ry`, `rz`, `p`, `u3`, `cx`, `cy`, `cz`, `cp`, `swap`, `rxx`, `rzz`), comparing `to_pytket().get_unitary()` against Qrisp's own `get_unitary()` up to global phase.
* Renamed `pytket_rand_test` to `test_pytket_rand_test` and guarded it with `pytest.importorskip("pytket.extensions.qiskit")` so it is collected by pytest (I only added pytket, not pytket-qiskit, since that test is probabilistic/shot-based and would be flaky). 
* Wrapped the `pytket` import in `pytket_converter.py` in a `try/except`, matching the `cirq_converter.py` pattern. Happy to move it to a module-level guard like `pennylane_converter.py` instead if you prefer that style.
* Added `pytket` to the `dev-testing` extra in `pyproject.toml` so the new tests run in CI, and rebased onto the latest `main`.

On your hunch about additional pain points, I probed `u3`, `sx`, `s_dg`, `t_dg`, `mcx`, and `mcrx` and they all convert correctly in isolation, so nothing new surfaced beyond the #630/#631 fixes! (The controlled-gate ordering issue looks context-dependent, which maps to the separate #632 issue).

Hopefully this resolves all the comments, please let me know if you'd like me to change/add anything else!

tests/interface_tests/test_tket_converter.py
purva-thakre: thanks for adding the extra unit tests! I will do a deeper review over the next couple of days. If you are interested, you could also make sure this unit test uses `pytest` functionality properly to test the function behaves as expected. We do not need any of these `print` statements.

tests/interface_tests/test_tket_converter.py
```diff
@@ -124,3 +131,119 @@ def pytket_rand_test():
...
+    pytest.importorskip("pytket")
+
+    theta = 0.7
+    qc = QuantumCircuit(2)
+    qc.append(CPGate(theta), [0, 1])
+
+    unitary = qc.to_pytket().get_unitary()
+    expected = np.diag([1, 1, 1, np.exp(1j * theta)])
+
+    assert _matches_up_to_global_phase(unitary, expected)
```
purva-thakre: if you want to, these two could also be combined via parametrize.

tests/interface_tests/test_tket_converter.py
```diff
@@ -124,3 +131,119 @@ def pytket_rand_test():
...
+_SINGLE_QUBIT_GATE_BUILDERS = {
+    "h": lambda qc: qc.h(0),
+    "x": lambda qc: qc.x(0),
+    "y": lambda qc: qc.y(0),
+    "z": lambda qc: qc.z(0),
+    "s": lambda qc: qc.s(0),
+    "s_dg": lambda qc: qc.s_dg(0),
+    "t": lambda qc: qc.t(0),
+    "t_dg": lambda qc: qc.t_dg(0),
+    "sx": lambda qc: qc.append(SXGate(), [0]),
```
purva-thakre: we also have a sx_dg

tests/interface_tests/test_tket_converter.py
```diff
@@ -124,3 +131,119 @@ def pytket_rand_test():
...
+
+    assert _matches_up_to_global_phase(unitary, expected)
+
+
+_GATE_THETA = 0.7
+
+# Gate builders whose pytket conversion must reproduce Qrisp's own unitary.
+# Comparing to_pytket().get_unitary() against QuantumCircuit.get_unitary() checks
+# that the converter preserves each gate's action (up to a global phase).
+_SINGLE_QUBIT_GATE_BUILDERS = {
```
purva-thakre: I think this is also missing an identity gate

purva-thakre: missing a unit test for `measure` as well.

purva-thakre:
> On your hunch about additional pain points, I probed u3, sx, s_dg, t_dg, mcx, and mcrx and they all convert correctly in isolation, so nothing new surfaced beyond the https://github.com/eclipse-qrisp/Qrisp/issues/630/https://github.com/eclipse-qrisp/Qrisp/issues/631 fixes!

Ah, but the pain points are not just limited to the converted circuit. I noticed the converter is referring to an `ryy` gate when a gate by that name does not even exist in qrisp :scream:

I will add an issue for this soon given that the `ryy` gate is also referenced in other converters as well. :dizzy_face:

alighazi288:
Hi @purva-thakre, thanks for the review! Pushed an update addressing the comments. 

The CI failure (u3) was in the test helper, since the `u3` conversion itself is correct. `_matches_up_to_global_phase` was normalizing each unitary against a single largest-magnitude pivot entry. For `u3` the two diagonal entries tie in magnitude (cos(θ/2) on both corners), so tiny floating-point differences between the pytket and Qrisp matrices made argmax pick different entries on the CI runner than locally, referencing the two matrices to different global phases and failing the comparison. I switched the helper to recover the phase from the Frobenius inner product `(np.vdot)` over the whole matrix, which is robust to those magnitude ties (and is a bit simpler than the pivot version).

Please let me know how this looks or if you want me to revert anything!

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -159,10 +165,9 @@ def pytket_converter(qc, boxFlag=False):
                 tket_ins = OpType.CZ
 
         elif op.name == "cp":
-            if hasattr(op, "ctrl_state"):
-                tket_ins = OpType.CRz
-            else:
-                tket_ins = OpType.Rz
+            # cp is the controlled-phase gate diag(1,1,1,e^{i*theta}); pytket's
+            # CU1 is exactly controlled-U1 (CRz is a different gate). (#630)
```
purva-thakre:
```suggestion
            # CU1 is exactly controlled-U1 (CRz is a different gate).
```

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -174,8 +179,7 @@ def pytket_converter(qc, boxFlag=False):
             tket_ins = OpType.SXdg
 
         elif op.name == "u1":
-            params[0] = params[0] / np.pi
-            # bugged
+            # angle already converted to pi-units in the block above (#631)
```
purva-thakre:
```suggestion
            # angle already converted to pi-units in the block above
```

tests/interface_tests/test_tket_converter.py
```diff
@@ -100,27 +107,159 @@ def pytket_rand_test():
...
+    return np.allclose(actual, expected * phase, atol=atol)
+
+
+_GATE_THETA = 0.7
+
+# Regression cases for the two mapping fixes. Each converted unitary must equal
+# the gate's textbook matrix up to a global phase, independently of Qrisp's own
+# get_unitary().
+_REGRESSION_CASES = {
+    # #631: the u1 angle must not be divided by pi twice. u1(theta) maps to a
```
purva-thakre:
```suggestion
    # the u1 angle must not be divided by pi twice. u1(theta) maps to a
```

tests/interface_tests/test_tket_converter.py
```diff
@@ -100,27 +107,159 @@ def pytket_rand_test():
...
+    # pytket Rz, which differs only by a global phase, so the converted unitary
+    # must equal diag(1, e^{i*theta}).
+    "u1": (
+        1,
+        lambda qc: qc.append(U1Gate(_GATE_THETA), [0]),
+        np.diag([1, np.exp(1j * _GATE_THETA)]),
+    ),
+    # #630: cp must map to controlled-phase, not CRz. cp(theta) is
```
purva-thakre:
```suggestion
    # cp must map to controlled-phase, not CRz. cp(theta) is
```

purva-thakre:
@alighazi288 Sorry to request this again, could you please rebase this PR?

alighazi288:
@purva-thakre No worries at all! Done!

purva-thakre:
Thanks for the rebase! Looking at the pytest-coverage report, the pytket converter still needs additional unit tests. Since we want to make sure the pytket converter is properly covered by unit tests in this PR, it should be worth looking into. The lines it is flagging as uncovered are: `27-28, 35, 67-78, 87-88, 103-106, 146, 153, 159, 165, 190-219, 225, 228`. 

https://github.com/eclipse-qrisp/Qrisp/actions/runs/29135295756/job/86498320652?pr=662#step:5:327

The pytest-coverage is a bit messy because there is a large portion of the qrisp functionality that lacks unit tests. You need to scroll and look at the `interface` module in the report. 

Screenshot from 2026-07-14 17-04-34

purva-thakre:
There is also a more complicated test in the Cirq converter unit tests that should be added for the pytket converter as well (Grover's). Right now, your unit tests are only testing a simple circuit.

alighazi288:
Thanks @purva-thakre! Pushed an update addressing both comments: full converter coverage and a Grover test, plus a couple of things I hit along the way.

**Test coverage.** The converter is now exercised end to end. New tests: a Grover search (`test_to_pytket_grover`), multi-controlled gates (`test_to_pytket_controlled_gate_unitary`, MCX + MCRX), a composite/non-elementary gate, an exact global-phase check, an unsupported-gate error, and two missing-pytket `ImportError` paths. With these, every reachable line in `pytket_converter.py` is covered — the only remaining uncovered line is the `ryy` mapping, which is the phantom gate you're tracking in the separate cross-converter issue, so I left it for that.

**Grover test.** I used a small instance (`QuantumFloat(2, -1)` → 9 qubits) rather than porting the Cirq test's larger one, because pytket's built-in statevector simulator caps out at 11 qubits. It still exercises the full machinery: composite/controlled gates as `CircBox`es, qubit (de)allocation, and the diffuser's global phase.

**Global-phase gap (found via the Grover test).** `to_pytket` was raising `"Could not convert operation gphase to PyTket"`, so *any* circuit built with the diffuser failed to convert. The Cirq and Qiskit converters already handle `gphase`, so I added the equivalent mapping to pytket's circuit-wide `add_phase`. That's what unblocked the Grover test.

**Dead-code removal.** To get those lines out of the coverage report I removed some unreachable code: the `if 0 == 1:` relic block, the identical `hasattr(ctrl_state)` if/else branches for cx/cy/cz (both sides assigned the same `OpType`), and the `QControlBox` path (dead because `pytket_converter` always returns a `Circuit`). No behavioural change. **If any of that was intentionally parked for future work, let me know and I'll restore it (or gate it behind a `# pragma: no cover`) instead.**

**Changelog.** Added a Bug Fixes entry, which also gets the `changelog` check passing.

Happy to adjust anything!

documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -15,6 +15,11 @@ Other New Features
 Bug Fixes
 ---------
 
+* Fixed the ``to_pytket`` converter: ``u1`` angles are no longer divided by
+  ``pi`` twice, ``cp`` now maps to controlled-phase (``CU1``) instead of
+  ``CRz``, and global-phase (``gphase``) gates are now supported
+  (`PR #662 <https://github.com/eclipse-qrisp/Qrisp/pull/662>`_).
```
purva-thakre, line 67: you missed mentioning the increased coverage and that the pytket converter is refactored :)

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -21,8 +21,11 @@
 
 
 def create_tket_instruction(op):
```
purva-thakre: add docstrings and type hints. going forward, we should make things easier for future maintainers.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -78,8 +81,11 @@ def create_tket_instruction(op):
 
 
 def pytket_converter(qc, boxFlag=False):
```
purva-thakre: add docstring and type hints here as well. take a look at the docstrings in other modules if you want an idea of how the docstring of a public facing function should be structured.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -78,8 +81,11 @@ def create_tket_instruction(op):
...
-    from pytket import Circuit, OpType, Qubit
-    from pytket.circuit import CircBox, Op, QControlBox
+    try:
+        from pytket import Circuit, OpType, Qubit
+        from pytket.circuit import CircBox
+    except (ModuleNotFoundError, ImportError) as exc:
+        raise ImportError("PyTket must be installed to be able to use the Qrisp to PyTket converter.") from exc
```
purva-thakre, line 222: I think this should be in the main function only. The same lines from the helper function could be removed.

src/qrisp/interface/converter/pytket_converter.py
purva-thakre: I know you didn't make this change. Even then, this variable is defined and not used by the function.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -64,10 +67,10 @@ def create_tket_instruction(op):
...
         # if complex definition we create an abstract circBox for the section
         tket_definition = pytket_converter(op.definition, boxFlag=True)
-        # might need adjustment
-        # if len(op.definition.clbits):
-        if tket_definition.n_qubits != op.num_qubits:
-            raise Exception
+        # Defensive check: the boxed definition should span the same number of
+        # qubits as the operation it represents.
+        if tket_definition.n_qubits != op.num_qubits:  # pragma: no cover
+            raise Exception("Converted definition of '" + str(op.name) + "' has a mismatched qubit count")
```
purva-thakre: Should be `ValueError`

src/qrisp/interface/converter/pytket_converter.py
purva-thakre: this should also be `ValueError`

purva-thakre: this appears to be a very outdated comment. would you be interested in verifying if this is still the issue?

I also think instead of the string, the lines 117-119 should be commented out instead.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -139,30 +145,25 @@ def pytket_converter(qc, boxFlag=False):
...
         elif op.name == "cp":
-            if hasattr(op, "ctrl_state"):
-                tket_ins = OpType.CRz
-            else:
-                tket_ins = OpType.Rz
+            # cp is the controlled-phase gate diag(1,1,1,e^{i*theta}); pytket's
+            # CU1 is exactly controlled-U1 (CRz is a different gate). (#630)
+            tket_ins = OpType.CU1
 
         elif op.name == "sx":
```
purva-thakre: maybe it's just me but the way these special cases are defined makes no sense. There is already a helper function defined at the beginning of the file to handle the code conversions. All of these `if elif` lines for ops could be moved to the `create_tket_instruction` function. Even this helper function could be replaced with a mapping dictionary such that the function is not full of many `if elif elif...` chains. 

I know you did not write the original code. Since we are taking a look at the dead code blocks, increasing test coverage etc, it makes sense to also refactor the code. Let me know if I am asking too much of you. I am also fine with merging this as is with the minor corrections and then letting someone else refactor it completely.

tests/interface_tests/test_tket_converter.py
```diff
@@ -20,13 +25,22 @@
...
     ZGate,
 )
+from qrisp.interface.converter.pytket_converter import (
+    create_tket_instruction,
+    pytket_converter,
+)
 
 
-def pytket_rand_test():
+def test_pytket_rand_test():
```
purva-thakre: I have not taken a deeper look at the unit tests. I let an LLM review it. 

----

**Kill test_pytket_rand_test or rewrite it as a deterministic integration test**
It has the logic bug (op in c_gates or mc_gates is always true), uses randomness (flaky), mixes instances/classes/strings in the same list, and the ±40% shot-noise tolerance adds little value. The separate per-gate unitary tests already cover every gate deterministically. The Grover test covers end-to-end integration. This test should either be deleted or rewritten as a deterministic multi-gate circuit that checks unitary equivalence.

**Standardize gate builder style**
sx uses qc.append(SXGate(), [0]) but sx_dg uses qc.sx_dg(0). Pick one pattern. Also u3 uses qc.append(U3Gate(...), [0]) — all single-qubit gates should ideally use the same qc.<name>(...) method if one exists.

**Align the dict shapes for gate builders**
_CONTROLLED_GATE_BUILDERS stores (num_qubits, build) tuples while _SINGLE_QUBIT_GATE_BUILDERS stores just the builder. The parametrized tests are forced to handle both shapes differently (line 229 vs 312-315). Standardize: either store (num_qubits, build) everywhere, or infer num_qubits from the circuit constructor call.

tests/interface_tests/test_tket_converter.py
```diff
@@ -100,27 +114,300 @@ def pytket_rand_test():
...
+
+def test_to_pytket_gphase():
+    """to_pytket() preserves a global-phase gate exactly.
+
+    ``x, x`` is the identity, so ``gphase(theta)`` leaves the unitary as
+    ``e^{i*theta} * I``. Global phase is unobservable up to a phase factor, so
+    this compares the converted unitary against Qrisp's own get_unitary()
+    *exactly* to prove the phase itself is carried across.
+    """
+    pytest.importorskip("pytket")
```
purva-thakre: adding this line for all unit tests makes it redundant. 

move this to the module level.

tests/interface_tests/test_tket_converter.py
```diff
@@ -100,27 +114,300 @@ def pytket_rand_test():
...
+    with _make_import_fail():
+        with pytest.raises(ImportError, match="PyTket must be installed"):
+            pytket_converter(QuantumCircuit(1))
+
+
+def test_create_tket_instruction_import_error():
+    """create_tket_instruction() raises a clear ImportError when pytket is missing."""
+    with _make_import_fail():
+        with pytest.raises(ImportError, match="PyTket must be installed"):
+            create_tket_instruction(XGate())
```
purva-thakre: sorry, I do not understand what the purpose of defining each these unit tests is? Are they not better contained as one parametrized unit test?

tests/interface_tests/test_tket_converter.py
```diff
@@ -100,27 +114,300 @@ def pytket_rand_test():
...
+    ),
+    # #630: cp must map to controlled-phase, not CRz. cp(theta) is
+    # diag(1, 1, 1, e^{i*theta}); CRz splits the phase across the two target
+    # states and is a different gate.
+    "cp": (
+        2,
+        lambda qc: qc.append(CPGate(_GATE_THETA), [0, 1]),
+        np.diag([1, 1, 1, np.exp(1j * _GATE_THETA)]),
+    ),
+}
```
purva-thakre: I think it is fine if we move these gates to their specific unit tests and get rid of the `regression_cases` ones. `u1` could be moved to the single qubit ones and `cp` to the two qubit ones.

purva-thakre (changes requested):
Thanks! Very close to merging this. 

I have made some suggestions because with the newer changes the code was becoming spaghetti code. Let me know if you are up for it. :)

alighazi288:
Thanks for the review @purva-thakre! Everything I addressed:

**Refactor.** The gate cases now live in one module-level `{name: OpType}` lookup resolved in `create_tket_instruction`; `pytket_converter`'s loop only handles the special cases (π-unit params, `gphase` → `add_phase`, qubit (de)allocation, controlled ops → `CircBox`). Hopefully much less spaghetti.

**Docstrings + type hints** added to both functions (numpydoc, with `from __future__ import annotations` so the pytket return types aren't imported eagerly).

**Smaller fixes:** import guard is now only in `pytket_converter`; both raises are `ValueError`; removed the unused `tketClbits`; converted the disabled Bit-based block from a string literal to comments. On your question there, I installed pytket-qiskit and verified the single-classical-register claim is outdated because a circuit with multiple classical registers is valid and runs on the current `AerBackend`. I updated the comment to reflect that (and left the single register as a harmless simplification).

**Tests:** deleted `test_pytket_rand_test`; consolidated the per-gate tests into one parametrized `test_to_pytket_gate_unitary` with a uniform `(num_qubits, builder)` shape (folded in `u1`, dropped `_REGRESSION_CASES`); standardized builders to `qc.<gate>()` methods; moved `importorskip` to module level; collapsed the import-error tests to one.

While consolidating, I found that a strict *unitary* comparison for `mcx`/`mcrx` is flaky. Qrisp's multi-controlled synthesis interacts with the context-dependent controlled-gate ordering from #632, so the converted gate can come out bit-reversed depending on what else ran in the session (it fails when the tket tests run alongside the cirq/qiskit ones). I switched `mcx`/`mcrx` to a **structural** conversion check (asserts it converts to a CircBox on the right qubits) so it's robust; the Grover test still exercises multi-controlled gates end-to-end, and exact multi-controlled equivalence is left to #632. Happy to revisit if you'd prefer a different approach.

purva-thakre:
@alighazi288 Thank you. 

While I appreciate your efforts so far, I just want to ask if you have verified the changes on your own? Your last comment comes off as an LLM feedback. I am not against using LLMs as long as a human contributor understands the changes.

alighazi288:
@purva-thakre Totally fair to ask. I use an LLM as a bit of a ghostwriter for my GitHub comments because I'm really bad at formatting Markdown and don't want to subject you to my brain dumps! That last one definitely came out sounding quite robotic since I just copy pasted. I do take full responsibility for my code and verify all the changes myself locally before pushing!

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -15,76 +15,123 @@
...
         tket_definition = pytket_converter(op.definition, boxFlag=True)
-        # might need adjustment
-        # if len(op.definition.clbits):
-        if tket_definition.n_qubits != op.num_qubits:
-            raise Exception
+        if tket_definition.n_qubits != op.num_qubits:  # pragma: no cover
+            raise ValueError("Converted definition of '" + str(op.name) + "' has a mismatched qubit count")
 
-        tket_ins = CircBox(tket_definition)
+        return CircBox(tket_definition)
```
purva-thakre, line 110: I think this is a duplicate of what is in the main function

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -120,113 +168,39 @@ def pytket_converter(qc, boxFlag=False):
...
-                    tket_definition.name = base_name
-                    tket_ins = CircBox(tket_definition)
-
-                else:
-                    # else a simpler multi controlled gate can be created
-                    tket_ins = QControlBox(Op.create(base_gate), len(op.controls))
-
+            tket_definition = pytket_converter(op.definition, boxFlag=True)
+            tket_definition.name = base_name
+            tket_ins = CircBox(tket_definition)
```
purva-thakre: this is a duplicate of what is in the `create_tket_instructions` function. I have added my comment to the relevant lines. Either keep it here or in the helper.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -120,113 +168,39 @@ def pytket_converter(qc, boxFlag=False):
...
+        if op.name in ["sx", "sx_dg", "id"]:
+            # Qrisp attaches spurious params to these; pytket takes none.
             params = []
-            # bugged
-            tket_ins = OpType.noop
 
-        elif issubclass(op.__class__, ControlledOperation):
+        if op.name not in _GATE_OPTYPES and issubclass(op.__class__, ControlledOperation):
+            # A composite controlled operation (e.g. mcx) is emitted as an abstract
+            # CircBox of its definition.
```
purva-thakre: I do not think this takes `ctrl_state` into consideration. By default, all the controlled gates are controlled on value `111....` in this converter.

src/qrisp/interface/converter/pytket_converter.py
```diff
@@ -15,76 +15,123 @@
...
-    # This dic gives the qiskit qubits/clbits when presented with their identifier
+    Returns
+    -------
+    pytket.Circuit
+        A pytket Circuit equivalent to the input Qrisp circuit.
+
+    Raises
+    ------
+    ImportError
+        If pytket is not installed.
```
purva-thakre, line 213: there are a couple of gates that are not supported which leads to the converter raising an error. they should be mentioned here

purva-thakre:
I added some smoke tests with the help of an LLM in https://github.com/eclipse-qrisp/Qrisp/pull/662/commits/22636d79f56d9be5c200c8d5600025a9448aebf2

Hopefully, the comments should be detailed enough to point to the failures in the pytket converter. Those that have documented issues, mark them as xfail when you are done. 

One of the added failures could be due to #786 

Edit: Either way, I can wrap it up after you make your changes.

PR #744, opened 2026-07-16
Backend sampling

positr0nium:
## Description

Introduces `@backend_sampler` — a Jasp decorator that routes `sample()` and `expectation_value()` calls through a real quantum backend instead of the Jaspify simulator. The quantum circuit is extracted once, executed on the backend for all shots, and the classical post-processing is replayed via the Jaspr's own while-loop — JIT-compiled through JAX. Control-flow primitives (`fori_loop`, `while`, `cond`, `scan`, `jit`) are handled via a custom equation evaluator that propagates through all nesting levels. Example:

```python
from qrisp import QuantumFloat, h, measure
from qrisp.jasp import sample, backend_sampler
from qrisp.interface import StimBackend

backend = StimBackend()

@backend_sampler(backend=backend)
def main(k):
    def kernel(k):
        qf = QuantumFloat(4)
        h(qf[0])
        return measure(qf)
    return sample(kernel, shots=100)(k)

result = main(1)
# result is a JAX array of shape (100,) with backend results
```

## Related Issues

Closes #
Related to the `sampling_upgrades` branch

## Type of Change

- [x] Documentation
- [ ] Performance improvement
- [ ] CI / Build
- [ ] Bug Fix

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **New decorator: `@backend_sampler(backend=...)`** — routes `sample()` / `expectation_value()` calls through a real backend. Uses `make_jaxpr` (not `make_jaspr`) for the outer trace — no quantum tracing context, so direct quantum operations outside sampling kernels fail cleanly at trace time.

- **New interpreter module: `backend_sampling_interpreter.py`** — contains the Jaspr-level logic:
  - `_extract_to_qc_args()` — mini-interpreter that runs the eval Jaspr until the first `sampling_body_func` call, capturing all arguments automatically (handles JAX-implicitly-prepended closure variables).
  - `_make_backend_sampling_fn()` — factory returning a `backend_sampling_fn` that extracts the circuit via `Jaspr.to_qc()`, runs it once on the backend, and replays the post-processing loop.
  - `_body_loop_evaluator()` — intercepts `sampling_body_func` in the eval Jaspr, replacing it with `Jaspr.extract_post_processing()` using pre-computed measurement bits.

- **Control-flow propagation** — the outer equation evaluator handles `jit`/`pjit`, `while`, `cond`, and `scan` primitives by recursively calling `eval_jaxpr` with the custom evaluator, ensuring `sample()`/`expectation_value()` calls are intercepted regardless of nesting depth inside control flow.

- **Real-time feedback detection** — catches the `"Tried to convert real-time feedback into QuantumCircuit"` exception from `to_qc()` and re-raises as `RuntimeError` with a clear message pointing to `jaspify`.

- **`_backend_shots_marker`** — identity `@jax.jit` marker inserted at the top of `sampling_eval_function` and `expectation_value_eval_function` so the backend sampler can reliably locate the shot count in the traced Jaxpr without fragile position-based extraction.

- **Supporting changes:**
  - `sampling.py` / `ev.py`: added `_backend_shots_marker` call; renamed `state_prep` → `sampling_kernel` in `expectation_value` for consistency.
  - `measurement_primitive.py`: fixed `measure_implementation` to index into the qubit array.
  - `tracing_quantum_session.py`: allow qubit reuse from templates.
  - `qaoa_problem.py`: accept `QuantumVariableTemplate` in `qarg` parameter.
  - `qc_extraction_interpreter.py`: support `QuantumVariableTemplate` and parity handle fixes.

- **Documentation:**
  - Enriched `backend_sampler` docstring with usage examples, parameter docs, Raises section, and architecture notes.
  - Added `BackendSampling.rst` to Sphinx docs and updated `Jasp/index.rst` toctree.

- **Tests (57 total):**
  - 51 tests covering sampling patterns, expectation values, corner cases, error handling.
  - 6 control-flow propagation tests: `fori_loop`, `cond`, `while_loop`, `scan`, nested `jit`, `switch`.

## How was it tested?

All 57 tests pass locally:

```bash
pytest tests/jax_tests/test_backend_sampling.py \
       tests/jax_tests/test_backend_sampling_ev.py \
       tests/jax_tests/test_to_qc_in_tracing_context.py -q
# 57 passed in ~25s
```

## Checklist
- [x] All tests pass locally

## Reviewer Notes

- The split between `backend_sampling.py` (decorator + outer evaluator) and `backend_sampling_interpreter.py` (Jaspr interpreters) keeps the architecture modular.
- The `make_jaxpr` vs `make_jaspr` choice is deliberate: `make_jaxpr` avoids opening a quantum tracing context for the outer orchestration function, so any accidental direct quantum operations fail at trace time with a clear error.
- The control-flow handlers (`jit`/`while`/`cond`/`scan`) follow the same pattern as the post-processing interpreter — propagating the custom evaluator downward via `eval_jaxpr(eqn_evaluator=...)`.


src/qrisp/algorithms/qaoa/qaoa_problem.py
```diff
@@ -588,6 +590,10 @@ def run(
             def qarg_prep():
                 return QuantumArray(qtype=template.construct(), shape=shape)
 
+        elif isinstance(qarg, QuantumVariableTemplate):
```
renezander90, line 596: Should the `run` method of `VQE` be modified similarly?

src/qrisp/jasp/evaluation_tools/backend_sampling.py
```diff
@@ -0,0 +1,454 @@
...
+
+        a, b = main()
+        # Each call is independently routed through the backend.
+
+    """
+    return lambda func: _make_backend_sampler_wrapper(func, backend)
+
+
+# ===========================================================================
+# Control-flow handlers — propagate the custom evaluator downwards
```
renezander90, line 337: Can't we use the shared helpers from `traced_control_flow_interpretation.py`?

src/qrisp/jasp/evaluation_tools/backend_sampling.py
```diff
@@ -0,0 +1,454 @@
...
+        All **orchestration logic** (the code
+        in the decorated function that calls :func:`~qrisp.jasp.sample`
+        and :func:`~qrisp.jasp.expectation_value`, passes arguments,
+        and combines results) is traced into a Jaspr and compiled via
+        :func:`jax.jit`.  This means the non-coherence wrapping logic
+        runs at JAX speed, even when orchestrating many sampling calls.
+
+    Parameters
+    ----------
+    backend : :ref:`BackendInterface`
```
renezander90, line 228:
```suggestion
    backend : BackendLike
```

renezander90: See `qrisp.interface.backend.py`. Should `BackendLike` be mentioned under `Typing` in the documentation @PietropaoloFrisoni ?

src/qrisp/jasp/evaluation_tools/backend_sampling.py
```diff
@@ -0,0 +1,454 @@
...
+            def kernel(k):
+                qf = QuantumFloat(4)
+                h(qf[0])
+                return measure(qf)
+            return sample(kernel, shots=100)(k)
+
+        result = main(1)
+        # result is a JAX array of shape (100,) with backend results
+
+    Using a custom backend:
```
renezander90: This uses the same ` QrispSimulatorBackend()` as the example above. Use different backend (e.g., `AerSimulator`) or remove.

src/qrisp/jasp/evaluation_tools/backend_sampling.py
PietropaoloFrisoni: Yes, that would be ideal!

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,453 @@
...
+                    m:QuantumState = jasp.consume_quantum_kernel k
+                  in (l, m) }
+                ...
+              ] ...
+            in (f,) }
+        ] ...
+      in (b,) }
+
+The three interception points ①②③ correspond to the architecture
+described above.  Interception ② lives in the evaluation-tools module;
```
renezander90: And where is interception 1?

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,453 @@
...
+    │  └────────────────────────────────────────────────────────┘ │
+    │  result = acc (or acc / shots for EV)                      │
+    └─────────────────────────────────────────────────────────────┘
+
+Key design properties
+---------------------
+
+* **Single backend call per invocation** — the circuit is built once,
+  the backend runs once, and only lightweight post-processing varies
+  per shot.
```
renezander90:
```suggestion
  the backend runs once with the specified number of shots,
  and only lightweight post-processing is applied per shot.
```

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,453 @@
...
+    insert_outvalues,
+)
+from qrisp.jasp.jasp_expression.centerclass import Jaspr
+
+# ===========================================================================
+# Jaspr traversal utility
+# ===========================================================================
+
+
+def find_named_jaxpr(jaxpr, target_name):
```
renezander90: Mark is internal: `_find_named_jaxpr`

src/qrisp/jasp/evaluation_tools/backend_sampling.py
positr0nium: ``BackendLike`` would imply that this can also be used with batched backends, or? This doesn't work with the overall execution model. The results need to be available immediately to continue with the post-processing.

src/qrisp/jasp/evaluation_tools/backend_sampling.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/d0eaeebbf855baa787c03f87657b35d0b4465433

src/qrisp/jasp/evaluation_tools/backend_sampling.py
```diff
@@ -0,0 +1,374 @@
...
+    evaluate_cond_under_trace,
+    evaluate_scan_under_trace,
+    evaluate_while_loop_under_trace,
+)
+
+__all__ = ["backend_sampler", "find_named_jaxpr"]
+
+
+# ===========================================================================
+# Eqn evaluator that intercepts eval functions with pure_callback
```
renezander90: Some remarks from AI on this file:

**Findings**

1. **High: loop index detection is fragile.**  
   backend_sampling_interpreter.py identifies `acc_pos` by the first matching aval, then backend_sampling_interpreter.py identifies `i_pos` as the first preceding `int64[]`. Captured closure values are allowed before `i` and `acc`, so an equal-shaped accumulator or scalar integer can be selected instead. This can index the wrong backend shot or raise unexpectedly. Carry positions should be derived from the surrounding `while` equation or validated structurally, not inferred solely from avals.

2. **High: backend execution is placed inside `jax.pure_callback`.**  
   backend_sampling.py invokes `backend.run(...)` and `np.random.shuffle(...)` through `pure_callback`. That callback is not pure: it submits external jobs and consumes randomness. Under transformations or compiler decisions that duplicate/replay callbacks, this can submit extra backend executions or produce inconsistent results. An effectful callback boundary such as `io_callback`, or an explicit restriction against transformations that may replay it, is needed.

3. **Medium: `return_dict=True` is not intercepted.**  
   ev.py renames the evaluator to `dict_sampling_eval_function`, but backend_sampling.py only intercepts `sampling_eval_function` and `expectation_value_eval_function`. If `return_dict` is intended to be supported with `backend_sampler`, the call bypasses this interpreter and may reach unsupported quantum primitives.

**Test gaps**

The existing tests cover common sampling and control-flow cases, but not captured-value aval collisions, `return_dict=True`, effectful callback replay.

**Counter example (1.)**

```python
@backend_sampler(backend=QrispSimulatorBackend())
def main(k):
    captured = jnp.zeros((4,), dtype=jnp.float64) + k

    def kernel():
        q = QuantumFloat(2)
        h(q[0])
        return measure(q) + captured[0]

    return sample(kernel, shots=4)()
```

Result:

```text
RuntimeError: Could not identify loop-index position in sampling_body_func invars
```

The traced body inputs were:

```text
0: captured array  float64[4]
1: i               int64[]
2: acc             float64[4]
3: QuantumState
```

Since the accumulator output is `float64[4]`, backend_sampling_interpreter.py incorrectly selects input `0` as `acc_pos`. It then searches for `i` only before position `0`, finds nothing, and raises at backend_sampling_interpreter.py.

As a control, changing the captured array to shape `(3,)` succeeds and returns shape `(4,)`. So finding 1 is a valid, reproducible bug specifically for dynamic captured values whose aval matches the accumulator.

**Recommended Design (1.)**

The bug is caused by inferring `i` and `acc` from avals in backend_sampling_interpreter.py. The generated `fori_loop` already gives us a structural answer:

```python
sampling_body_func(i, args)
```

After flattening, the enclosing `while` body’s first loop input is always `i`, and the second is `acc`. Captured values may appear as additional call inputs, but they do not change the loop-carry order.

Therefore:

1. Find the `while` equation containing `sampling_body_func`.
2. Get its `body_jaxpr`.
3. Take `body_jaxpr.invars[0]` as the loop-index SSA variable.
4. Find that same variable in the `sampling_body_func` call’s `invars`.
5. Use that call-input position as `i_pos`.
6. Remove `acc_pos` inference entirely. It is not actually used by `_body_loop_evaluator`.

The mapping should be based on SSA variable identity, not `shape` and `dtype`.

**Patch Outline**

Add a helper that returns the sampling call and its loop-index position:

```python
def find_sampling_body_context(jaxpr):
    for eqn in jaxpr.eqns:
        if eqn.primitive.name == "while":
            body_jaxpr = eqn.params["body_jaxpr"]

            for body_eqn in body_jaxpr.eqns:
                if (
                    body_eqn.primitive.name in ("jit", "pjit")
                    and body_eqn.params.get("name") == "sampling_body_func"
                ):
                    index_var = body_jaxpr.invars[0]

                    for index, var in enumerate(body_eqn.invars):
                        if var is index_var:
                            return (
                                body_eqn.params.get("jaxpr")
                                or body_eqn.params.get("call_jaxpr"),
                                index,
                            )

                    raise RuntimeError(
                        "Could not map sampling loop index to sampling_body_func"
                    )

            # Recurse into nested control-flow bodies as needed.

    return None
```

Then replace the aval-based block at backend_sampling_interpreter.py with:

```python
body_jaxpr, i_pos = find_sampling_body_context(inner_jaxpr.jaxpr)
if body_jaxpr is None:
    raise RuntimeError("sampling_body_func not found inside eval function Jaxpr")

body_jaspr = Jaspr(body_jaxpr)
```

The evaluator can keep using:

```python
iteration = invalues[i_pos]
results = post_proc(meas_results[iteration], *invalues[:-1])
```

`extract_post_processing()` already expects all non-QuantumState arguments in Jaxpr order, so captured values should continue to be passed through unchanged.

**Regression Test**

Add a test to test_backend_sampling.py:

```python
def test_captured_array_matching_accumulator_shape():
    @backend_sampler(backend=_get_backend())
    def main(k):
        captured = jnp.zeros((4,), dtype=jnp.float64) + k

        def kernel():
            qf = QuantumFloat(2)
            h(qf[0])
            return measure(qf) + captured[0]

        return sample(kernel, shots=4)()

    result = main(2.0)

    assert result.shape == (4,)
    assert {float(value) for value in result} <= {2.0, 3.0}
```

This currently raises:

```text
RuntimeError: Could not identify loop-index position in sampling_body_func invars
```

With structural SSA mapping, it should return four samples successfully. No files were modified.

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,434 @@
...
+# * This Source Code may also be made available under the following Secondary
+# * Licenses when the conditions for such availability set forth in the Eclipse
+# * Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+# * with the GNU Classpath Exception which is
+# * available at https://www.gnu.org/software/classpath/license.html.
+# *
+# * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+# ********************************************************************************
+
+r"""Backend-sampling Jaspr interpreter.
```
renezander90, line 17: Some remarks from AI on this file:

**Findings**

1. **High: loop index detection is fragile.**  
   backend_sampling_interpreter.py identifies `acc_pos` by the first matching aval, then backend_sampling_interpreter.py identifies `i_pos` as the first preceding `int64[]`. Captured closure values are allowed before `i` and `acc`, so an equal-shaped accumulator or scalar integer can be selected instead. This can index the wrong backend shot or raise unexpectedly. Carry positions should be derived from the surrounding `while` equation or validated structurally, not inferred solely from avals.

2. **High: backend execution is placed inside `jax.pure_callback`.**  
   backend_sampling.py invokes `backend.run(...)` and `np.random.shuffle(...)` through `pure_callback`. That callback is not pure: it submits external jobs and consumes randomness. Under transformations or compiler decisions that duplicate/replay callbacks, this can submit extra backend executions or produce inconsistent results. An effectful callback boundary such as `io_callback`, or an explicit restriction against transformations that may replay it, is needed.

3. **Medium: `return_dict=True` is not intercepted.**  
   ev.py renames the evaluator to `dict_sampling_eval_function`, but backend_sampling.py only intercepts `sampling_eval_function` and `expectation_value_eval_function`. If `return_dict` is intended to be supported with `backend_sampler`, the call bypasses this interpreter and may reach unsupported quantum primitives.

**Test gaps**

The existing tests cover common sampling and control-flow cases, but not captured-value aval collisions, `return_dict=True`, effectful callback replay.

**Counter example (1.)**

```python
@backend_sampler(backend=QrispSimulatorBackend())
def main(k):
    captured = jnp.zeros((4,), dtype=jnp.float64) + k

    def kernel():
        q = QuantumFloat(2)
        h(q[0])
        return measure(q) + captured[0]

    return sample(kernel, shots=4)()
```

Result:

```text
RuntimeError: Could not identify loop-index position in sampling_body_func invars
```

The traced body inputs were:

```text
0: captured array  float64[4]
1: i               int64[]
2: acc             float64[4]
3: QuantumState
```

Since the accumulator output is `float64[4]`, backend_sampling_interpreter.py incorrectly selects input `0` as `acc_pos`. It then searches for `i` only before position `0`, finds nothing, and raises at backend_sampling_interpreter.py.

As a control, changing the captured array to shape `(3,)` succeeds and returns shape `(4,)`. So finding 1 is a valid, reproducible bug specifically for dynamic captured values whose aval matches the accumulator.

**Recommended Design (1.)**

The bug is caused by inferring `i` and `acc` from avals in backend_sampling_interpreter.py. The generated `fori_loop` already gives us a structural answer:

```python
sampling_body_func(i, args)
```

After flattening, the enclosing `while` body’s first loop input is always `i`, and the second is `acc`. Captured values may appear as additional call inputs, but they do not change the loop-carry order.

Therefore:

1. Find the `while` equation containing `sampling_body_func`.
2. Get its `body_jaxpr`.
3. Take `body_jaxpr.invars[0]` as the loop-index SSA variable.
4. Find that same variable in the `sampling_body_func` call’s `invars`.
5. Use that call-input position as `i_pos`.
6. Remove `acc_pos` inference entirely. It is not actually used by `_body_loop_evaluator`.

The mapping should be based on SSA variable identity, not `shape` and `dtype`.

**Patch Outline**

Add a helper that returns the sampling call and its loop-index position:

```python
def find_sampling_body_context(jaxpr):
    for eqn in jaxpr.eqns:
        if eqn.primitive.name == "while":
            body_jaxpr = eqn.params["body_jaxpr"]

            for body_eqn in body_jaxpr.eqns:
                if (
                    body_eqn.primitive.name in ("jit", "pjit")
                    and body_eqn.params.get("name") == "sampling_body_func"
                ):
                    index_var = body_jaxpr.invars[0]

                    for index, var in enumerate(body_eqn.invars):
                        if var is index_var:
                            return (
                                body_eqn.params.get("jaxpr")
                                or body_eqn.params.get("call_jaxpr"),
                                index,
                            )

                    raise RuntimeError(
                        "Could not map sampling loop index to sampling_body_func"
                    )

            # Recurse into nested control-flow bodies as needed.

    return None
```

Then replace the aval-based block at backend_sampling_interpreter.py with:

```python
body_jaxpr, i_pos = find_sampling_body_context(inner_jaxpr.jaxpr)
if body_jaxpr is None:
    raise RuntimeError("sampling_body_func not found inside eval function Jaxpr")

body_jaspr = Jaspr(body_jaxpr)
```

The evaluator can keep using:

```python
iteration = invalues[i_pos]
results = post_proc(meas_results[iteration], *invalues[:-1])
```

`extract_post_processing()` already expects all non-QuantumState arguments in Jaxpr order, so captured values should continue to be passed through unchanged.

**Regression Test**

Add a test to test_backend_sampling.py:

```python
def test_captured_array_matching_accumulator_shape():
    @backend_sampler(backend=_get_backend())
    def main(k):
        captured = jnp.zeros((4,), dtype=jnp.float64) + k

        def kernel():
            qf = QuantumFloat(2)
            h(qf[0])
            return measure(qf) + captured[0]

        return sample(kernel, shots=4)()

    result = main(2.0)

    assert result.shape == (4,)
    assert {float(value) for value in result} <= {2.0, 3.0}
```

This currently raises:

```text
RuntimeError: Could not identify loop-index position in sampling_body_func invars
```

With structural SSA mapping, it should return four samples successfully.

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,434 @@
...
+    loop-carried *i* and *acc* values, then feeds the pre-computed
+    measurement bits for that shot.  All accumulator typing, indexing,
+    and update logic is handled by the Jaspr itself.
+    """
+    body_jaxpr = find_named_jaxpr(inner_jaxpr.jaxpr, "sampling_body_func")
+    if body_jaxpr is None:
+        raise RuntimeError("sampling_body_func not found inside eval function Jaxpr")
+    body_jaspr = Jaspr(body_jaxpr)
+
+    # ── Pre-compute invar positions ─────────────────────────────────
```
renezander90: This logic is fragile. See comment above.

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
```diff
@@ -0,0 +1,434 @@
...
+                    "kernel's quantum circuit to be fully static — "
+                    "use :func:`~qrisp.jasp.jaspify` for workloads "
+                    "with real-time feedback."
+                ) from e
+            raise
+
+        # Run the backend ONCE.
+        raw = backend.run(qc, shots=shots)
+
+        # Expand & shuffle into a list of *shots* measurement arrays.
```
renezander90: Here is a significantly faster and more memory-efficient way to write this.

The original code suffers from a major JAX/NumPy anti-pattern: it creates thousands of tiny JAX arrays inside a Python loop, appends them to a Python list, and then uses `jnp.stack` on that massive list. This causes massive overhead and will severely bottleneck your performance.

You can vectorize this process by parsing the unique strings once, using `np.repeat` to duplicate the rows in bulk, and moving the data to JAX only at the very end.

### Improved Code

```python
import numpy as np
import jax.numpy as jnp

# Run the backend ONCE.
raw = backend.run(qc, shots=shots)

if not raw:
    meas_results_array = jnp.array([], dtype=jnp.bool_)
else:
    # 1. Separate keys and values
    bitstrings = list(raw.keys())
    counts = list(raw.values())
    
    # 2. Convert unique bitstrings to a 2D boolean NumPy array first
    # (String parsing is best kept in standard Python/NumPy)
    unique_bits = np.array([[c == "1" for c in bs] for bs in bitstrings], dtype=np.bool_)
    
    # 3. Vectorized expansion: duplicate the rows based on their counts
    expanded_bits = np.repeat(unique_bits, counts, axis=0)
    
    # 4. Shuffle the expanded array in-place along the shot axis
    np.random.shuffle(expanded_bits)
    
    # 5. Push the fully formed array to JAX (device) in one single operation
    meas_results_array = jnp.array(expanded_bits)

```

### Why this is better:

1. **No looping over shots:** `np.repeat` performs a fast, C-level block memory copy to expand the array, replacing `meas_results_list.extend()`.
2. **Minimal String Parsing:** Instead of parsing `"1"` and `"0"` for every single shot (e.g., 10,000 times), it only parses the *unique* bitstrings returned by the backend, which is exponentially faster.
3. **Single Device Transfer:** Pushing a fully constructed NumPy array to `jnp.array()` happens in one fast memory transfer to the accelerator (GPU/TPU/CPU). Calling `jnp.stack` on a list of hundreds or thousands of JAX arrays forces the JAX compiler to track massive computational graphs and causes severe latency.

src/qrisp/algorithms/qaoa/qaoa_problem.py
```diff
@@ -580,6 +581,8 @@ def run(
             The optimal result after running QAOA problem for a specific problem instance. It contains the measurement results after applying the optimal QAOA circuit to the quantum argument.
 
         """
+        from qrisp import QuantumVariableTemplate
```
renezander90: Move to module-level imports

tests/jax_tests/test_backend_sampling.py
```diff
@@ -0,0 +1,794 @@
...
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+import jax.numpy as jnp
+import numpy as np
+import pytest
+
+from qrisp import *
```
renezander90: Using star imports (`from module import *`) in your test files—or any Python files—is widely considered a bad practice in software engineering.

Here is why you should avoid it, especially in `pytest` suites:

* **Namespace Collisions:** If `qrisp` and `qrisp.jasp` contain functions, variables, or classes with the identical names, the second import will silently overwrite the first. This leads to confusing bugs where your test is evaluating a completely different function than you intended. It can also accidentally overwrite built-in Python functions or `pytest` fixtures.
* **Zero Traceability:** Unit tests act as living documentation for your code. When another developer (or future you) reads `QuantumFloat()` in a test, a star import makes it impossible to know if that object came from `qrisp`, `qrisp.jasp`, or a local helper without manually searching the entire codebase.
* **Broken Tooling:** Linters (like Flake8 or Ruff) and type checkers (like Mypy) cannot reliably analyze code with star imports. Consequently, your IDE will struggle to provide accurate auto-complete, go-to-definition features, and automated refactoring.

### The Recommended Approach

Instead, explicitly declare your dependencies using one of these two methods:

**Approach 1: Explicit Imports (Best for readability)**
Import exactly the classes and functions you need for that specific test file.

```python
from qrisp import QuantumFloat, QuantumCircuit
from qrisp.jasp import specific_jasp_function

def test_quantum_float_initialization():
    qf = QuantumFloat(5)
    # ...

```

**Approach 2: Module Aliasing (Best if utilizing a large API surface)**
Import the module itself under a shorthand alias, which keeps the namespace clean while making it obvious where each function originates.

```python
import qrisp as qp
import qrisp.jasp as jasp

def test_quantum_float_initialization():
    qf = qp.QuantumFloat(5)
    # ...

```

tests/jax_tests/test_backend_sampling.py
```diff
@@ -0,0 +1,794 @@
...
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return sample(kernel, shots=500)()
+
+    res = main()
+    assert res.shape == (500,)
+    vals = {int(v) for v in res}
+    assert len(vals) >= 6, f"expected ≥6 values from 0-7, got {len(vals)}"
```
renezander90: Does this make sense?

tests/jax_tests/test_backend_sampling.py
```diff
@@ -0,0 +1,794 @@
...
+    def kernel():
+        qf = QuantumFloat(4)
+        h(qf[0])
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return sample(kernel, shots=0)()
+
+    try:
```
renezander90: Use `pytest.raises`

tests/jax_tests/test_backend_sampling.py
```diff
@@ -0,0 +1,794 @@
...
+def test_raises_without_sample():
+    """@backend_sampler raises if the function doesn't use sample()"""
+
+    @backend_sampler(backend=_get_backend())
+    def no_sample():
+        qf = QuantumFloat(4)
+        h(qf[0])
+        return measure(qf)
+
+    try:
```
renezander90: Use `pytest.raises`

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+import jax.numpy as jnp
+import numpy as np
+
+from qrisp import *
```
renezander90: Avoid star imports.

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+        qf = QuantumFloat(4)
+        qf[:] = 5
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return expectation_value(kernel, shots=200)()
+
+    res = main()
+    assert 4.5 < float(res) < 5.5, f"expected ~5.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 5.0)
``

File: tests/jax_tests/test_backend_sampling_ev.py [hunk 19]
```diff
@@ -0,0 +1,404 @@
...
+
+    @jaspify
+    def jaspify_main():
+        return expectation_value(kernel, shots=300)()
+
+    r_backend = float(backend_main())
+    r_jaspify = float(jaspify_main())
+
+    # Both should be ~1.5 (avg of 0,1,2,3)
+    assert abs(r_backend - r_jaspify) < 0.3, f"backend={r_backend:.3f}, jaspify={r_jaspify:.3f}, diff too large"
```
renezander90: The comments after the assertion are not needed (here, and for the other tests).

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+    # 0^2 and 1^2 equally likely → expectation ≈ 0.5
+    assert 0.3 < float(res) < 0.7, f"expected ~0.5, got {res}"
+
+
+# ===========================================================================
+# Comparison against jaspify
+# ===========================================================================
+
+
+def test_ev_compare_jaspify():
```
renezander90, line 176: The probability of this test failing randomly is roughly 0.1% (or about 1 in 1,000).

While 1 in 1,000 might sound low, in a CI/CD pipeline that runs multiple times a day across different branches, this is a flaky test. It will randomly fail on perfectly good code, causing developer frustration.

In general, many tests in this file are probabilistic, with very low failure probabilities. But, if a couple have "large" failure probabilities like this test, it will become an issue.

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+        qf = QuantumFloat(size)
+        qf[:] = 3
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main(s):
+        return expectation_value(kernel, shots=100)(s)
+
+    res = main(4)
+    assert 2.5 < float(res) < 3.5, f"expected ~3.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 3.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def kernel():
+        qf = QuantumFloat(4)
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return expectation_value(kernel, shots=0)()
+
+    try:
```
renezander90: Use `pytest.raises`. "Zero shots should raise or return NaN." Which is the actual behavior?

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+    @backend_sampler(backend=_get_backend())
+    def main():
+        ev_a = expectation_value(kernel_a, shots=100)()
+        ev_b = expectation_value(kernel_b, shots=100)()
+        return ev_a, ev_b
+
+    ra, rb = main()
+    assert 4.0 < float(ra) < 6.0, f"expected ~5.0, got {ra}"
+    assert 9.0 < float(rb) < 11.0, f"expected ~10.0, got {rb}"
+
```
renezander90:
```suggestion
    assert np.isclose(ra, 5.0)
    assert np.isclose(rb, 11.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+        qf = QuantumFloat(4)
+        qf[:] = 7
+        return measure(qf)
+
+    @backend_sampler(backend=custom)
+    def main():
+        return expectation_value(kernel, shots=50)()
+
+    res = main()
+    assert 6.0 < float(res) < 8.0, f"expected ~7.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 7.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+        qf = QuantumFloat(2)
+        qf[:] = 1
+        return measure(qf)
+
+    @backend_sampler(backend=QrispSimulatorBackend())
+    def main():
+        return expectation_value(kernel, shots=50)()
+
+    res = main()
+    assert 0.5 < float(res) < 1.5, f"expected ~1.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 1.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+    @backend_sampler(backend=_get_backend())
+    def main():
+        samples = sample(kernel, shots=30)()
+        ev = expectation_value(kernel, shots=100)()
+        return samples, ev
+
+    samples, ev = main()
+    assert samples.shape == (30,)
+    assert 2.0 < float(ev) < 4.0, f"expected ~3.0, got {ev}"
+    assert all(float(v) == 3.0 for v in samples)
```
renezander90:
```suggestion
    assert np.isclose(ev, 3.0)
    assert np.allclose(samples, 3.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+        qf = QuantumFloat(3)
+        x(qf)
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return expectation_value(kernel, shots=100)()
+
+    res = main()
+    assert float(res) == 7.0, f"expected 7.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 7.0)
```

tests/jax_tests/test_backend_sampling_ev.py
```diff
@@ -0,0 +1,404 @@
...
+    def kernel():
+        qf = QuantumFloat(5)
+        return measure(qf)
+
+    @backend_sampler(backend=_get_backend())
+    def main():
+        return expectation_value(kernel, shots=100)()
+
+    res = main()
+    assert float(res) == 0.0, f"expected 0.0, got {res}"
```
renezander90:
```suggestion
    assert np.isclose(res, 0.0)
```

renezander90 (changes requested):

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
positr0nium: 1. Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/a60c71a998b78b5c9e555356d4b129c7afbb296d
2. There is not really a user-facing way to transform the resulting Jaxp - it is purely internal. Besides: The ``measure`` primitive also consumes randomness.
3. ``return_dict`` is not a user-facing argument (it is not mentioned in the docstring). I agree that this is ugly but fixing the uglyness here is out of scope for this PR. There is now a better error message: https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/db352ad9979afb0ded81711e39eb5afaaafa637f

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/a60c71a998b78b5c9e555356d4b129c7afbb296d

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/fe6553b2bb0c8c870edc667ef9201d4d9a89a7a1

src/qrisp/algorithms/qaoa/qaoa_problem.py
positr0nium: Fixed in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/12ddece374b4003552f7d35c15410faa59207611

tests/jax_tests/test_backend_sampling.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Imrproved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Improved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3b27fb743b6e9e8d4192065e3e86139963544d73

tests/jax_tests/test_backend_sampling_ev.py
positr0nium: Resolved in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/3ff24b58bd3caaaae1315870031cef6f77ca1ece

src/qrisp/jasp/interpreter_tools/interpreters/backend_sampling_interpreter.py
renezander90: Regarding 2., Claude would favour `jax.experimental.io_callback`:

```python
outvals = pure_callback(fn, result_shapes, *invalues)
```

Reviewing that usage, since the JAX docs are explicit about the contract:

1. **Raising exceptions inside a `pure_callback` is undefined behavior.** JAX's docs state: *"intentionally raising an error within a pure_callback breaks the API contract, and the behavior of the resulting program is undefined."* The wrapped `fn` (`backend_sampling_fn`) explicitly raises `ValueError`/`RuntimeError` for invalid shot counts, real-time-feedback kernels, and extraction failures — all inside the callback body. Because the whole wrapper is compiled with `jax.jit` at backend_sampling.py (`res = jit(eval_fn)(*flat_args)`), these validation errors run inside a jitted `pure_callback`, so their propagation to the caller isn't guaranteed by JAX's contract — this is the most concrete risk here.

2. **The callback is not actually pure.** It calls `backend.run(...)` against real/simulated quantum hardware and shuffles results with `np.random.shuffle` — a side-effecting, non-deterministic operation. JAX explicitly documents this primitive as assuming purity: outputs may be skipped entirely via dead-code elimination if unused, or (under `vmap`/similar transforms) invoked more than once for what looks like a single logical call. JAX's own docs point to `jax.experimental.io_callback` as *"callback designed for impure functions"* — that would be the semantically correct primitive for something that fires a hardware/simulator job, whereas `pure_callback` gives no guarantee the backend is called exactly once per logical invocation.

3. **`vmap_method` is left unspecified** in the `pure_callback(fn, result_shapes, *invalues)` call. JAX currently defaults to `"sequential"` but documents this as deprecated, with a future version raising `NotImplementedError` unless set explicitly. If `backend_sampler`-decorated functions are ever vmapped, this call will start failing.

4. **`result_shapes` assumes a single output.** `ShapeDtypeStruct(eqn.outvars[0].aval.shape, eqn.outvars[0].aval.dtype)` only inspects `outvars[0]`, silently relying on `sampling_eval_function`/`expectation_value_eval_function` always producing exactly one outvar (true today per this file's own docstring example, but not asserted).

Items 1 and 2 are the ones worth acting on if you want this hardened — using `io_callback` instead of `pure_callback` would resolve both the purity mismatch and make exception propagation part of the documented contract instead of undefined behavior.

Regarding `io_callback` being part of `jax.experimental`:

Fair point — `io_callback` lives under `jax.experimental`, and the codebase currently has no other `jax.experimental` imports (I checked), so it would be a new dependency surface for qrisp. A few clarifications on what "experimental" means in practice here:

- It's been in that namespace for several JAX releases, is fully documented in the [external callbacks guide](https://docs.jax.dev/en/latest/external-callbacks.html) with a dedicated comparison table, and is the officially recommended callback for impure functions — it's not a fringe/unmaintained API, just one JAX hasn't promoted to the top-level `jax` namespace yet (unlike `pure_callback`, `debug.callback`, `debug.print`).
- Its guarantees are actually stronger where they matter here: "guaranteed execution" (✅ vs `pure_callback`'s ❌) and it's explicitly documented as *"appropriate for impure functions: e.g. functions which read or write data to disk."* Running a hardware/simulator job is squarely that category.
- Trade-offs versus `pure_callback`: no `grad` support (irrelevant — sampling isn't differentiated), and `vmap` only works with `ordered=False` (fine unless `backend_sampler` is vmapped, which isn't a documented use case today).

So the "not stable" concern is real as an API-surface/namespace-stability argument, but it doesn't outweigh the correctness issue: today, `pure_callback`'s own docs explicitly say not to use it to raise runtime errors, and `_make_backend_sampling_fn`'s callback does exactly that (shots validation, real-time-feedback detection). If you'd rather not take on `jax.experimental` as a dependency, the alternative is to move those validation checks out of the callback and into the tracing-time / outer decorator path (before `pure_callback` is invoked), keeping `pure_callback` for the actual backend call but never raising inside it.

I checked that `from jax.experimental import io_callback`is available in JAX 0.7.1.

Also tagging our JAX expert @PietropaoloFrisoni

positr0nium: Ok, the io callback looks like the better option

positr0nium: Implemented in https://github.com/eclipse-qrisp/Qrisp/pull/744/commits/10fd96338a1b344b1a3a01ff68ea0a818e6bae6c

renezander90 (approved):
LGTM 🎊

PR #784, opened 2026-08-05
Layerization pass

positr0nium:
## Description

Add two circuit passes and the scheduler behind them.

`layerize` reorders instructions into as-soon-as-possible (ASAP) layers: gates acting on disjoint qubits are pulled into the same layer, compacting the Stim timeline diagram without changing circuit semantics. With `insert_barriers=True` it writes the schedule back into the circuit as barriers, giving one `TICK` per time step.

`promote_barriers` widens every barrier to span the full circuit. This matters because a barrier constrains only the qubits it names, while a Stim `TICK` is a global time-step boundary — so `to_stim` now emits a `TICK` for full-width barriers only. That is what makes Jasp-traced circuits annotatable: inserting a full-width barrier in Jasp mode is impractical, so a traced barrier is usually partial, and promoting it on the extracted circuit turns round boundaries back into `TICK`s.

## Related Issues

Closes # 
Related to # 

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [x] Yes
- [ ] No

**A partial barrier no longer becomes a Stim `TICK`.** `to_stim` emits a `TICK` for full-width barriers only.

A barrier constrains the qubits it names; a `TICK` is a global time-step boundary. The two coincide exactly for a full-width barrier, and treating a partial one as a boundary claimed a synchronisation the circuit never asked for.

*Migration:* Code that relied on a partial barrier intended to produce a ``TICK`` should run the new `promote_barriers` pass, which widens every barrier to the full circuit. Circuits coming out of Jasp tracing always need this, since tracing can only write a barrier over a static qubit list.

## What was changed?

**`layerize`** — `passes/layerize.py`. A pass factory, so that `insert_barriers` can be configured.

**`promote_barriers`** — `passes/promote_barriers.py`. Its docstring carries a warning box: promotion adds scheduling constraints the circuit did not have, widening the schedule and inflating the idle-noise budget, and it promotes *every* barrier including compiler-inserted ones.

**Scheduler** — `pass_management/scheduling.py`. Walks the circuit's dependency graph and fills one layer at a time. Per qubit or clbit a layer takes one gate, one error channel, and any number of annotations; a barrier ends the layer for the qubits it names. Working on the dependency graph is what keeps the reordering faithful to the input.

**Stim converter** — `TICK` only for full-width barriers, per the breaking change above.

**Documentation** — `QuantumCircuit.barrier` gains the scope rule and the full-width→`TICK` rule; `to_stim` gains a "Time steps" section recording that a `TICK` is a pure annotation and that `find_detectors` strips and regenerates them. New `promote_barriers` page, both index tables, changelog.

## How was it tested?

108 unit tests plus 8 doctests. `tests/circuit_tests/` + `tests/interface_tests/`: **1673 passed, 3 skipped**.

| Test-ID | Status |
|---------|--------|
| `test_scheduling.py` — layer assignment, budget, barriers, error channels, emission order, classification (54) | ✅ |
| `test_layerize.py` — reordering, correctness, barriers, `insert_barriers`, parallel rendering, detector semantics (40) | ✅ |
| `test_promote_barriers.py` — widening, idempotence, effects on schedule and `TICK`s (14) | ✅ |
| doctests — `scheduling` (4), `layerize`, `promote_barriers`, `QuantumCircuit.barrier`, `is_full_width_barrier` (8) | ✅ |

Before:
layerize_before

After:
layerize_after

The canonical shape `insert_barriers=True` produces, one `TICK` per time step with every channel beside its gate:

```
CX 0 1
DEPOLARIZE2(0.01) 0 1
TICK
X_ERROR(0.005) 1
M 1
TICK
```

## Checklist

- [ ] All tests pass locally

## Reviewer Notes

- **The converter change and `promote_barriers` have to ship together.** On its own the converter change silently removes the round boundaries of every Jasp-traced circuit, and `promote_barriers` is inert without barriers being scoped per qubit.
- **`promote_barriers` is a consent gate, not a fix** for noise-model inflation. It turns a silent surprise into a declared choice; it does not reduce the noise budget.
- **Layer indices are not monotone along `qc.data`** — two independent chains written back to back both start at layer 0. A coherent `TICK` stream only exists after reordering, which is why `insert_barriers` lives in this pass.
- ASAP remains intentionally conservative: it never reorders instructions sharing a qubit. Exploiting gate commutation would need permeability analysis.
- **Known pre-existing issue, not addressed:** the `QuantumCircuit.to_stim` doctest hardcodes `qb_0`/`qb_1` while the global qubit counter yields whatever is next in the process, so it fails under `--doctest-modules` at `main` too. Several other passes' docstrings share the fragility.


purva-thakre:
@positr0nium This branch has merge conflicts.

positr0nium:
I know - it is sitting on top of a serious of PRs that are waiting to be reviewed by @renezander90 😇

purva-thakre:
Ah, thanks for the clarification. I did not know this.

renezander90:
If I understand it correctly, the content of this PR is not related to the other PRs? In this case, it would have been better to use a branch from current main. Otherwise, we have to wait until we work through the other PRs.

positr0nium:
It is not super urgent that we merge the PR, so having it in order is not a big deal. However, I would like to work with these tools ASAP and I would like to avoid creating some local branch chaos in my machine.

purva-thakre:
@coderabbit review

PR #785, opened 2026-08-05
Error insertion pass

positr0nium:
## Description

Add an `insert_stim_noise` circuit pass that annotates a `QuantumCircuit` with the circuit-level depolarizing noise model used for quantum error-correction benchmarks. Every qubit receives exactly one noise channel per physical time step: `DEPOLARIZE2` after a two-qubit gate, `DEPOLARIZE1` after a one-qubit gate or per idle layer, `X_ERROR` before a measurement and after a reset. The channels are `StimNoiseGate` operations, which act as identities for the Qrisp simulator and become genuine noisy channels under `to_stim()`.

The pass **only inserts** — the instruction order of the input circuit, and with it the Stim measurement record, is preserved.

## Related Issues

Closes # 
Related to # 

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

Purely additive: one new pass, one new export, new documentation pages. No existing behaviour is modified.

## What was changed?

- New pass `insert_stim_noise` at `src/qrisp/circuit/pass_management/passes/insert_stim_noise.py` (678 lines incl. docstrings). A factory returning a `CircuitPass`, parameterised by `depolarize_1_strength`, `depolarize_2_strength`, `X_error_strength` and `only_necessary`. 
- Instructions with no physical time step are passed through without noise and without advancing the layer clock: `qb_alloc` / `qb_dealloc`, `gphase` (a one-qubit `Operation` in Qrisp, but only a global phase) and `parity` (which becomes a Stim `DETECTOR` / `OBSERVABLE_INCLUDE`, i.e. classical post-processing). Without this, a `DETECTOR` would occupy a layer of its own and collect a full round of idle `DEPOLARIZE1` on every qubit — one spurious noise layer per detector in a QEC circuit.
- `only_necessary=True` (default) suppresses a channel only on an **exact** match: same Stim channel type, on exactly the same qubits, in the expected position (after the gate / before the measurement / after the reset). The error probability is deliberately not compared, so user-placed noise of a different strength is respected. Exactness matters because the rules overlap positionally — the instruction after a gate is also the instruction before whatever comes next — so a looser match would let a single user-placed `X_ERROR` count both as a gate's depolarizing channel and as the following measurement's bit flip, silently dropping one of the two channels the model calls for.
- 31 unit tests at `tests/circuit_tests/passes/test_insert_stim_noise.py` (499 lines).

10 files changed, 1409 insertions(+), 0 deletions(-).

## How was it tested?

31 unit tests, all passing. `tests/circuit_tests/` as a whole: 849 passed, 2 skipped.

| Test-ID | Status |
|---------|--------|
| Factory behaviour (2 tests) — returns a `CircuitPass`, rejects probabilities outside `[0, 1]` | ✅ |
| Noise insertion rules (8 tests) — `DEPOLARIZE1`/`DEPOLARIZE2` per gate arity, idle noise, `X_ERROR` before measurement / after reset, one noise per qubit per layer, gates preserved, >2-qubit gate raises | ✅ |
| Order and purity (2 tests) — instruction order and Stim measurement map unchanged, input circuit not mutated | ✅ |
| Instructions without a time step (4 tests) — `parity`, `gphase`, `gphase`-only circuit, `qb_alloc` / `qb_dealloc` | ✅ |
| Barriers (3 tests) — passed through without noise, enforces layer separation, no extra layer between two gates on one qubit | ✅ |
| `only_necessary` matching (9 tests) — skips after one-/two-qubit gate, before measurement, after reset; requires matching channel type and matching qubit set; ignores strength; `only_necessary=False` applies the full model; idle noise still added | ✅ |
| Composition and equivalence (3 tests) — `PassManager` with `compress_layers`, `compare_unitary`, `compare_measurement` | ✅ |

One round of repetition-code syndrome extraction, `insert_stim_noise(0.001, 0.01, 0.005)`:

```
CX 0 1
DEPOLARIZE2(0.01) 0 1
DEPOLARIZE1(0.001) 2
CX 2 1
DEPOLARIZE2(0.01) 2 1
X_ERROR(0.005) 1
M 1
DETECTOR rec[-1]
R 1
X_ERROR(0.005) 1
DEPOLARIZE1(0.001) 0 0 0 2 2
TICK
```

Four time steps (two `CX` layers, the measurement, the reset). The ancilla `1` is busy in all four and collects four channels; data qubits `0` and `2` are busy in one each and idle through three. The `DETECTOR` correctly produces no noise layer.

Two output characteristics are documented in the docstring `Notes`, since both look surprising at first:

- **Repeated targets.** Stim merges consecutive identical channels, so `DEPOLARIZE1(0.001) 0 0 0 2 2` applies the channel three times to qubit `0` and twice to qubit `2` — equivalent to five separate instructions, one per idle time step.
- **Placement of idle noise.** Because instruction order is preserved, idle noise appears at the latest still-correct position rather than distributed across the round. Nothing acts on the qubit in between, so the channels compose identically; run `compress_layers` afterwards to distribute them over the timeline.

Timeline diagrams of two rounds, before and after the pass, are embedded in the rendered docs:
`documentation/source/_static/insert_stim_noise_before.svg`, `insert_stim_noise_after.svg`.

## Checklist

## Reviewer Notes

Worth focusing on:

1. **Order preservation and lazy idle-noise flushing** — the central design decision, described in the module docstring. The correctness argument is that noise on a qubit only needs to be ordered relative to other instructions on *that* qubit, plus a monotonicity property of the scheduler (for a fixed qubit, its instructions appear in `qc.data` with non-decreasing layer indices) which is what makes a single forward cursor per qubit sound.

   For context on whether preserving order was necessary: it is robustness rather than correctness. Every in-tree `to_stim()` consumer is already order-independent — `stim_backend.py` permutes sample columns through `measurement_map`, `find_detectors.py` inverts it to recover `Clbit` sets, and `stim_extraction.py` rewrites returns through all three maps. Only someone holding a bare `to_stim()` result and writing `rec[-k]` offsets by hand would notice. The benefit is that the output is now the input plus insertions, which keeps diffs and pass composition predictable.

2. **The transparent-instruction list** (`_TRANSPARENT_OP_NAMES`) — neither `gphase` nor `parity` can be recognised structurally (`gphase` really does carry a qubit; `parity` carries clbits), which is why they are named explicitly. This is the fix for the spurious per-detector noise layer.

3. **Qubit liveness is deliberately not tracked.** `qb_alloc` / `qb_dealloc` pass through, but idle noise covers the whole register in every time step regardless of allocation state. This matches the fixed physical qubit register of the intended use case and is recorded under "Deliberate non-goals" in the module docstring. It would need revisiting only if the pass were pointed at circuits programming *on top of* error correction.


purva-thakre:
si1000 is defined in appendix A of https://arxiv.org/abs/2312.04522

PR #788, opened 2026-08-06
Fix `QiskitBackend` submission for providers that report non-hex `memory`

positr0nium:
## Description

`QiskitBackend` submitted circuits through Qiskit's `BackendSamplerV2` primitive, which rebuilds
counts from the result's `memory` field and parses each entry as a **hexadecimal** string. Providers
that write plain **binary** bitstrings there — `qiskit-iqm` among them — either crashed the job or
returned silently wrong counts. Submission now goes through the wrapped backend's own `run()` and
reads counts from `Result.get_counts()`, which is unambiguous. `QiskitRuntimeBackend` keeps
`SamplerV2`, because IBM Runtime offers no other route.

## Related Issues

Closes #
Related to #

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

No public API changes. `Backend`, `run()`, `run_async()`, `Job`, and the measurement result types are
untouched, and counts returned for conformant backends (Aer, fake backends, IBM Runtime) are
unchanged.

`QiskitBackend.sampler` was removed, but it was never public: `QiskitBackend.rst` uses a bare
`.. autoclass::` with no `:members:`, and `conf.py` sets `autodoc_default_options` to
`members: False` / `undoc-members: False` with `numpydoc_show_class_members = False`. Only the class
docstring and `run_async` were ever rendered, and neither documented the attribute — confirmed
against the built HTML. `QiskitRuntimeBackend.sampler` is unaffected and still holds the IBM Runtime
`SamplerV2`.

## What was changed?

- `QiskitBackend` now submits via `self.backend.run(circuits, shots=...)` instead of
  `BackendSamplerV2.run(...)`, and reads counts from `Result.get_counts()`. Qrisp never consumed
  per-shot `memory`, so nothing is lost by not reconstructing it.
- Added two paired extension points: `QiskitBackend._submit()` (how circuits are handed to the
  provider) and `QiskitBackend._job_class` (which `QiskitJob` subclass parses the result).
- Split result parsing by result *type* rather than by flag: `QiskitJob._extract_counts()` reads a
  `Result`; the new `QiskitRuntimeJob` subclass overrides it to read a `PrimitiveResult` `DataBin`
  (the previous logic, relocated unchanged).
- `QiskitRuntimeBackend` pairs `_submit()` → `self.sampler.run(...)` with
  `_job_class = QiskitRuntimeJob`. It must keep the primitive: `IBMBackend.run()` still exists in
  `qiskit-ibm-runtime` 0.47 but raises `IBMBackendError: Support for backend.run() has been removed`.
- Extracted the shared register-separator stripping into `_merge_counts()`.
- A backend whose `run()` returns something without `get_counts()` now raises a `TypeError` naming
  `_extract_counts()`, rather than an opaque `AttributeError` deep in the parsing loop.
- Removed the now-unused `BackendSamplerV2` / `BackendV2` imports and the `QiskitBackend.sampler`
  attribute.
- Docs: updated the `QiskitBackend` class docstring (with a note explaining *why* the primitive is
  avoided, so it does not get reinstated) and the narrative section in
  `documentation/source/reference/Backend Interface/index.rst`.

## How was it tested?

| Test | Status |
|------|--------|
| `TestQiskitBackendNonHexMemory::test_binary_memory_does_not_overflow` | ✅ |
| `TestQiskitBackendNonHexMemory::test_binary_memory_is_not_silently_misread` | ✅ |
| `TestQiskitBackendRunAsync::test_run_async_submits_via_backend_run_not_a_sampler` | ✅ |
| `TestQiskitJob::test_result_parses_batch_from_backend_result` | ✅ |
| `TestQiskitJob::test_result_rejects_result_without_get_counts` | ✅ |
| `TestQiskitJob::test_result_parses_counts_from_sampler_result` | ✅ |
| `TestQiskitJob::test_result_parses_batch_from_sampler_result` | ✅ |

The two `TestQiskitBackendNonHexMemory` tests drive a stub Qiskit backend that reports results in
`qiskit-iqm`'s exact shape — binary `memory`, binary `Counts` keys, and a header carrying no
`creg_sizes`/`memory_slots`. No hardware or network access is required.

Both were confirmed to fail when `QiskitBackend` is forced back onto `BackendSamplerV2`, so they are
load-bearing rather than tautological:

```
4q GHZ   via SamplerV2 -> JobFailureError: ... int too big to convert   [RAISES]
2q Bell  via SamplerV2 -> {'00': 50, '01': 50}                          [WRONG COUNTS]
```

Existing coverage was reworked, not reduced: the fixture that mocked `BackendSamplerV2` now mocks
`run()` on a real `AerSimulator`, so `transpile()` still runs against a real target. Test count went
82 → 89 with none removed.

Full suite: **776 passed, 1 skipped** (`tests/interface_tests/` + `tests/algorithms_tests/cd_tests/test_cold.py`).
The skip is `qiskit-ibm-runtime`-gated and pre-existing.

The original failure, on a 4-qubit GHZ circuit submitted to IQM Sirius:

```
File .../qiskit/primitives/backend_sampler_v2.py:313, in _memory_array
--> 313  data = b"".join(int(i, 16).to_bytes(num_bytes, "big") for i in memory)
OverflowError: int too big to convert

JobFailureError: Qiskit job '1ec7fa55-...' failed: int too big to convert
```

The mechanism: `IQMJob.result()` writes `'1111'` into `memory`; Qiskit allocates one byte for a
4-bit classical register; `int('1111', 16) == 4369` does not fit.

The quieter half of the same bug — a Bell state, no exception raised at all:

```
true distribution : {'00': 50, '11': 50}
via QiskitBackend : {'00': 50, '01': 50}      # int('11', 16) == 17, fits in one byte
```

## Checklist
- [x] Breaking changes are documented with migration path — *n/a, no breaking changes*

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -49,6 +49,18 @@ Bug Fixes
   in ``CircuitOperation`` calls
   (`PR #709 <https://github.com/eclipse-qrisp/Qrisp/pull/709>`_).
 
+* Fixed :class:`~qrisp.interface.QiskitBackend` failing with
+  ``OverflowError: int too big to convert`` — or, for small classical
+  registers, silently returning wrong counts — on providers that report
+  measurement results as binary rather than hexadecimal strings, such as
+  ``qiskit-iqm``.  Circuits are now submitted through the wrapped backend's
+  own ``run()`` method and counts are read via ``Result.get_counts()``,
+  instead of going through Qiskit's ``BackendSamplerV2`` primitive, which
+  rebuilds counts from the hex-encoded ``memory`` field.
+  :class:`~qrisp.interface.QiskitRuntimeBackend` continues to use
+  ``SamplerV2``, which IBM Runtime requires
+  (`PR #XXX <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
+
 Compatibility
 -------------
 
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -58,8 +58,11 @@ Bug Fixes
   instead of going through Qiskit's ``BackendSamplerV2`` primitive, which
   rebuilds counts from the hex-encoded ``memory`` field.
   :class:`~qrisp.interface.QiskitRuntimeBackend` continues to use
-  ``SamplerV2``, which IBM Runtime requires
-  (`PR #788 <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
+  ``SamplerV2``, which IBM Runtime requires.  Passing a real IBM Quantum
+  backend to :class:`~qrisp.interface.QiskitBackend` now raises a ``TypeError``
+  pointing at :class:`~qrisp.interface.QiskitRuntimeBackend`; IBM *fake*
+  backends are unaffected
+  (`PR #788 <https://github.com/eclipse-qrisp/Qrisp/pull/788>`_).
 
 Compatibility
 -------------
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -59,7 +59,7 @@ Bug Fixes
   rebuilds counts from the hex-encoded ``memory`` field.
   :class:`~qrisp.interface.QiskitRuntimeBackend` continues to use
   ``SamplerV2``, which IBM Runtime requires
-  (`PR #XXX <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
+  (`PR #788 <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
 
 Compatibility
 -------------
--- a/documentation/source/reference/Backend Interface/index.rst
+++ b/documentation/source/reference/Backend Interface/index.rst
@@ -613,7 +613,9 @@ For more details, see the :ref:`BatchedBackend` documentation.
 :class:`~qrisp.interface.QiskitBackend` is a :ref:`Backend` that wraps any
 Qiskit-compatible backend (simulators or real hardware). Circuits are converted
 directly to Qiskit ``QuantumCircuit`` objects, transpiled for the target device,
-and submitted through the backend's own ``run()`` method.
+and submitted through the backend's own ``run()`` method. Real IBM Quantum
+hardware is the exception, since IBM removed support for ``run()``: use
+:class:`~qrisp.interface.QiskitRuntimeBackend` for those.
 
 .. code-block:: python
 
--- a/documentation/source/reference/Backend Interface/index.rst
+++ b/documentation/source/reference/Backend Interface/index.rst
@@ -613,7 +613,7 @@ For more details, see the :ref:`BatchedBackend` documentation.
 :class:`~qrisp.interface.QiskitBackend` is a :ref:`Backend` that wraps any
 Qiskit-compatible backend (simulators or real hardware). Circuits are converted
 directly to Qiskit ``QuantumCircuit`` objects, transpiled for the target device,
-and submitted through Qiskit's ``SamplerV2`` primitive.
+and submitted through the backend's own ``run()`` method.
 
 .. code-block:: python
 
--- a/src/qrisp/interface/provider_backends/qiskit_backend.py
+++ b/src/qrisp/interface/provider_backends/qiskit_backend.py
@@ -398,8 +398,8 @@ class QiskitBackend(Backend):
 
         shots : int or list[int] or None, optional
             Number of shots.  If ``None``, the backend's ``shots`` option is
-            used. If a ``list[int]`` is provided, the Qiskit sampler does not
-            support per-circuit shot counts, so all circuits are run at
+            used. If a ``list[int]`` is provided, Qiskit applies a single shot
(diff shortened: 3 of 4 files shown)


documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -49,6 +49,18 @@ Bug Fixes
...
+  ``OverflowError: int too big to convert`` — or, for small classical
+  registers, silently returning wrong counts — on providers that report
+  measurement results as binary rather than hexadecimal strings, such as
+  ``qiskit-iqm``.  Circuits are now submitted through the wrapped backend's
+  own ``run()`` method and counts are read via ``Result.get_counts()``,
+  instead of going through Qiskit's ``BackendSamplerV2`` primitive, which
+  rebuilds counts from the hex-encoded ``memory`` field.
+  :class:`~qrisp.interface.QiskitRuntimeBackend` continues to use
+  ``SamplerV2``, which IBM Runtime requires
+  (`PR #XXX <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
```
PietropaoloFrisoni:
```suggestion
  (`PR #788 <https://github.com/eclipse-qrisp/Qrisp/pull/XXX>`_).
```

PietropaoloFrisoni (approved):

renezander90:
The QiskitBackend should prioritize compatibility with IBM backends. 

The documentation states: ``This allows easy access to any Qiskit-compatible simulator or real quantum hardware through the Qrisp backend interface.``

However, the `BackendSamplerV2` itself is **not compatible with real IBM-backends** (it used to be until recently):

```
from qrisp import QuantumFloat
from qiskit_ibm_runtime import QiskitRuntimeService
from qrisp.interface import QiskitBackend

instance = "XYZ"
channel = "ibm_quantum_platform"
token = "XYZ"

# 1. Initialize the Runtime Service
service = QiskitRuntimeService(channel=channel, instance=instance, token=token)

# 2. Select a real quantum backend (e.g., the least busy one)
backend = service.least_busy(operational=True, simulator=False)
print(backend)
ibm_backend = QiskitBackend(backend=backend)

qf = QuantumFloat(2)
qf[:] = 2
res = qf * qf
res.get_measurement(backend=ibm_backend)
```

> JobFailureError: Qiskit job '502feb88-87ef-4176-abc3-9589359e0793' failed: 'Support for backend.run() has been removed. Please see our migration guide https://quantum.cloud.ibm.com/docs/migration-guides/qiskit-runtime for instructions on how to migrate to the primitives interface.

It seems that only the `SamplerV2` from `qiskit_ibm_runtime` is compatible with real ibm backends. Hence, replacing `BackendSamplerV2` with `SamplerV2` in the `QiskitBackend` would resolve the current IBM-backend incompatibility:

```
        #self.sampler = BackendSamplerV2(backend=cast(BackendV2, backend))
        from qiskit_ibm_runtime import SamplerV2
        self.sampler = SamplerV2(backend)
```

Wouldn't the actual problem be, that `qiskit-iqm` is incompatible with the `SamplerV2`? (And does the problem even persist if we substitute `BackendSamplerV2` by `SamplerV2`?)

renezander90 (changes requested):
I initially had concerns that moving away from the V2 primitives in QiskitBackend would break compatibility with real IBM hardware, as IBM has entirely removed support for backend.run().

I realize now that this is strictly handled by QiskitRuntimeBackend, which rightfully keeps the SamplerV2 primitive. However, this does highlight a potential UX trap. It is very easy for a user to accidentally pass an IBM backend to the generic QiskitBackend like this:

```
from qiskit_ibm_runtime import QiskitRuntimeService
from qrisp.interface import QiskitBackend

service = QiskitRuntimeService(...)
backend = service.least_busy(operational=True, simulator=False)

# Accidentally using the generic backend for IBM hardware
ibm_backend = QiskitBackend(backend=backend)
```

Since we want to maintain easy access to real quantum hardware, would it make sense to add a safeguard in QiskitBackend.__init__ and a note in the documentation? For instance, if it detects an instance of a real IBM backend, it could:
**Raise a helpful TypeError** explicitly instructing the user to use Qrisp.interface.QiskitRuntimeBackend instead.

```
from qiskit_ibm_runtime.ibm_backend import IBMBackend as RealIBMBackend
if isinstance(backend, RealIBMBackend):
    print("Backend is a RealIBMBackend instance.")
else:
    print("Backend is NOT a RealIBMBackend instance.")
```

For real IBM backends, it yields:
> Backend is a RealIBMBackend instance.

But for fake IBM backends, it yields:

```
from qrisp import QuantumFloat
from qiskit_ibm_runtime.ibm_backend import IBMBackend as RealIBMBackend
from qiskit_ibm_runtime.fake_provider import FakeWashingtonV2
from qrisp.interface import QiskitBackend

backend=FakeWashingtonV2()

if isinstance(backend, RealIBMBackend):
    print("Backend is a RealIBMBackend instance.")
else:
    print("Backend is NOT a RealIBMBackend instance.")
```

> Backend is NOT a RealIBMBackend instance.

positr0nium:
Can you check if the latest commit now works with IBM hardware, @renezander90 ?

renezander90 (approved):
I checked that `QiskitBackend` rejects real IBM quantum backends; `QiskitRuntimeBackend` executes real IBM quantum backends.

PR #799, opened 2026-08-13
Bosonic operator

philipp-heinen:
## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- A new class `BosonicOperator` was implemented. This provides a conversion of bosonic operators to QubitOperators by means of three different encodings (standard binary, Gray code, one hot).
- Fermionic ladder operators were renamed from a and c to a_f and c_f, whereas the bosonic ones are now called a_b and c_b.

## How was it tested?
It was tested that the `.to_sparse_matrix()` gives back the correct matrix representation of the bosonic ladder operators; also a simulation of a Bose-Hubbard model both with the new module as well numerically were performed and compared.

## Checklist

## Reviewer Notes

The fermionic ladder operators a and c are now renamed in a_f and c_f but a and c are still provided as an alias, albeit with a deprecation notice, as discussed on Discord.

### DIFF
--- a/src/qrisp/operators/bosonic/__init__.py
+++ b/src/qrisp/operators/bosonic/__init__.py
@@ -1,0 +1,21 @@
+"""
+********************************************************************************
+* Copyright (c) 2024 the Qrisp authors
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+# -*- coding: utf-8 -*-
+from qrisp.operators.bosonic.bosonic_operator import *
+from qrisp.operators.bosonic.bosonic import *
--- a/src/qrisp/operators/bosonic/bosonic_operator.py
+++ b/src/qrisp/operators/bosonic/bosonic_operator.py
@@ -31,7 +31,7 @@ threshold = 1e-9
 
 
 class BosonicOperator(Hamiltonian):
-    r"""This class provides an efficient implementation of ladder term operators, i.e.,
+    r"""This class provides an efficient implementation of bosonic ladder term operators, i.e.,
     operators of the form
 
     .. math::
@@ -49,13 +49,13 @@ class BosonicOperator(Hamiltonian):
 
     Examples
     --------
-    A ladder term operator can be specified conveniently in terms of ``a`` (lowering, i.e., annihilation), ``c`` (raising, i.e., creation) operators:
+    A ladder term operator can be specified conveniently in terms of ``a_b`` (lowering, i.e., annihilation), ``c_b`` (raising, i.e., creation) operators:
 
     ::
         
-        from qrisp.operators.bosonic import a, c
+        from qrisp.operators.bosonic import a_b, c_b
 
-        O = a(2)*c(1)+a(3)*c(2)
+        O = a_b(2)*c_b(1)+a_b(3)*c_b(2)
         O
 
     Yields $a_2c_1+a_3c_2$.
@@ -95,7 +95,7 @@ class BosonicOperator(Hamiltonian):
 
             from qrisp.operators.bosonic import *
 
-            O = a(0)*a(1) - a(1)*a(0)
+            O = a_b(0)*a_b(1) - a_b(1)*a_b(0)
             print(O.reduce())
             # Yields: 2*a0*a1
 
@@ -103,7 +103,7 @@ class BosonicOperator(Hamiltonian):
         that has redundant terms, if hermitized.
 
 
-        >>> O = a(0)*a(1) + c(1)*c(0)
+        >>> O = a_b(0)*a_b(1) + c_b(1)*c_b(0)
         >>> reduced_O = O.reduce(assume_hermitian = True)
         >>> print(reduced_O)
         2*a0*a1
@@ -114,35 +114,12 @@ class BosonicOperator(Hamiltonian):
         1.0*a0*a1 + 1.0*c1*c0
 
         """
-        # This function performs some non trivial logic.
-
-        # The problem here is that each bosonic term can
-        # be reshaped into several different forms and still express
-        # the same operator.
-        # For instance, the operator can be arbitrarily reordered
-        # if the permutation sign is take care of and no creators/annihilators
-        # with the same index are swapped.
-        # Furthermore the Hamiltonian must be Hermitian, so each
-        # term must be equivalent to is Hermitian conjugate.
-
-        # This function implements a storage system, that combines
-        # the coefficients of differing terms representing the same operator
-        # into a single term.
-
-        # This dictionary will contain the new terms, where redundancies
-        # are taken care of.
+        # code is adapted from the FermionicOperator class
         new_terms_dict = {}
 
         for term, coeff in self.terms_dict.items():
-            # We only store the sorted version of each term.
-            # Sorting here means permuting the creators/annihilators
-            # while considering the sign of the permutation applied by the sort.
-            # The sort is performed in a stable manner, so terms like a(0)*c(0)
-            # don't get permuted (this would be a non-trivial anti-commutator).
             sorted_term = term.sort()
             if sorted_term not in new_terms_dict and assume_hermitian:
-                # If the sorted term is not in the terms dict, the sorted version
-                # of the daggering might be.
                 daggered_sorted_term = term.dagger().sort()
                 if daggered_sorted_term in new_terms_dict:
                     sorted_term = daggered_sorted_term
@@ -170,8 +147,8 @@ class BosonicOperator(Hamiltonian):
 
         Examples
         --------
-        >>> from qrisp.operators import a, c
-        >>> O = a(0)*c(1)+a(1)*c(0)+0.5*a(1)+0.5*c(1)
+        >>> from qrisp.operators import a_b, c_b
+        >>> O = a_b(0)*c_b(1)+a_b(1)*c_b(0)+0.5*a_b(1)+0.5*c_b(1)
         >>> O.coeffs()
         array([1. , 1. , 0.5, 0.5])
 
@@ -226,7 +203,7 @@ class BosonicOperator(Hamiltonian):
 
             from qrisp.operators import *
 
-            O = a(0)*c(1)*a(2) + a(3)
+            O = a_b(0)*c_b(1)*a_b(2) + a_b(3)
             print(O.dagger())
             # Yields: c2*a1*c0 + c3
 
@@ -252,7 +229,7 @@ class BosonicOperator(Hamiltonian):
 
             from qrisp.operators import *
 
-            O = a(0)*c(1)*a(2) + a(3)
+            O = a_b(0)*c_b(1)*a_b(2) + a_b(3)
             print(O.hermitize())
             # Yields: 0.5*a0*c1*a2 + 0.5*a3 + 0.5*c2*a1*c0 + 0.5*c3
 
@@ -516,7 +493,7 @@ class BosonicOperator(Hamiltonian):
             A sparse matrix representing the operator.
         truncation: How many bosonic occupation numbers to take into account.
         binary_encoding : string, optional
-            How to embed the bosonic terms into a QubitOperator.
+            How to embed the bosonic terms into a QubitOperator. Possible values are "gray_code", "standard_binary" and "one_hot".
 
         """
         return self.to_qubit_operator(truncation=truncation, binary_encoding=binary_encoding).to_sparse_matrix()
@@ -540,7 +517,7 @@ class BosonicOperator(Hamiltonian):
         truncation : int, optional
             How many bosonic occupation numbers to take into account
         binary_encoding : str, optional
-            The way of embedding the bosonic operator into qubits
+            How to embed the bosonic terms into a QubitOperator. Possible values are "gray_code", "standard_binary" and "one_hot".
 
         Returns
         -------
@@ -575,7 +552,7 @@ class BosonicOperator(Hamiltonian):
             This is because a quantum value would need to be copied for each sampling iteration, which is prohibited by the no-cloning theorem.
         truncation: How many bosonic occupation numbers to take into account.
         binary_encoding : string, optional
-            How to embed the bosonic terms into a QubitOperator.
+            How to embed the bosonic terms into a QubitOperator. Possible values are "gray_code", "standard_binary" and "one_hot".
         measurement_kwargs : dict, optional
             The keyword arguments of :meth:`QubitOperator.expectation_value <qrisp.operators.qubit.QubitOperator.expectation_value>`.
 
@@ -593,9 +570,7 @@ class BosonicOperator(Hamiltonian):
     #
 
     def trotterization(self, truncation=8, binary_encoding="gray_code", forward_evolution=True):
-        r"""Returns a function for performing Hamiltonian simulation, i.e., approximately implementing the unitary operator $U(t) = e^{-itH}$ via Trotterization.
-
-        """
+        r"""Returns a function for performing Hamiltonian simulation, i.e., approximately implementing the unitary operator $U(t) = e^{-itH}$ via Trotterization."""
         qubit_operator = self.to_qubit_operator(truncation=truncation, binary_encoding=binary_encoding)
         return qubit_operator.trotterization(forward_evolution=forward_evolution)
 
--- a/src/qrisp/operators/bosonic/bosonic_operator.py
+++ b/src/qrisp/operators/bosonic/bosonic_operator.py
@@ -1,0 +1,611 @@
+"""********************************************************************************
+* Copyright (c) 2024 the Qrisp authors
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+import numpy as np
+import sympy as sp
+
+from qrisp.operators import Hamiltonian
+from qrisp.operators.bosonic.bosonic_term import BosonicTerm
+from qrisp.operators.hamiltonian_tools import group_up_iterable
+from qrisp.operators.qubit import QubitOperator
+
+threshold = 1e-9
+
+#
+# BosonicOperator
+#
+
+
+class BosonicOperator(Hamiltonian):
+    r"""This class provides an efficient implementation of ladder term operators, i.e.,
+    operators of the form
+
+    .. math::
+        
+        O=\sum\limits_{j}\alpha_jO_j 
+            
+    where each term $O_j$ is a product of bosonic raising $a_i^{\dagger}$ and lowering $a_i$ operators acting on the $i$ th bosonic mode.
+
+    The ladder operators satisfy the commutation relations
+
+    .. math::
+
+        [a_i,a_j^{\dagger}] &= a_ia_j^{\dagger}+a_j^{\dagger}a_i = \delta_{ij}\\
+        [a_i^{\dagger},a_j^{\dagger}] &= [a_i,a_j] = 0
+
+    Examples
+    --------
+    A ladder term operator can be specified conveniently in terms of ``a`` (lowering, i.e., annihilation), ``c`` (raising, i.e., creation) operators:
+
+    ::
+        
+        from qrisp.operators.bosonic import a, c
+
+        O = a(2)*c(1)+a(3)*c(2)
+        O
+
+    Yields $a_2c_1+a_3c_2$.
+
+    """
+
+    def __init__(self, terms_dict={}):
+
+        self.terms_dict = dict(terms_dict)
+
+    def reduce(self, assume_hermitian=False):
+        """Applies the bosonic commutation laws to bring the operator into
+        a standard form. This can reduce the amount of terms because several
+        terms might be the permuted version of each other and therefore their
+        coefficients add up.
+
+        This function can reduce the amount of terms even further if the user
+        can guarantee that the operator will be hermitized. In this case more
+        identifications can be made.
+
+        Parameters
+        ----------
+        assume_hermitian : bool, optional
+            If set to True the function will assume that the result will be
+            hermitized. The default is False.
+
+        Returns
+        -------
+        BosonicOperator
+            The reduced BosonicOperator.
+
+        Examples
+        --------
+        We create a BosonicOperator with redundant term definitions:
+
+        ::
+
+            from qrisp.operators.bosonic import *
+
+            O = a(0)*a(1) - a(1)*a(0)
+            print(O.reduce())
+            # Yields: 2*a0*a1
+
+        To demonstrate the ``assume_hermitian`` feature, we create a BosonicOperator
+        that has redundant terms, if hermitized.
+
+
+        >>> O = a(0)*a(1) + c(1)*c(0)
+        >>> reduced_O = O.reduce(assume_hermitian = True)
+        >>> print(reduced_O)
+        2*a0*a1
+
+        Hermitizing gives the original operator.
+
+        >>> print(reduced_O.hermitize())
+        1.0*a0*a1 + 1.0*c1*c0
+
+        """
+        # This function performs some non trivial logic.
+
+        # The problem here is that each bosonic term can
+        # be reshaped into several different forms and still express
+        # the same operator.
+        # For instance, the operator can be arbitrarily reordered
+        # if the permutation sign is take care of and no creators/annihilators
+        # with the same index are swapped.
+        # Furthermore the Hamiltonian must be Hermitian, so each
+        # term must be equivalent to is Hermitian conjugate.
+
+        # This function implements a storage system, that combines
+        # the coefficients of differing terms representing the same operator
+        # into a single term.
+
+        # This dictionary will contain the new terms, where redundancies
+        # are taken care of.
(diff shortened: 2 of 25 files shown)


src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,273 @@
...
+                temp += block1
+            else:
+                temp += block2
+        code.append(temp)
+
+    return np.transpose(np.asarray(code)).tolist()
+
+
+# Return standard binary
+def standard_binary(n):
```
renezander90: Gemini proposes: 

```
from itertools import product

def standard_binary(n):
    # product returns tuples like (0, 1, 0), so we convert them to lists
    return [list(bits) for bits in product((0, 1), repeat=n)]
```

Image

renezander90:
We aim to introduce typing across the Qrisp codebase. While the ``QubitOperator`` and ``FermionicOperator`` do not have this yet, it is preferred that new code contributions include type hints.

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: I've integrated this suggestion.

philipp-heinen:
Added type hints.

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+
+threshold = 1e-9
+
+#
+# BosonicOperator
+#
+
+
+class BosonicOperator(Hamiltonian):
+    r"""This class provides an efficient implementation of bosonic ladder term operators, i.e.,
```
positr0nium, line 36: Having some references here to read about encodings, typical examples etc. would be nice

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+        terms might be the permuted version of each other and therefore their
+        coefficients add up.
+
+        This function can reduce the amount of terms even further if the user
+        can guarantee that the operator will be hermitized. In this case more
+        identifications can be made.
+
+        Parameters
+        ----------
+        assume_hermitian : bool, optional
```
positr0nium, line 129: A better design would be to automatically check if the operator is hermitian and apply the corresponding optimization in this case. A note in the docs would mention that hermitizing before gives stronger optimizations. This way the user can't write code that silently breaks.

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+                if coeff == 0:
+                    del new_terms_dict[term]
+
+        return BosonicOperator(new_terms_dict)
+
+    def len(self):
+        return len(self.terms_dict)
+
+    def coeffs(self):
+        """Returns the coefficients of the operator.
```
positr0nium, line 189: This needs info about the order of the coefficients, such that users can associate them to the terms

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+        ----------
+        truncation : int, optional
+            How many bosonic occupation numbers to take into account
+        binary_encoding : str, optional
+            How to embed the bosonic terms into a QubitOperator. Possible values are "gray_code", "standard_binary" and "one_hot".
+
+        Returns
+        -------
+        O : :ref:`QubitOperator`
+            The resulting QubitOperator.
```
positr0nium, line 597: Would be nice to have an example here that is explaining the logic behind it and also refering to the paper

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+        truncation: How many bosonic occupation numbers to take into account.
+        binary_encoding : string, optional
+            How to embed the bosonic terms into a QubitOperator. Possible values are "gray_code", "standard_binary" and "one_hot".
+        measurement_kwargs : dict, optional
+            The keyword arguments of :meth:`QubitOperator.expectation_value <qrisp.operators.qubit.QubitOperator.expectation_value>`.
+
+        Returns
+        -------
+        callable
+            A function returning an array containing the expectation value.
```
positr0nium, line 656: Example please 😇

src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,274 @@
...
+            1) Raising operators preceed lowering operators
+            2) Operators are ordered in descending order of bosonic modes
+
+        Example: a_5^dagger a_2^dagger a_3 a_1
+
+        """
+        pass
+
+    def sort(self):
+        # Sort ladder operators (ladder operator semantics are order in-dependent)
```
positr0nium:
```suggestion
        # Sort ladder operators (ladder operator semantics are order independent)
```

src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,274 @@
...
+        sorting_list = [-index for index, is_creator in self.ladder_list]
+        perm = np.argsort(sorting_list, kind="stable")
+        ladder_list = [self.ladder_list[i] for i in perm]
+
+        return BosonicTerm(ladder_list)
+
+    def bosonic_swap(self, permutation: list):
```
positr0nium: Some of the methods below are only needed for hamiltonian simulation of ``FermionicOperators`` (I think). We should remove them if they are indeed called nowhere.

src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,274 @@
...
+        Apart from the Gray encoding, a standard binary and a one-hot encoding are provided.
+        https://www.nature.com/articles/s41534-020-0278-0 contains more details on their respective advantages and disadvantages.
+
+        Parameters
+        ----------
+        truncation: int, optional
+            The number of bosonic occupation numbers one wants to describe.
+            Note that 0 also counts as an occupation number, so occupation numbers from 0 to truncation-1 are represented.
+        binary_encoding: str, optional
+            How to embed the bosonic matrix into qubits. Possible values are "gray_code", "standard_binary" and "one_hot".
```
positr0nium: Returns section is missing. This docstring (but also the other encoding related docstring) should contain more info about the returned shape. I would assume that a one-hot encoded term requires more qubits than binary. This is currently still a bit opaque.

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+        
+        O=\sum\limits_{j}\alpha_jO_j 
+            
+    where each term $O_j$ is a product of bosonic raising $a_i^{\dagger}$ and lowering $a_i$ operators acting on the $i$ th bosonic mode.
+
+    The ladder operators satisfy the commutation relations
+
+    .. math::
+
+        [a_i,a_j^{\dagger}] &= a_ia_j^{\dagger}+a_j^{\dagger}a_i = \delta_{ij}\\
```
positr0nium:
```suggestion
        [a_i,a_j^{\dagger}] &= a_ia_j^{\dagger}-a_j^{\dagger}a_i = \delta_{ij}\\
```

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+
+        """
+        qubit_operator = self.to_qubit_operator(truncation=truncation, binary_encoding=binary_encoding)
+        return qubit_operator.expectation_value(state_prep, **measurement_kwargs)
+
+    #
+    # Trotterization
+    #
+
+    def trotterization(self, truncation: int = 8, binary_encoding: str = "gray_code", forward_evolution: bool = True):
```
positr0nium, line 680: Better docstring with examples, please :)

src/qrisp/operators/fermionic/fermionic.py
```diff
@@ -15,19 +15,38 @@
...
 
-def c(arg):
+def c_f(arg: int):
     if isinstance(arg, int):
         return FermionicOperator({FermionicTerm([(arg, True)]): 1})
     else:
         raise Exception("Cannot initialize operator from type " + str(type(arg)))
+
+
+def a(arg: int):
```
positr0nium, line 37: The ``a``/``c`` naming convention is till found in a lot of places throughout the project. To fix this the testing pipeline should pass and the documentation might need to be scanned (with Claude?)

src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,274 @@
...
+import numpy as np
+import warnings
+from typing import Self
+
+from qrisp.operators.bosonic.visualization import a_, c_
+from qrisp.operators.qubit import A, C, Z, P0, P1
+
+
+class BosonicTerm:
+    r""" """
```
positr0nium: Needs a docstring

tests/operators_tests/test_bosonic_term.py
```diff
@@ -0,0 +1,45 @@
...
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
+from qrisp import *
+
+
+def test_fermionic_term():
```
positr0nium:
```suggestion
def test_bosonic_term():
```

tests/operators_tests/test_bosonic_term.py
```diff
@@ -0,0 +1,45 @@
...
+    from qrisp.operators.bosonic import a_b as a, c_b as c
```
positr0nium: This will break once the deprecation cycle is over

tests/operators_tests/test_bosonic_to_qubit.py
```diff
@@ -0,0 +1,86 @@
...
+*
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
```
positr0nium, line 16: Would be nice to have some end-to-end test

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,590 @@
...
+                del res_terms_dict[ladder_term]
+
+        for ladder_term, coeff in other.terms_dict.items():
+            res_terms_dict[ladder_term] = res_terms_dict.get(ladder_term, 0) + coeff
+            if abs(res_terms_dict[ladder_term]) < threshold:
+                del res_terms_dict[ladder_term]
+
+        result = BosonicOperator(res_terms_dict)
+        return result
+
```
positr0nium, line 344: Would be nice have a power method (example: ``a_b(0)**5``)

src/qrisp/operators/bosonic/bosonic_term.py
```diff
@@ -0,0 +1,274 @@
...
+                if index_list[i] == index_list[i + 1]:
+                    index_list.pop(i)
+                    index_list.pop(i)
+                    continue
+                i += 1
+
+            self.unipolars = index_list[::-1]
+            return list(self.unipolars)
+
+    def to_qubit_term(self, truncation: int = 8, binary_encoding: str = "gray_code"):
```
positr0nium, line 113: There is a hidden layer of complexity here. The intra-mode order of a term can influence various quantities because currently the truncation is applied before multiplying terms together. Intuitively speaking, if you have 3 creation operators and 3 annihilation operators and you truncate with 2, "walking through them" ends you (temporarily) in a state that is outside of the truncation (|3>), even though in reality you are still "walking back" with the annihilation operators. This introduces another type of error.

```python
from qrisp.operators.bosonic import a_b, c_b

O = a_b(0)*a_b(0)*c_b(0)*c_b(0)
print(O.to_sparse_matrix(truncation = 4))
#   (0, 0)        (2.0000000000000004+0j)
#   (1, 1)        (6.000000000000001+0j)
print(O.to_sparse_matrix(truncation = 2))
#   (0, 0)        0.0

# The (1,1) entry is missing even though it would
# fit into the truncated matrix

# Applying commutation rules to make sure creators
# and annihilators only appear in alternating sequences
# resolves the problem but introduces additional
# terms.
O = c_b(0)*a_b(0)*c_b(0)*a_b(0) + 3*c_b(0)*a_b(0) + 2
print(O.to_sparse_matrix(truncation = 2))
#   (0, 0)        (2+0j)
#   (1, 1)        (6+0j)
```

Fortunately there is a more elegant fix that doesn't require additional resources: You temporarily build the qubit operator with a conservative truncation threshold that guarantees that you will not "walk outside", and only apply the truncation afterwards (i.e. on the finished ``QubitTerm``) by removing all factors in the ``QubitTerm`` that go beyond the truncation.

This guarantees that the users truncation demand is still satisfied but you don't have this "temporarily walking outside" issue.

positr0nium: @renezander90

positr0nium (changes requested):

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: Yes, this is a good point. You can't get "right" the commutation relation of the bosonic operators with finite matrices, because for a finite matrix $a$ we have $\mathrm{Tr}(aa^\dagger-a^\dagger a) = 0 \neq \mathrm{Tr}(\mathbb{1})$, so there is necessarily some ambiguity. So if I understand correctly what you are suggesting means that for two bosonic operators that according to the commutation relations are equivalent, the same truncated qubit operator should be output, namely the one that comes out when you do a normal ordering of the bosonic operator before truncation. This makes a lot of sense actually and is probably the most user-friendly behavior, I agree. 
The way you suggest to achieve this by temporarily increasing the truncation is a good idea and it's the easiest to implement. However, I am wondering if it would not be the most elegant solution to implement some method of the BosonicOperator class that brings a bosonic operator into normal ordered form, with all the creators left to the annihilators. One would then just have to apply this method before converting to a qubit operator in the normal way. I didn't implement a normal ordering method so far because it is also absent from the fermionic operator, but actually it might be useful also for other purposes.

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Yes, I implemented it. Wouldn't it be nice to have this also for the `FermionicOperator`? Of course, for a single ladder operator it doesn't make much sense because e.g. `a_f(0)**2` is just zero, but e.g. for something like `(a_f(0)+a_f(1)+a_f(2))**3` it would make sense.

src/qrisp/operators/bosonic/bosonic_term.py
positr0nium: There might be a missunderstanding here: I'm not proposing any reordering based approached. Here is a rephrased version of my proposal.

We alter the way that the ``to_qubit_term`` method works: If the user requires truncation ``N``, we temporarily build the ``QubitOperator`` with the currently implemented function using truncation ``N+K``. We then "cut off" all factors from the ``QubitOperator`` that are ``>N``.
By choosing ``K`` appropriately, we avoid the "temporarily walking outside" scenario. But how big do we have to choose ``K``?. By counting the amount of raising operators per mode, add one and take the maximum over all modes. If there are at most ``J`` raising operators in a mode and we choose ``K = J+1``, even if you start in the highest state of the ladder (``|N>``), you are not temporarily walking out in the ``N+K`` sized ``QubitOperator`` because the ladder state (``|N+K>``) has not been truncated.

philipp-heinen: Sorry, my comment was a bit unclear. I understand your proposal, which I find a very good idea. I was only suggesting a further alternative, which should have the same effect I think: With the commutation relation, one brings the operator to a form where all c's are to the left and all a's to the right. Then this problem of walking out of the truncation space does not occur, or at least only as far as it is unavoidable, because all the operators that make you go "down" come first and only then the operators that move you "up".
I'm not sure what is the better approach. For the latter, one would have to invest a bit of work to write a method for this normal ordering, but it could be useful also for other purposes, so it might be worth investing it, and then it is just a one-liner to avoid the running out of truncation space. Your proposal is more straightforward to implement in contrast.

positr0nium: The problem with your approach is that applying the commutation relations inserts additional terms, that might give an overhead in hamiltonian simulation and expectation value resource requirements. It might however also be the case for the ``N+K``-strategy and that both strategies are indeed equivalent in additional resources.

positr0nium: The normal ordering might be more demanding in terms of classical ressources though (performing the ordering, applying the commutators etc.)

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: This I took over from the `FermionicOperator` one to one (only with the necessary adaptation of the commutation relation). As far as I can see, how it is implemented, if the operator happens to be hermitian already, then the optimization is applied automatically, i.e. the assume_hermitian flag doesn't have any effect in this case. Setting this flag rather means that in case the operator is not hermitian, it is hermitianized before doing the optimization.

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: Done

tests/operators_tests/test_bosonic_term.py
philipp-heinen: Why? It imports the operators a_b and a_c, just renames them for simplicity.

tests/operators_tests/test_bosonic_to_qubit.py
philipp-heinen: I've implemented an additional test that checks for the expectation value of the number operator in Fock states.

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: I've implemented your suggestion now, it was very straightforward. Taking the maximum over all modes for determining k was not necessary, as the implementation handles the modes independently already, so it is taken for every single mode.

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Do you mean that the function also returns the terms that the coefficients belong to? E.g. 
```Python
return np.array(list(self.terms_dict.values())), list(self.terms_dict.keys())
```
Or should the coefficients be returned in some order (which is described in the docstring)? One could order the terms in many different ways. One possibility is e.g. to sort by the smallest mode present in the term, if two terms are equal in this by the second-smallest mode and so on.

I copied this function one to one from `FermionicOperator`.

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: I could open another PR that adds this also to `FermionicOperator`.

tests/operators_tests/test_bosonic_term.py
renezander90: Yes, the imports work, but should be module-level, and not inside the function.

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_term.py
philipp-heinen: Done

tests/operators_tests/test_bosonic_term.py
philipp-heinen: Done

src/qrisp/operators/fermionic/fermionic.py
renezander90: Yes, but I still find using `a_f`, `c_f` a bit clunky. I think it's best to only adjust the imports, i.e.,

`from qrisp.operators.fermionic import a_f as a`

Then the rest of the code can stay the same.

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,758 @@
...
+    def __mul__(self, other: Self):
```
renezander90:
```suggestion
    def __mul__(self, other: int | float | complex | Self) -> Self:
```

Similarly, for other arithmetic methods.

philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
positr0nium: ``a_f(k)**n = 0`` for any ``n >1``

philipp-heinen: Yes, I was thinking of exponentiating sums of operators such as `(a_f(0)+a_f(1)+a_f(2))**3` or `(a_f(0)+c_f(0))**2` which doesn't evaluate to 0. But probably that doesn't have too much of a use case.

src/qrisp/operators/fermionic/fermionic.py
philipp-heinen: I've implemented this change in the tests and the documentation now.

philipp-heinen:
The ruff lint test fails only because of the files in the fermionic module. There, I replaced only a, c with a_f, c_f but didn't touch anything else, so it gets tested because of the change but it fails because it doesn't fully comply with the standards. If you like, I could fix the issues in the fermionic module to make this pass, I did this for the bosonic one before. It is mostly missing docstrings, missing blank lines, a missing \_\_hash\_\_ method and the like. Just let me know.

renezander90:
> The ruff lint test fails only because of the files in the fermionic module. There, I replaced only a, c with a_f, c_f but didn't touch anything else, so it gets tested because of the change but it fails because it doesn't fully comply with the standards. If you like, I could fix the issues in the fermionic module to make this pass, I did this for the bosonic one before. It is mostly missing docstrings, missing blank lines, a missing __hash__ method and the like. Just let me know.

Thanks! We should keep this PR focussed. More changes in `FermionicOperator` should be in a separate PR.
The ruff checks are non-blocking.

src/qrisp/operators/bosonic/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""
```
renezander90, line 1: Please adapt the new comment format for the license headers (you find it any other file on Qrisp's main branch).

renezander90, line 1: Each new file also needs a short module-level docstring after the comment license header.

philipp-heinen: Done

philipp-heinen: Done

philipp-heinen:
ruff check wasn't happy about the changes, it complains about the format of the license header and that imports should be at the very top (instead of the module level docstring). But from your previous comment I think we can safely ignore this ;)

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,847 @@
...
+                if daggered_sorted_term in new_terms_dict:
+                    sorted_term = daggered_sorted_term
+
+            # Compute the new coefficient.
+            new_terms_dict[sorted_term] = coeff + new_terms_dict.get(sorted_term, 0)
+
+        for term, coeff in list(new_terms_dict.items()):
+            if isinstance(coeff, (int, float)):
+                if coeff == 0:
+                    del new_terms_dict[term]
```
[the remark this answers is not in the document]
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,847 @@
...
+    def __eq__(self, other: Self):
+        """Check the equality of two operators."""
+        reduced_self = self.reduce()
+        reduced_other = other.reduce()
+
+        if len(reduced_self.terms_dict) != len(reduced_other.terms_dict):
+            return False
+
+        for term, coeff in reduced_self.terms_dict.items():
+            if term not in other.terms_dict:
```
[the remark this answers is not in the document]
philipp-heinen: Done

src/qrisp/operators/bosonic/bosonic_operator.py
```diff
@@ -0,0 +1,847 @@
...
+            raise TypeError("Cannot multipliy BosonicOperator and " + str(type(other)))
+
+        res_terms_dict = {}
+
+        for ladder_term1, coeff1 in self.terms_dict.items():
+            for ladder_term2, coeff2 in other.terms_dict.items():
+                curr_ladder_term = ladder_term1 * ladder_term2
+                res_terms_dict[curr_ladder_term] = res_terms_dict.get(curr_ladder_term, 0) + coeff1 * coeff2
+
+        self.terms_dict = res_terms_dict
```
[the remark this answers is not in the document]
philipp-heinen: Done

src/qrisp/operators/fermionic/fermionic.py
```diff
@@ -15,19 +15,38 @@
...
+        "Using 'a' for the fermionic annihilation operator is deprecated; use 'a_f' instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return a_f(arg)
+
+
+def c(arg: int):
+    warnings.warn(
+        "Using 'c' for the fermionic annihilation operator is deprecated; use 'c_f' instead.",
```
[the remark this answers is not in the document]
philipp-heinen: Done

philipp-heinen:
The bugs that CodeRabbit reports for the `bosonic_operator` module actually equally concern the fermionic module, as the code was just adapted from there. Might be worth fixing them also there.

PR #800, opened 2026-08-16
Fix jasp assembly format

### LINKED ISSUE #783: Inconsistent custom assembly (create_qubits) between xdsl and MLIR
### Priority

Medium

### Responsible

_No response_

### Qrisp Version

0.9.6

### Python Version

3.14

### Operating System

Linux

### Short Description

This diff summarizes it:

```diff
-    %1, %2 = jasp.create_qubits %0, %arg0 : !jasp.QuantumState, tensor<i64> -> !jasp.QubitArray, !jasp.QuantumState
+    %1, %2 = jasp.create_qubits %0, %arg0 : tensor<i64>, !jasp.QuantumState -> !jasp.QubitArray, !jasp.QuantumState
```

The upper line is what is expected by the tablegen. The lower line shows that is emitted by xdsl. A recent fix of the custom assembly format (https://github.com/eclipse-qrisp/Qrisp/commit/d63b9ce09f0be1fb9b8ea4f08ec9e8e140dd1421) has not been applied to the tablegen files.

### Steps to Reproduce

See source files.

### Expected Behavior

Should be consistent.

### Actual Behavior

Not consistent.

### Additional Context

We observed this while trying to update our qrisp integration tests from version `0.9.5` to `0.9.6` in the [qcc project](https://github.com/FullStaQD/compiler) (fullstaqd). Over there we have to duplicate the tablegen files of this repo in order to parse the mlir output of qrisp. The issue is that xdsl is responsible for printing hence the inconsistency surfaces. It is not urgent for us since we now switched to letting qrisp generate the generic assembly format, fixing the problem for us. Might still be nice to fix the inconsistency over here though.

### Definition of Done

- [ ] Root cause identified
- [ ] Regression test added
- [ ] Changelog entry added

positr0nium:
## Description

The custom assembly format of the Jasp dialect is declared twice: by
`assembly_format` in `xdsl_dialect.py`, which is what Qrisp prints with, and by
`assemblyFormat` in `dialect_definition/JaspOps.td`, which is what an MLIR-based
consumer builds its parser from. The two had drifted apart for
`jasp.create_qubits`, so MLIR emitted by `to_mlir()` could not be parsed by a
dialect built from the shipped TableGen files.

This PR fixes that, fixes a second printer/parser mismatch found in the same
place, and adds round-trip tests so the two definitions cannot silently diverge
again.

## Related Issues

Closes #783

Reported by the [qcc / FullStaQD compiler](https://github.com/FullStaQD/compiler)
project, which duplicates the Jasp TableGen files to parse Qrisp's MLIR output.

## Type of Change

- [x] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

Not breaking, but two textual details of the emitted MLIR change, which matters
for anyone diffing against stored output:

- `jasp.quantum_gate` and `jasp.parity` lose the stray whitespace their
  hand-written printers emitted: `"h" (%3) , %1` is now `"h"(%3), %1`, which is
  what MLIR itself would print.
- A parser generated from the previous `JaspOps.td` expected the operand types
  of `create_qubits` in the other order. That parser could not read Qrisp's
  output before this PR either, so nothing that worked stops working.

## What was changed?

- **`JaspOps.td`** — `create_qubits` now lists its operand types in operand
  order, `type($amount), type($qst_in)`, matching what Qrisp prints. This is the
  fix for #783; the corresponding change to `xdsl_dialect.py` had already been
  made in d63b9ce but was never applied to the TableGen.
- **`xdsl_dialect.py`** — `jasp.quantum_gate` and `jasp.parity` had a
  hand-written `print()` and no matching parser, so a printed module could not
  be read back at all. Both now use the same declarative `assembly_format` as
  the other ten operations, copied from their TableGen counterparts. This also
  stops their printers from silently dropping attributes, since the attribute
  dictionary was hardcoded.
- **`xdsl_dialect.py`** — renamed the variadic operand of `quantum_gate` from
  `gate_params` to `gate_operands`, the name already used by the TableGen and
  the generated bindings. It holds the qubits the gate acts on followed by any
  float parameters, so `params` was misleading — for `h` or `cx` it contains no
  parameters at all.
- **`tests/jax_tests/test_mlir_roundtrip.py`** (new) — round-trip and
  consistency tests, see below.
- **`tests/jax_tests/jasp_dialect_syntax.mlir`** (new) — a reference module with
  one instance of every operation in custom syntax. It pins the textual contract
  in one reviewable place and is written so that a consumer who has built the
  dialect from the TableGen files can check its own parser against it.
- **`_ods_common.py`** — unrelated to #783, but found while regenerating the
  bindings: `mlir-tblgen` from LLVM 22 emits a call to
  `_cext.globals.register_traceback_file_exclusion()` that jaxlib 0.7.1's MLIR
  runtime does not provide, so a freshly generated `_jasp_ops_gen.py` fails to
  import. `_ods_common` now wraps jaxlib's C extension and supplies that API as
  a no-op (it only affects traceback prettification). The wrapper is skipped
  entirely once jaxlib provides the call.
- **Documentation** — a "Textual Formats and Interoperability" section in
  `MLIR Interface.rst` describing the custom and the generic form, and pointing
  consumers at the TableGen files and the reference module; plus changelog
  entries.

The generated bindings themselves are **not** part of this PR. Generating them
from the pre-fix and post-fix TableGen produces byte-identical output —
`assemblyFormat` is compiled into a C++ parser/printer, which the Python
bindings do not contain — so no regeneration is required for this change.

## How was it tested?

Issue #783 does not define Test-IDs, so the tests are listed by name. All were
confirmed to fail before the fix and pass after it.

| Test-ID | Status |
|---------|--------|
| `test_tablegen_declares_the_same_assembly_format` — every xDSL assembly format occurs verbatim in `JaspOps.td` | ✅ |
| `test_syntax_reference_roundtrips` — the reference module parses, verifies and re-prints as a fixed point | ✅ |
| `test_syntax_reference_covers_every_op` — every op of the dialect appears in the reference module | ✅ |
| `test_emitted_mlir_roundtrips[False]` / `[True]` — MLIR emitted by `to_mlir()` parses back, both `lower_stablehlo` settings | ✅ |
| Existing `test_mlir.py` and `test_mlir_rewrites.py` (57 tests total with the above) | ✅ |
| `ruff format --check` | ✅ |

Additionally, outside the test suite:

- The TableGen files were validated with real `mlir-tblgen` (LLVM 22):
  `-gen-op-defs`, `-gen-op-decls`, `-gen-dialect-defs` and `-gen-typedef-defs`
  all succeed. `-gen-op-defs` compiles `assemblyFormat` into a C++
  parser/printer, so the corrected format is well-formed, not merely plausible.
- Reverting only the `.td` line reproduces #783 as a test failure naming
  `jasp.create_qubits`, confirming the check is a real regression test.

Not covered: no test feeds `jasp_dialect_syntax.mlir` through `mlir-opt`. That
requires an out-of-tree dialect built from the TableGen files, which does not
exist in this repository — see Reviewer Notes.

The inconsistency reported in #783, `.td` versus what Qrisp prints:

```diff
-    %1, %2 = jasp.create_qubits %0, %arg0 : !jasp.QuantumState, tensor<i64> -> !jasp.QubitArray, !jasp.QuantumState
+    %1, %2 = jasp.create_qubits %0, %arg0 : tensor<i64>, !jasp.QuantumState -> !jasp.QubitArray, !jasp.QuantumState
```

The second defect, reading back a module Qrisp had just printed (before):

```
xdsl.utils.exceptions.ParseError: <unknown>:6:28
    %4 = jasp.quantum_gate "h" (%3) , %1 : (!jasp.Qubit) , !jasp.QuantumState -> !jasp.QuantumState
                           ^^^
                           Operation jasp.quantum_gate does not have a custom format.
```

And what the consistency test reports when the two definitions disagree:

```
AssertionError: assembly format of ['jasp.create_qubits'] not found in JaspOps.td. xDSL prints:
  jasp.create_qubits: $amount attr-dict `,` $qst_in `:` type($amount) `,` type($qst_in) `->` ...
```

## Checklist

## Reviewer Notes

- **The `_ods_common.py` shim is independent of #783** and can be split into its
  own PR if you would rather keep this one focused. It is inert against the
  currently checked-in bindings, which never make the call it patches; it exists
  so that the next regeneration with a current LLVM produces a file that
  imports.
- **Why the TableGen check is a substring test rather than a parser.** It
  normalises whitespace and asserts each xDSL format string occurs in
  `JaspOps.td`. Parsing TableGen properly would mean maintaining a parser for it
  in the test suite; the only assumption made here is that format strings are
  written on one line. Type constraints are deliberately not compared: `_TensorInt`
  and the unconstrained `gate_operands` are looser on the xDSL side on purpose,
  so an exact comparison would need a table of per-op exceptions, which is where
  the next drift would hide.


renezander90:
Please fix the new documentation warnings:
```
/Qrisp/documentation/source/reference/Jasp/MLIR Interface.rst:274: WARNING: Lexing literal_block '%0, %1 = jasp.create_qubits %arg0, %arg1 : tensor<i64>, !jasp.QuantumState -> !jasp.QubitArray, !jasp.QuantumState' as "python" resulted in an error at token: '!'. Retrying in relaxed mode. [misc.highlighting_failure]
/Qrisp/documentation/source/reference/Jasp/MLIR Interface.rst:293: WARNING: Lexing literal_block '%0, %1 = "jasp.create_qubits"(%arg0, %arg1) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)' as "python" resulted inan error at token: '!'. Retrying in relaxed mode. [misc.highlighting_failure]
```

renezander90:
The example for ``Jaspr.to_mlir``(centerclass.py) uses `print(jaspr.to_mlir())` which yields custom assembly format:

``%0, %1 = jasp.create_qubits %arg13, %arg14 : tensor<i64>, !jasp.QuantumState -> !jasp.QubitArray, !jasp.QuantumState``

But the output cell shows generic format:

``%0, %1 = "jasp.create_qubits"(%arg0, %arg1) : (tensor<i64>, !jasp.QuantumState) -> (!jasp.QubitArray, !jasp.QuantumState)``

tests/jax_tests/jasp_dialect_syntax.mlir
```diff
@@ -0,0 +1,44 @@
+// Reference module exercising the custom assembly format of every operation in
```
renezander90: Needs a license header

PR #802, opened 2026-08-18
Add modulus to traced attributes

### LINKED ISSUE #801: QuantumModulus with a traced modulus leaks a JAX tracer across independent jaspify calls via qache
**Priority:** High

**Responsible:** René Zander

**Qrisp Version:** 0.9.7

**Python Version:** 3.11.15

**Operating System:** macOS

**Short Description:**
Calling the same `@jaspify`-decorated function twice in one process raises `jax.errors.UnexpectedTracerError` on the second call, whenever the function builds a `QuantumModulus(q)` from a modulus `q` that is itself a traced argument, and any `@qache`-cached subroutine touches that variable. The first call always succeeds; only the second (and later) independent calls fail.

**Steps to Reproduce:**
1. Define a `@qache`-decorated helper that operates on a `QuantumModulus`.
2. Define a `@jaspify` function that constructs `QuantumArray(QuantumModulus(q), ...)` where `q` is one of the function's own arguments (not a hardcoded literal), and calls the helper.
3. Call the `@jaspify` function twice with identical arguments in the same process.

```python
from qrisp import QuantumArray, QuantumModulus, jaspify, measure
from qrisp.jasp import qache

@qache
def op(f, root):
    f[0] += root

@jaspify
def solve(q, root):
    f = QuantumArray(QuantumModulus(q), shape=(1,))
    op(f, root)
    return measure(f)

print(solve(13, 5))  # ok
print(solve(13, 5))  # raises jax.errors.UnexpectedTracerError
```

**Expected Behavior:**
Both calls to `solve(13, 5)` should succeed and return identical results, since they are independent, functionally equivalent invocations with identical arguments.

**Actual Behavior:**
The first call succeeds (`[0.]`). The second call raises:

```
jax.errors.UnexpectedTracerError: Encountered an unexpected tracer. A function
transformed by JAX had a side effect, allowing for a reference to an
intermediate value with type int64[] wrapped in a DynamicJaxprTracer to escape
the scope of the transformation.
...
The function being traced when the value leaked was amended_function at
src/qrisp/jasp/jasp_expression/centerclass.py:1612 traced for jit.
The leaked intermediate value was created on line
src/qrisp/jasp/jasp_expression/centerclass.py:1468 (make_jaxpr_mod.<locals>.jaxpr_creator).
```

**Root Cause Analysis:**

`QuantumModulus.__init__` (quantum_modulus.py) sets `self.modulus = modulus` but never registers `"modulus"` in `self.traced_attributes` or `self.static_attributes` (it inherits `QuantumFloat`'s `traced_attributes = ["exponent"]`, `static_attributes = ["signed"]` unchanged). A dynamic/traced `modulus` is nonetheless an intentionally supported feature — `smallest_power_of_two()` (jasp_mod_tools.py) has an explicit branch for a "traced JAX scalar" modulus.

`flatten_qv()` (qv_flattening.py) builds the pytree's `aux_data` as `(QuantumVariableTemplate(qv, False), *static_attrs)`. `QuantumVariableTemplate.__init__` does `self.qv = copy.copy(qv)` — a shallow copy of the *entire* QuantumVariable, including `modulus` (since it isn't stripped out via `traced_attributes`). When `modulus` is a live JAX tracer, it gets embedded inside `aux_data`, which JAX's pytree contract requires to be static/hashable/trace-independent — this is never validated, so it's silently accepted.

`qache` (qaching.py) wraps functions in `jax.jit`, whose compilation cache is **process-global**, persisting across separate, independent top-level `jaspify()`/`make_jaspr()` calls. Because the cache key/structure includes the tainted `aux_data`, the tracer captured during the first call's (already-concluded) trace is reused/touched during the second, independent call — at which point JAX's tracer-liveness check raises `UnexpectedTracerError`.

Confirmed via elimination:
- Reproduces with plain `@qache` alone; `custom_inversion`/`conjugate` are **not** required (initially suspected but ruled out).
- Does **not** reproduce with `QuantumFloat` (whose dynamic `exponent` attribute *is* correctly listed in `traced_attributes`).
- Does **not** reproduce if `q` is a hardcoded Python literal (e.g. `QuantumModulus(13)`) instead of a traced function argument.
- Does **not** reproduce without any `@qache`-wrapped subroutine touching the QuantumModulus.

**Means of Verification:**

| Test-ID | Type     | Description | Status |
|---------|----------|-------------|--------|
| T-001   | Positive| Traced modulus (`q` as a `@jaspify` function argument) inside `QuantumArray(QuantumModulus(q), ...)`, with a `@qache`-cached subroutine, called repeatedly (previously raised `UnexpectedTracerError` on 2nd call) | ✅ |
| T-002   | Positive | Same traced-modulus scenario as T-001, but going through `custom_inversion`/`conjugate` (original notebook pattern) | ✅ |
| T-003   | Positive | Dynamic, hardcoded modulus (`QuantumModulus(jnp.array(13))`) inside `@jaspify`, called repeatedly | ✅ |
| T-004   | Positive | Static, hardcoded modulus (`QuantumModulus(13)`) inside `@jaspify`, called repeatedly | ✅ |

**Definition of Done:**
- [x] Root cause identified
- [x] Regression test added
- [x] Changelog entry added

renezander90:
## Description

Fixes a bug where `QuantumModulus` constructed with a traced (Jasp-dynamic) modulus leaked a stale JAX tracer across independent `jaspify`/`make_jaspr` calls, causing `jax.errors.UnexpectedTracerError` on the second and subsequent calls to a function that touched the affected `QuantumModulus` inside a `@qache`-cached subroutine (including via `custom_inversion`/`conjugate`, which use `qache` internally). The root cause was that `modulus` was never registered as a traced attribute of the `QuantumVariable` pytree, so it got smuggled into static `aux_data` (via `QuantumVariableTemplate`'s `copy.copy(qv)`) whenever it happened to be a live tracer — violating JAX's requirement that pytree aux_data be static/hashable/trace-independent.

## Related Issues

Closes #801
Related to #

## Type of Change

- [x] Bug Fix

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- `QuantumModulus.__init__` (quantum_modulus.py): added `"modulus"` to `self.traced_attributes` so it is threaded through the `QuantumVariable` pytree as a proper traced leaf instead of being embedded in static `aux_data`.
- Added 4 regression tests to test_jasp_modulus.py, each calling the affected function repeatedly to catch the cross-call tracer leak:
  - `test_traced_modulus_qache_repeated_calls`
  - `test_traced_modulus_custom_inversion_repeated_calls`
  - `test_traced_modulus_jnp_array_repeated_calls`
  - `test_static_modulus_hardcoded_literal_repeated_calls`
- Added a changelog entry under Bug Fixes in changelog-dev.rst.

## How was it tested?

| Test-ID | Type     | Description | Status |
|---------|----------|-------------|--------|
| T-001   | Positive| Traced modulus (`q` as a `@jaspify` function argument) inside `QuantumArray(QuantumModulus(q), ...)`, with a `@qache`-cached subroutine, called repeatedly (previously raised `UnexpectedTracerError` on 2nd call) | ✅ |
| T-002   | Positive | Same traced-modulus scenario as T-001, but going through `custom_inversion`/`conjugate` (original notebook pattern) | ✅ |
| T-003   | Positive | Dynamic, hardcoded modulus (`QuantumModulus(jnp.array(13))`) inside `@jaspify`, called repeatedly | ✅ |
| T-004   | Positive | Static, hardcoded modulus (`QuantumModulus(13)`) inside `@jaspify`, called repeatedly | ✅ |

Minimal reproduction (fails before this fix, passes after):

```python
from qrisp import QuantumArray, QuantumModulus, jaspify, measure
from qrisp.jasp import qache

@qache
def op(f, root):
    f[0] += root

@jaspify
def solve(q, root):
    f = QuantumArray(QuantumModulus(q), shape=(1,))
    op(f, root)
    return measure(f)

print(solve(13, 5))  # ok
print(solve(13, 5))  # previously: jax.errors.UnexpectedTracerError
```

## Checklist
- [x] I have added a changelog entry to changelog-dev.rst
- [ ] Breaking changes are documented with migration path (N/A — no breaking change)

## Reviewer Notes

The core defect is in `flatten_qv()`/`QuantumVariableTemplate` (qv_flattening.py): any `QuantumVariable` attribute not explicitly listed in `traced_attributes`/`static_attributes` gets shallow-copied wholesale into pytree `aux_data`, with no validation that it isn't a live tracer. This class of bug could in principle affect other `QuantumVariable` subtypes with similar untracked dynamic attributes — worth a broader audit, but out of scope for this PR.


positr0nium (approved):

PietropaoloFrisoni (approved):
🚀

PR #803, opened 2026-08-18
Use functools.wraps for decorators across the code base

renezander90:
## Description

Internal Qrisp functions decorated with `@custom_inversion` and/or `@custom_control` (e.g. `q_switch`/`qswitch`) lost their docstring and signature, since only `custom_control` applied `functools.wraps` — `custom_inversion` did not. This required manual `temp = f.__doc__; f = decorator(f); f.__doc__ = temp` workarounds scattered across the codebase, which still left signatures broken.

Adds `functools.wraps` to `custom_inversion`, `qache`, `RUS`, and `auto_uncompute` (matching the pattern already used by `custom_control`/`gate_wrap`), and removes the now-redundant manual docstring-copy workarounds in `quantum_switch.py`, switch_case.py, grover_tools.py, lcu.py, ripple_carry_adder.py, and SBP_arithmetic.py.

## What was changed?

- custom_inversion_environment.py, qaching.py, rus.py, uncomputation.py: added `functools.wraps` to the returned wrapper functions.
- Removed manual `__doc__`-copy workarounds in `quantum_switch.py`, switch_case.py, grover_tools.py, lcu.py, ripple_carry_adder.py, SBP_arithmetic.py.
- Changelog entry added under Improvements.

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -29,6 +29,12 @@ Improvements
   ``terminal_sampling()`` to use "sampling kernel" terminology and document
   the new arbitrary-return-value capability.
 
+- Decorators ``qache``, ``custom_inversion``, ``custom_control``,
+  ``RUS``, ``auto_uncompute`` now propagate the wrapped function's
+  docstring, name, and signature via ``functools.wraps``, removing the need
+  for manual docstring-copy workarounds at call sites.
+  (`PR #803 <https://github.com/eclipse-qrisp/Qrisp/pull/803>`_).
+
 Other New Features
 ------------------
 
--- a/src/qrisp/alg_primitives/arithmetic/SBP_arithmetic.py
+++ b/src/qrisp/alg_primitives/arithmetic/SBP_arithmetic.py
@@ -1353,17 +1353,6 @@ def app_phase_polynomial(qf_list, poly, symbol_list=None, t=1):
     app_sb_phase_polynomial(qf_list, sb_polynomial, symbol_list=new_symbol_list, t=t)
 
 
-# Workaround to keep the docstring but still gatewrap
-
-temp = app_sb_phase_polynomial.__doc__
-
 app_sb_phase_polynomial = gate_wrap(permeability="args", is_qfree=True)(app_sb_phase_polynomial)
 
-app_sb_phase_polynomial.__doc__ = temp
-
-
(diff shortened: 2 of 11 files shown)


purva-thakre:
@PietropaoloFrisoni How strictly do we want to follow the `reviewdog` failures for small self-contained PRs? If Rene were to rebase his branch, the reviewdog linting workflow will most likely fail for the files that appear under `changes`.

IMO we want to take `reviewdog` into consideration for new features that are being added. But for small maintenance PRs, it would be not be feasible. Thoughts?

PietropaoloFrisoni:
> If Rene were to rebase his branch, the reviewdog linting workflow will most likely fail for the files that appear under changes.

Reviewdog is supposed to only comment on the changed lines of code, not the entire files (or, at least, this is what happened when I tried it in my original PR), so the failures should hopefully be self-contained

purva-thakre:
## Code Review

Reviewed the `functools.wraps` decorator fix (`custom_inversion`, `RUS`, `qache`, `auto_uncompute`) and removal of the manual `__doc__` copy workarounds at their call sites.

**Correctness:** No bugs found. Verified across multiple independent passes that:
- `func`/`trial_function` are always unwrapped from the `*args` tuple (`func = func[0]`) before `functools.wraps` is applied — never receives a tuple.
- `inspect.getfullargspec` calls elsewhere in the codebase are unaffected, since it does not follow `__wrapped__` chains.
- Composition cases (e.g. `custom_control(custom_inversion(f))`) correctly propagate the original docstring/name through both wrapping layers.
- All 6 removed call-site workarounds map to a decorator that now correctly applies `functools.wraps`.

**One cleanup suggestion (non-blocking):**
In [`src/qrisp/algorithms/gqsp/hamiltonian_simulation.py` lines 231-234](https://github.com/eclipse-qrisp/Qrisp/blob/main/src/qrisp/algorithms/gqsp/hamiltonian_simulation.py#L231-L234), there's a commented-out block doing the old manual `__doc__` copy workaround for `hamiltonian_simulation`. It's inert (already commented out), but now stale since `qache` handles this via `functools.wraps`. Worth deleting or updating for clarity, since this PR's purpose is removing exactly this pattern elsewhere.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

purva-thakre:
Trying out Claude as a reviewer. Ignore the noise.

PR #814, opened 2026-08-25
Make split_cycle_shift function private + rename its internal vars

### LINKED ISSUE #810: Make "split_cycle_shift" from Dicke state preparation private
### Priority

Low

### Responsible

Jiri Guth Jarkovsky

### Type of Documentation Change

Existing documentation incorrect

### Affected Page / Section

https://qrisp.eu/reference/Primitives/DickeStates.html

### Short Description

The function's name is not prepended with an underscore, so it's not treated as a private function and therefore it's rendered in the Documentation. From what it does, it **should** be privated and therefore not visible in the Documentation.

This was pointed out here: https://github.com/eclipse-qrisp/Qrisp/pull/767#issuecomment-5353701596

### Long Description

_No response_

### Definition of Done

- [ ] Documentation written / updated
- [ ] Peer-reviewed for technical accuracy
- [ ] Code examples tested and working
- [ ] Renders correctly (checked locally or in preview)

JiriGuthJarkovsky:
## Description

Prepend the function name `split_cycle_shift` with an underscore to make it private. Change its internal variable names to follow more closely the referenced paper.

## Related Issues

Closes #810 

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [x] Yes
- [ ] No

Technically renaming a formerly-public function is a breaking change. Migration path? Just renaming the function and its inputs suffices. But it's unlikely to be required, since this was always an internal function with seemingly no use case outside of Dicke state preparation.

## What was changed?

- Function `split_cycle_shift` renamend to `_split_cycle_shift`.
- Its internal variables renamed from `highIndex` and `lowIndex` to `n` and `k`.
- Its internal loop variable renamed from `i` to `l` and its range has been slightly shifted, to follow the notation from the paper.

## How was it tested?

No tests for documentation change / refactoring. All current tests for the module still pass.

## Checklist

## Reviewer Notes

The renaming of the internal loop variables could theoretically introduce some bugs, but I've double-checked it (and all the tests pass), so it seems all good.

Technically renaming the function and its internal parameters is a breaking change.

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -144,7 +144,7 @@ API Changes
   ``_split_cycle_shift``, marking it private.  Its parameters were renamed
   from ``highIndex``/``lowIndex`` to ``n``/``k`` to match the notation of
   `arXiv:1904.07358 <https://arxiv.org/abs/1904.07358>`_.  The unitary
-  implemented is unchanged
+  implemented is unchanged.
   (`PR #814 <https://github.com/eclipse-qrisp/Qrisp/pull/814>`_).
 
 .. Add API changes above this line
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -139,6 +139,14 @@ API Changes
   not installed.
   (`PR #757 <https://github.com/eclipse-qrisp/Qrisp/pull/757>`_).
 
+* Renamed the *Split & Cyclic Shift* helper used by
+  :func:`~qrisp.dicke_state` from ``split_cycle_shift`` to
+  ``_split_cycle_shift``, marking it private.  Its parameters were renamed
+  from ``highIndex``/``lowIndex`` to ``n``/``k`` to match the notation of
+  `arXiv:1904.07358 <https://arxiv.org/abs/1904.07358>`_.  The unitary
+  implemented is unchanged
+  (`PR #814 <https://github.com/eclipse-qrisp/Qrisp/pull/814>`_).
+
 .. Add API changes above this line
 
 Development
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -144,7 +144,7 @@ API Changes
   ``_split_cycle_shift``, marking it private.  Its parameters were renamed
   from ``highIndex``/``lowIndex`` to ``n``/``k`` to match the notation of
   `arXiv:1904.07358 <https://arxiv.org/abs/1904.07358>`_.  The unitary
-  implemented is unchanged.
+  implemented is unchanged
   (`PR #814 <https://github.com/eclipse-qrisp/Qrisp/pull/814>`_).
 
 .. Add API changes above this line
--- a/tests/block_encodings_tests/test_block_encoding_sim.py
+++ b/tests/block_encodings_tests/test_block_encoding_sim.py
@@ -68,4 +68,4 @@ def test_block_encoding_sim():
 
     # Deviation floors at ~9.5e-07 with a bit of wobble, so a comparison against 1e-06 may sometimes fail.
     err = np.linalg.norm(c - amps)
-    assert err < 1e-5, f"Deviation {err:.3e} exceeds tolerance."
+    assert err < 1e-5
--- a/tests/block_encodings_tests/test_block_encoding_sim.py
+++ b/tests/block_encodings_tests/test_block_encoding_sim.py
@@ -65,4 +65,7 @@ def test_block_encoding_sim():
         return psi
 
     c = np.abs(psi_(0.5))
-    assert np.linalg.norm(c - amps) < 1e-6
+
+    # Deviation floors at ~9.5e-07 with a bit of wobble, so a comparison against 1e-06 may sometimes fail.
+    err = np.linalg.norm(c - amps)
+    assert err < 1e-5, f"Deviation {err:.3e} exceeds tolerance."
--- a/documentation/source/reference/Primitives/DickeStates.rst
+++ b/documentation/source/reference/Primitives/DickeStates.rst
@@ -6,5 +6,3 @@ Dicke State Preparation
 .. currentmodule:: qrisp
 
 .. autofunction:: dicke_state
-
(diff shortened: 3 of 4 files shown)


purva-thakre:
Given that the other workflow passed, I would say this is a non-deterministic failure. Could you please add a seed to the failing test? This would make it deterministic for the sake of the testing pipeline. Should be straightforward as adding the following line to the failing unit test

```py
np.random.seed(42)  # Deterministic for reproducible test results
```

JiriGuthJarkovsky:
> Given that the other workflow passed, I would say this is a non-deterministic failure. Could you please add a seed to the failing test? This would make it deterministic for the sake of the testing pipeline. Should be straightforward as adding the following line to the failing unit test
> 
> ```python
> np.random.seed(42)  # Deterministic for reproducible test results
> ```

Thanks, I tried that, but `numpy` seed doesn't derandomize the test result. The tested value `np.linalg.norm(c - amps)` floats around `9.5e-07` with a bit of noise, so a comparison against `1e-06` may occasionally fail. I don't understand this code and even boosted Claude was not able to find the source of the randomness.

To make sure it passes, I moved the comparison to `1e-05`, but I don't know if that doesn't invalidate the test somehow.

purva-thakre:
>  I tried that, but numpy seed doesn't derandomize the test result. 

😞 

> I moved the comparison to 1e-05, but I don't know if that doesn't invalidate the test somehow.

I am not too sure either as I haven't worked with the `BlockEncoding` class. Maybe @renezander90 or @MatP1337 could help?

tests/block_encodings_tests/test_block_encoding_sim.py
```diff
@@ -65,4 +65,7 @@ def psi_(t):
         return psi
 
     c = np.abs(psi_(0.5))
-    assert np.linalg.norm(c - amps) < 1e-6
+
+    # Deviation floors at ~9.5e-07 with a bit of wobble, so a comparison against 1e-06 may sometimes fail.
+    err = np.linalg.norm(c - amps)
+    assert err < 1e-5, f"Deviation {err:.3e} exceeds tolerance."
```
renezander90:
```suggestion
    assert err < 1e-5
```

The string for printing should not be necessary. Changing the threshold is fine. 
This test should actually be deterministic and completely unrelated to Dicke states. Any deviations likely originate from some numerical errors during the simulation.

documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -139,6 +139,14 @@ API Changes
   not installed.
   (`PR #757 <https://github.com/eclipse-qrisp/Qrisp/pull/757>`_).
 
+* Renamed the *Split & Cyclic Shift* helper used by
+  :func:`~qrisp.dicke_state` from ``split_cycle_shift`` to
+  ``_split_cycle_shift``, marking it private.  Its parameters were renamed
+  from ``highIndex``/``lowIndex`` to ``n``/``k`` to match the notation of
+  `arXiv:1904.07358 <https://arxiv.org/abs/1904.07358>`_.  The unitary
+  implemented is unchanged
+  (`PR #814 <https://github.com/eclipse-qrisp/Qrisp/pull/814>`_).
```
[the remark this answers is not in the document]
JiriGuthJarkovsky: Actually, I don't like the trailing period there. I like treating the parentheses afterwards as part of the same sentence. I see that in the changelog, this style varies.

tests/block_encodings_tests/test_block_encoding_sim.py
```diff
@@ -65,4 +65,7 @@ def psi_(t):
         return psi
 
     c = np.abs(psi_(0.5))
-    assert np.linalg.norm(c - amps) < 1e-6
+
+    # Deviation floors at ~9.5e-07 with a bit of wobble, so a comparison against 1e-06 may sometimes fail.
+    err = np.linalg.norm(c - amps)
+    assert err < 1e-5
```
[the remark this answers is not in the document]
JiriGuthJarkovsky: The change to 1e-5 was approved by a human reviewer (presumably) familiar with the test. So I'd keep it.

documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -139,6 +139,14 @@ API Changes
   not installed.
   (`PR #757 <https://github.com/eclipse-qrisp/Qrisp/pull/757>`_).
 
+* Renamed the *Split & Cyclic Shift* helper used by
+  :func:`~qrisp.dicke_state` from ``split_cycle_shift`` to
+  ``_split_cycle_shift``, marking it private.  Its parameters were renamed
+  from ``highIndex``/``lowIndex`` to ``n``/``k`` to match the notation of
```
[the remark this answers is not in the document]
JiriGuthJarkovsky: Good catch! Removed it manually.

documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -139,6 +139,14 @@ API Changes
   not installed.
   (`PR #757 <https://github.com/eclipse-qrisp/Qrisp/pull/757>`_).
 
+* Renamed the *Split & Cyclic Shift* helper used by
```
positr0nium, line 142: In my perception, the changelog should be a useful resource for developers that want to understand relevant changes. While this change is necessary, I would not deem it as super relevant since I'm not sure if anyone actually used this function outside of the Dicke-State preparation. What do you think, @renezander90 ?

JiriGuthJarkovsky: For what it's worth, it makes sense to me too.

Here, I decided to err on the side of caution, since I'm quite new to software development in general, and to Qrisp in particular (and it is technically speaking a breaking change).

renezander90: Technically, this is a breaking change, and the method is actually shown in the public documentation: https://qrisp.eu/reference/Primitives/DickeStates.html 

Hence, we should mention this in the change log. However, we could be less verbose regarding the implementation details.

positr0nium: Alright. I guess we can still make the final call when compiling the release notes.

positr0nium (approved):

PR #821, opened 2026-08-27
Add `mypy` via `reviewdog`

purva-thakre:
## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ docs = [
     "pydata-sphinx-theme==0.17.1",
 ]
 dev-code-style = [
+    "mypy==2.3.1",
     "ruff==0.15.18",
 ]
 dev-testing = [
@@ -127,6 +128,12 @@ ignore = [
     "PLC0415",  # import not at top of file (lazy/function-level imports to avoid circular imports)
     "E402",  # module-level import not at top of file (imports after a module docstring)
 ]
+
+[tool.mypy]
+python_version = "3.11"
+no_implicit_optional = true
+warn_redundant_casts = true
+warn_unused_ignores = true
 
 [tool.coverage.run]
 omit = ["*/__init__.py"]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -128,6 +129,12 @@ ignore = [
     "E402",  # module-level import not at top of file (imports after a module docstring)
 ]
 
+[tool.mypy]
+python_version = "3.11"
+no_implicit_optional = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+
```
purva-thakre, line 137: @PietropaoloFrisoni As asked by you, I have not added `check_untyped_defs = true` in order to ignore untyped ` *args/**kwargs`.

PietropaoloFrisoni:
Thanks! A couple of questions from my end:

1) Is this going to flag missing type hints in functions and methods arguments only or also in class attributes, instance attributes, global variables, etc. ?

2) Is this check going to be mixed with the violation of PEP rules (typically flagged by `ruff`, `pylint`, etc.) or will it be in a separate workflow/tab?

I would treat this is a very, very secondary check (definitely not with the same importance as PEP violations) because Python is dynamically typed by design. What do you think?

purva-thakre:
`mypy` is a type checker, `ruff` is a linter and formatter. I do not think there is an overlap in the two given that Astral has a separate tool that is supposed to be a faster drop-in replacement for `mypy`. See #822 

I have never used `pylint` directly. Only via `ruff`: https://docs.astral.sh/ruff/rules/#pylint-pl 
I think you could answer this question better than me about the `pylint` `mypy` overlap. 

> is this going to flag missing type hints in functions and methods arguments only or also in class attributes, instance attributes, global variables, etc. ?

Let me play around with all the things that `mypy` flags locally. I will come back with an answer on this. I only tested it on a couple of qrisp functions and called it a day. 😅

PR #832, opened 2026-08-28
Linear Combination of BlockEncodings

renezander90:
## Description

This PR improves `BlockEncoding` arithmetic by lowering linear combinations into a single flattened LCU construction. Mutually exclusive LCU branches reuse one shared ancilla workspace, reducing unnecessary ancilla allocation from recursively nested combinations.

## Related Issues

No related issue linked.

## Type of Change

- [x] Performance improvement
- [x] Documentation

## Breaking Change?

- [ ] Yes
- [x] No

## What was changed?

- Added `BlockEncoding.linear_combination()` for explicitly combining block encodings.
- Flattened chained addition, subtraction, and scalar multiplication into one LCU representation.
- Added JAX pytree support for preserving flattened LCU terms.
- Added `_AncillaLayout` to reconstruct heterogeneous typed ancilla views from one shared workspace.
- Added static and Jasp tests for ancilla reuse, flattening, serialization, validation, and numerical equivalence.
- Added API documentation and a changelog entry.

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001 | ✅ Arithmetic regression tests passed |
| T-002 | ✅ Flattened and direct linear-combination equivalence tested |
| T-003 | ✅ Ancilla layout tests passed in static and Jasp execution |
| T-004 | ✅ Input validation exception tests passed |

Additional validation: `git diff --check main...BE_experiments` passed.

## Screenshots / Output

Not applicable.

## Checklist

- [x] All relevant tests pass locally
- [x] I have added a changelog entry to changelog-dev.rst

## Reviewer Notes

Please review the shared ancilla workspace construction and the preservation of flattened LCU terms through JAX pytree serialization.

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -25,6 +25,14 @@ New Features
 Improvements
 ------------
 
+- **Flattened block-encoding arithmetic**
+  Block-encoding linear combinations now use a single LCU construction instead
+  of recursively nested binary combinations. Mutually exclusive LCU branches
+  reuse one shared ancilla workspace, including for heterogeneous ancilla
+  layouts. Block-encoding linear combinations can be constructed explicitly with
+  :meth:`~qrisp.block_encodings.BlockEncoding.linear_combination`
+  (`PR #832 <https://github.com/eclipse-qrisp/Qrisp/pull/832>`_).
+
 - :class:`~qrisp.interface.QiskitJob` and :class:`~qrisp.interface.AQTJob`
   now skip the live provider query and return the cached status once a job
   is done, cancelled, or errored. The :class:`~qrisp.interface.Job` base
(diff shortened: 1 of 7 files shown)


positr0nium:
I'm unsure about this architecture... What I don't like are two things:
1. The ``_lcu_terms`` makes the ``BlockEncoding`` class some sort of "silent subclass", in the sense that some objects have a meaningfull attribute and some don't. If we add it, from now on we always have to make sure that we cover both cases when handling it. A newcomer could also easily forget about the fact that it exists.
2. It seems to be possible for the ``_lcu_terms`` to carry contradicting information to the core attribute. This makes it error prone and difficult to bugfix because you always have to consider, which "source of truth" we are currently considering.

A better design: A new ``BlockEncoding`` subclass (``BlockEncodingPolynomial``?) that carries a polynomial in the following data structure:
``tuple[tuple[ArrayLike, tuple[BlockEncoding, ...]], ...]``
The core attributes ``alpha``, ``_anc_templates`` and ``unitary`` are exposed as properties so there is only a single source of truth.

What do you think?

renezander90:
Let's give it another try :)

Issue #830, opened 2026-08-28
Sphinx doctest build: fix failing docstring & tutorial examples


purva-thakre:
## Context

> [!Note]
> 
> This is closely related to #818, #795, #779, #334 

We should add two CI workflows to test the documentation:

- **`qrisp_docs_doctest.yml`** – runs `sphinx-build -b doctest` to execute the doctest blocks in docstrings (via autodoc) and in the `.rst`/`.md` tutorial and example files. Notebook execution is disabled (`NBSPHINX_EXECUTE=never`) so this job stays fast.
- **`qrisp_docs_tutorials.yml`** – executes the Jupyter tutorial notebooks by stripping their stored outputs and building the docs with nbsphinx (`NBSPHINX_EXECUTE=always`), so every code cell must run without error.

A trial run of the doctest job reports **~685 failures out of 1930 tests**. A large share of them are **false negatives** (missing setup or drifting output), but some reveal real bugs. This issue tracks cleaning up the docstrings and the doctest configuration so these workflows can be added in a green state.

## Tooling options: Sphinx doctest vs pytest doctest

The documentation examples can be tested with either Sphinx's doctest builder or pytest's built-in doctest support. Everything below applies to both, but the runner choice affects speed, coverage, and where the setup lives.

### Option A – Sphinx doctest

- Runs `sphinx-build -b doctest`; executes `>>>` blocks in docstrings (via autodoc) and in the `.rst`/`.md` tutorial and example files.
- Requires the full documentation build (autodoc imports, autosummary, nbsphinx conversion of notebooks), which is **slow**.
- Setup goes in `documentation/source/conf.py`: `doctest_global_setup = "from qrisp import *"` plus doctest flags (`ELLIPSIS`, `NORMALIZE_WHITESPACE`).
- Covers both source docstrings and the `.rst`/`.md` example files in one run.

### Option B – pytest doctest (faster)

- Runs `pytest --doctest-modules src/qrisp`; executes `>>>` blocks directly from the source files with **no Sphinx overhead**. Much faster and integrates with the existing test suite.
- Setup goes in `pyproject.toml`:
  ```toml
  [tool.pytest.ini_options]
  addopts = "--doctest-modules src/qrisp"
  doctest_namespace = "from qrisp import *"
  ```
  (`doctest_namespace` execs the code into every doctest's globals, the pytest equivalent of `doctest_global_setup`.)
[...]

PR #838, opened 2026-09-01
Add doctest workflow

renezander90:
## Description

Implementation strategy: run only on src/qrisp/qtypes, and expand incrementally throughout the code base as flagged issues are resolved.

## Related Issues

Closes #
Related to #830 

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist

### DIFF
--- a/.github/workflows/qrisp_doctest.yml
+++ b/.github/workflows/qrisp_doctest.yml
@@ -1,0 +1,47 @@
+name: Qrisp doctests via Pytest
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  doctest:
+    name: Doctest Py${{ matrix.python-version }} (${{ matrix.package.name }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12.*"]
+        package:
+          - name: qtypes
+            path: src/qrisp/qtypes
+
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v7
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[all]"
+      - name: Run ${{ matrix.package.name }} doctests
+        run: |
+          python -m pytest --doctest-modules --doctest-continue-on-failure \
+            "${{ matrix.package.path }}"
--- a/src/qrisp/qtypes/quantum_bool.py
+++ b/src/qrisp/qtypes/quantum_bool.py
@@ -41,7 +41,7 @@ class QuantumBool(QuantumVariable):
 
     >>> from qrisp import QuantumBool, h
     >>> q_bool_0 = QuantumBool()
-    >>> h(q_bool_0)
+    >>> _ = h(q_bool_0)
     >>> print(q_bool_0)
     {False: 0.5, True: 0.5}
 
@@ -66,7 +66,7 @@ class QuantumBool(QuantumVariable):
 
 
     >>> qf = QuantumFloat(4)
-    >>> h(qf[3])
+    >>> _ = h(qf[3])
     >>> print(qf)
     {0: 0.5, 8: 0.5}
     >>> q_bool_3 = (qf >=  4)
@@ -76,10 +76,8 @@ class QuantumBool(QuantumVariable):
     To use a QuantumBool as a :ref:`ControlEnvironment`, we simply put it in a ``with``
     statement:
 
-    ::
-
-        with q_bool_3:
-            qf += 2
+    >>> with q_bool_3:
+    ...     qf += 2
 
     >>> print(qf)
     {0: 0.5, 10: 0.5}
@@ -87,12 +85,10 @@ class QuantumBool(QuantumVariable):
     QuantumBools that are created directly after a ``with`` statement are uncomputed
     automatically:
 
-    ::
+    >>> with qf == 10:
+    ...     _ = q_bool_3.flip()
 
-        with qf == 10:
-            q_bool_3.flip()
-
-    >>> print(qf.qs)
+    >>> _ = qf.qs
 
     .. code-block:: none
 
--- a/src/qrisp/qtypes/quantum_float.py
+++ b/src/qrisp/qtypes/quantum_float.py
@@ -226,7 +226,7 @@ class QuantumFloat(QuantumVariable):
 
     >>> d = a//b
     >>> print(d)
-    {3.0: 1.0}
+    {3: 1.0}
 
     Inversion:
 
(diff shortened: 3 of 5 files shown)


PietropaoloFrisoni:
Thanks @renezander90 . Should we start adding `<BLANKLINE>` in docstrings?

purva-thakre:
@PietropaoloFrisoni Feel free to edit the example docstring example in https://github.com/eclipse-qrisp/Qrisp/issues/830 if you think it needs something additional. 

we likely won't be turning on doctest for quite a while. I was planning to take a look at this myself after QCE. Thankfully, Rene jumped on this :)

purva-thakre:
@renezander90 Since you started looking into this, could you check if `jupyter-sphinx` is a better alternative to `doctest`? Pietro's comment reminded me that we would have to spend a lot of time making sure everything in the docstring examples output is formatted correctly to make doctest happy. 

With `jupyter-sphinx`, you simply add the code in the examples section  (minus the output) and the code block gets executed when you run the sphinx build. I have used this in a past project before the project moved to a md based docs workflow. https://jupyter-sphinx.readthedocs.io/en/latest/

renezander90:
It is recommended to use a combination of `doctest` and `jupyter-sphinx`:

Using only `jupyter-sphinx` could work, but I would not recommend it for Qrisp.

The main issue is that `jupyter-sphinx` is primarily an **executable documentation renderer**, whereas `pytest` doctest is also a **regression-testing mechanism**. With only `jupyter-sphinx`:

- existing `>>>` docstring examples must be rewritten as `jupyter-execute` directives;
- docstrings are no longer automatically tested as Python source examples;
- examples that merely run without exceptions may still produce incorrect quantum results unless they contain explicit `assert` statements;
- documentation builds require a working Jupyter kernel and all runtime dependencies;
- build time and reproducibility become more sensitive to simulations, measurements, randomness, and backend behavior;
- Qrisp already uses `nbsphinx` for complete notebook tutorials, so there would be overlap.

For example, this is weaker as a regression test:

```rst
.. jupyter-execute::

   qf = QuantumFloat(3, -1)
   qf[:] = 2.5
   print(qf)
```

It verifies that the code runs, but an incorrect printed value may not fail the build in a meaningful way because the output is generated automatically.

This is stronger:

```rst
.. jupyter-execute::

   qf = QuantumFloat(3, -1)
   qf[:] = 2.5
   assert qf.get_measurement() == {2.5: 1.0}
```

My recommendation remains:

```text
pytest doctest       short API/docstring regression tests
jupyter-sphinx       executable RST/Markdown examples with generated output
nbsphinx             complete notebook tutorials
```

Using only `jupyter-sphinx` is reasonable only if the project intentionally wants to migrate all examples out of docstrings and accepts that documentation execution, rather than exact doctest expectations, will be the primary verification mechanism.

renezander90:
@PietropaoloFrisoni `<BLANKLINE>` should be used wherever it is appropriate:

Use `<BLANKLINE>` when the **actual program output** contains an empty line:

```python
>>> print("start")
>>> print()
>>> print("end")
start
<BLANKLINE>
end
```

Do **not** use it for formatting the docstring or separating examples:

```python
>>> x = 1

>>> x + 1
2
```

That blank line separates two doctest interactions; it is not output.

Also do not use it when the output has no blank line:

```python
>>> print("start")
start
>>> print("end")
end
```

So the rule is: use `<BLANKLINE>` only where the executed code actually prints an empty line.

PietropaoloFrisoni:
It should even appear in the rendered `sphinx` version if used properly :+1:

renezander90:
> @PietropaoloFrisoni Feel free to edit the example docstring example in #830 if you think it needs something additional.
> 
> we likely won't be turning on doctest for quite a while. I was planning to take a look at this myself after QCE. Thankfully, Rene jumped on this :)

Actually, I would like to turn it on as soon as possible. We could first keep it restricted to `src/qrisp/qtypes` as soon as the issues in this module are fixed (after #846, #848). This would ensure that all documentation examples in this module will work perfectly:) Afterwards, we can have a series of follow-up PR's to extend it gradually to more modules of Qrisp.

Regarding the necessary formatting changes, this can be done quite fast with AI. Only bugs or more fundamental problems have to be resolved manually (e.g. the arithmetic bugs I discovered when using doctest on the `qtypes` module).

purva-thakre:
@renezander90 If that's the case, before turning it on, could you compare `sphinx doctest` to `pytest doctest`? From experience, `sphinx doctest` takes a loooot of time. The LLM I am using recommended `pytest doctest` as a better alternative. 

> Regarding the necessary formatting changes, this can be done quite fast with AI.

If you say so! I am a bit cautious that the LLM could make unnecessary changes that I would fail to catch.

renezander90:
Some recommendation on dealing with randomness:

## Recommendation for nondeterministic examples

The goal should be to keep examples natural while avoiding assertions about values that are inherently unstable.

| Situation | Recommended doctest approach |
|---|---|
| Deterministic output | Show and verify the exact output |
| Floating-point variation | Check a tolerance instead of printing the value |
| Variable formatting or generated names | Use `ELLIPSIS` or `NORMALIZE_WHITESPACE` |
| Probabilistic quantum result | Check an invariant or statistical property |
| Random, backend-dependent, unavailable, or expensive example | Use `SKIP` with a clear reason |

#### Deterministic output

```python
>>> qf = QuantumFloat(3, -1)
>>> qf[:] = 2.5
>>> qf.get_measurement()
{2.5: 1.0}
```

#### Floating-point results

Instead of maintaining a fragile decimal representation:

```python
>>> computed_value = calculate_value()
>>> abs(computed_value - expected_value) < 1e-8
True
```

This keeps the example readable while checking the meaningful behavior.

#### Variable formatting

For generated names or partially unstable textual output:

```python
>>> print(circuit)  # doctest: +ELLIPSIS
QuantumCircuit: ...
```

Use this only when the omitted portion is genuinely irrelevant. Do not hide the actual result behind `...`.

For insignificant whitespace differences:

```python
>>> print(description)  # doctest: +NORMALIZE_WHITESPACE
A description with normalized whitespace.
```

#### Probabilistic quantum results

Avoid showing one exact measurement result when several outcomes are valid:

```python
>>> qv = QuantumVariable(1)
>>> h(qv)
>>> measurement = measure(qv)
>>> set(measurement).issubset({"0", "1"})
True
>>> sum(measurement.values())
1.0
```

For statistical behavior, use a sufficiently large number of shots and test a broad property rather than exact frequencies. Detailed statistical checks are usually better placed in unit tests than in doctests.

#### Random or unsuitable examples

`SKIP` is acceptable when you genuinely want to show a random result, but do not want the documentation build to verify it:

```python
>>> random_result()  # doctest: +SKIP
0.731842...
```

The result is displayed as illustrative documentation, but the code is not executed or checked. Add prose making this clear:

```text
The following output is an example; the actual result varies between runs.
```

`SKIP` is also appropriate for examples that are expensive, backend-specific, require unavailable credentials, or depend on external services:

```python
>>> run_on_hardware_backend()  # doctest: +SKIP
...
```

Use it sparingly, because skipped examples provide no automated correctness guarantee.

The resulting policy is:

```text
deterministic output  -> exact doctest output
floating-point drift  -> tolerance check
variable formatting   -> ELLIPSIS or NORMALIZE_WHITESPACE
probabilistic result  -> invariant or statistical property
random example to show -> SKIP with an explicit explanation
unavailable/expensive  -> SKIP with a clear reason
```

Issue #842, opened 2026-09-01
QuantumFloat encoder


renezander90:
**Priority:** Medium  
**Responsible:**  
**Qrisp Version:** `main`  
**Python Version:** `3.11`  
**Operating System:** macOS  

**Short Description**

`QuantumFloat.encoder` silently truncates non-representable values instead of rejecting them. For a `QuantumFloat` with exponent `-1`, assigning `0.5102341` silently encodes `0.5`. (found is scope of adding doctest workflow #838)

**Steps to Reproduce**

```python
from qrisp import QuantumFloat

qf = QuantumFloat(4, -1)

# The representable values are multiples of 0.5:
# 0.0, 0.5, 1.0, ..., 7.5
qf[:] = 0.5102341

print(qf.encoder(0.5102341))
```

**Expected Behavior**

The assignment should raise a `ValueError` because `0.5102341` is not exactly representable by `QuantumFloat(4, -1)`. Explicit rounding should remain available through `encode(..., rounding=True)` or `truncate(...)`.

**Actual Behavior**

No exception is raised. The value is scaled by the exponent:

```python
0.5102341 / 0.5 = 1.0204682
```

Then `astype(int)` truncates this to `1`, causing the value `0.5` to be encoded silently.

**Means of Verification**

| Test-ID | Type | Description | Expected Result |
|---------|------|-------------|-----------------|
| T-001 | Positive | Encode `0.5` into `QuantumFloat(4, -1)` | Succeeds |
| T-002 | Negative | Encode `0.5102341` without rounding | Raises `ValueError` |
| T-003 | Boundary | Encode `0.0` | Succeeds |
| T-004 | Boundary | Encode `7.5` | Succeeds |

**Additional Context**

The relevant conversion is:

```python
res = i / jnp.float64(2) ** self.exponent
return res.astype(int)
```

`astype(int)` truncates toward zero; it does not perform nearest-integer rounding.

vrinda-054:
Hi! I'd like to work on this issue if it's still open. I've traced the bug to QuantumFloat.encoder's truncation logic and have a fix in progress. Could I get this assigned to me?

merelyacryptid:
@purva-thakre hi, i am interested in working on this! can i be assigned this issue?

Issue #840, opened 2026-09-01
QuantumFloat subtraction


renezander90:
### Priority

High

### Qrisp Version

0.9.7

### Python Version

3.11.15

### Operating System

macOS

### Short Description

TypeError for QuantumFloat subtraction (found is scope of adding doctest workflow #838)

### Steps to Reproduce

```
from qrisp import QuantumFloat
a = QuantumFloat(3, -1, signed = False)
b = QuantumFloat(2, -2, signed = True)

a[:] = 1.5
b[:] = 0.25
c = a + b

d = a - c
print(d)
```
>TypeError: Integers cannot be raised to negative powers, got integer_pow(~int64[], -2)

### Expected Behavior

>{-0.25: 1.0}

### Actual Behavior

>TypeError

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive |             |                 |
| T-002   | Negative |             |                 |
| T-003   | Boundary |             |                 |
| T-004   | Boundary |             |                 |

### Additional Context

The **root cause** is in quantum_float.py:

```python
exponent = jnp.minimum(...)
max_sig = jnp.maximum(...)
```

`create_output_qf()` is used during ordinary static arithmetic, but these calls produce JAX scalar arrays. Thus:

```python
c = a + b
type(c.exponent)  # ArrayImpl, value -2
```

Then `a - c` calls `sbp_sub()`, which evaluates:

```python
return 2**self.exponent * poly
```

in quantum_float.py. Because `self.exponent` is a JAX integer array with value `-2`, JAX dispatches integer exponentiation and raises:

```text
TypeError: Integers cannot be raised to negative powers
```

So the root issue is **JAX array metadata leaking into a static `QuantumFloat.exponent` field**, not a subtraction algorithm error. The likely fix is to use Python `min`/`max` in the static `create_output_qf()` path, while preserving JAX operations only for tracing mode.

### Definition of Done

- [x] Root cause identified
- [ ] Fix implemented and peer-reviewed
- [ ] Regression test added
- [ ] All tests pass (CI green)
- [ ] Changelog entry added

renezander90:
Fixed in #846

PR #841, opened 2026-09-01
Refactor create_output_qf to reduce complexity and fix issue #840

renezander90:
## Description

Refactor `create_output_qf` into focused helpers and fix static QuantumFloat arithmetic leaking JAX scalar metadata.

## Related Issues

Closes #840
Related to #

## Type of Change

- [ ] Feature (new functionality)
- [x] Change Request (modification of existing functionality)
- [x] Bug Fix
- [x] Refactoring (no behavior change)
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?

- [ ] Yes
- [x] No

## What was changed?

- Split output qtype inference into polynomial, modulus, and arithmetic helpers.
- Preserve traced JAX values while keeping static exponents and sizes as Python integers.
- Added concise docstrings and a regression test for static add/subtract arithmetic.

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ Focused static arithmetic regression |
| T-002   | ✅ `tests/core_tests/test_quantum_float.py` |
| T-003   | ✅ `tests/core_tests/quantum_array/test_quantum_array.py` |
| T-004   | ✅ `tests/jax_tests/test_jasp_quantum_array.py` |

## Screenshots / Output (if applicable)

Not applicable.

## Checklist

- [x] My code follows the project's coding standards
- [x] I have performed a self-review
- [x] I have added/updated tests
- [ ] All tests pass locally and in CI
- [x] I have updated the documentation
- [ ] I have added a changelog entry to `changelog-dev.rst`
- [ ] Breaking changes are documented with migration path

PR #827, opened 2026-08-27
Small refactoring of the BigInteger class

PietropaoloFrisoni:
## Description

While studying the nice work that @diehoq did with the `BigInteger` class, I took the chance to perform some refactorings and small code improvements with the help of Claude. This PR refactors the `BigInteger`/Jasp-Montgomery arithmetic backend and Shor's-algorithm/RSA code used for large-integer (up to 2048-bit) factorization. It adds type hints and docstrings throughout, fixes several real crashes uncovered in the process, and adds test coverage for previously-untested code paths.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Feature (new functionality)
- [ ] Change Request (modification of existing functionality)
- [x] Bug Fix
- [x] Refactoring (no behavior change)
- [x] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

Five internal `qrisp.shor` helper functions were renamed to be private (leading underscore). None were part of the documented public API, but a missing `__all__` left them technically importable via `from qrisp.shor import *`.

## What was changed?

- Fixed real crashes: `bi_contfrac_best_approx`/`bi_shor_recover_denominator` (broken `lax.while_loop`/`lax.select` usage, plus a corrected semiconvergent formula), `montgomery_encoder`/`montgomery_decoder` crashing on mixed classical/`BigInteger` inputs, a completely broken traced `egcd`, and `QuantumArray @ classical_matrix` crashing for `numpy`/`jax.numpy` integer matrices.
- Sped up `BigInteger.__lshift__`/`__rshift__` from an O(total bits) bit-by-bit loop to a limb-wise shift (~3x faster at 2048-bit width).
- Added type hints and docstrings (with doctest-verified `Examples`) across `jasp_bigintiger.py`, `jasp_mod_tools.py`, `jasp_montgomery.py`, `shors_algorithm.py`, `crypto_tools.py`, and the static `modular_arithmetic/mod_tools.py` (which previously had no docstrings at all).
- Added test coverage for ~20 previously-untested functions in the BigInteger/Montgomery backend, and for three files with zero prior test references (`shors_algorithm.py`, `crypto_tools.py`, static `mod_tools.py`).
- Renamed five internal Shor/RSA helper functions to be private (see Breaking Change note above); resolves the resulting reviewdog `missing-docstring-in-public-function` failures.
- Left `TODO` comments for two known, out-of-scope issues found during review (see Reviewer Notes).

## How was it tested?

[...]

PR #846, opened 2026-09-02
Improvements to `QuantumFloat`

PietropaoloFrisoni:
## Description

Cleans up `QuantumFloat`: adds type hints, fixes a silent exponent-type bug, replaces an exponential-time rounding path with an O(1) one, tightens exception types, and substantially expands test coverage.

## Related Issues

Closes #
Related to #

## Type of Change

- [x] Bug Fix
- [x] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

`QuantumFloat` methods now raise specific exception types (`TypeError`, `ValueError`, `NotImplementedError`) instead of a generic `Exception`. Code catching `except Exception:` is unaffected; code catching the generic `Exception` type specifically would need to broaden its `except` clause.

## What was changed?

- Added type hints across `QuantumFloat` (`src/qrisp/qtypes/quantum_float.py`) and `q_mult`/`hybrid_mult` (`src/qrisp/alg_primitives/arithmetic/SBP_arithmetic.py`), using `TYPE_CHECKING`-guarded imports to avoid circular imports.
- Fixed a bug where `QuantumFloat` addition/subtraction with different exponents could silently produce a `jax.Array` exponent instead of a plain `int` outside tracing, crashing later `2**exponent` calls.
- Replaced `encode(rounding=True)`'s exponential-time (`O(2^size)`) enumeration with a call to `truncate()`, an O(1) round-and-clip against the uniform representable-value grid (per reviewer feedback on this PR). Verified equivalent to the original across 13,000+ probe cases; the only divergence is round-half-to-even tie-breaking at exact midpoints instead of an arbitrary encoding-order artifact.
- Sped up `significant()` and `init_from()`.
- Replaced 26 generic `raise Exception(...)` call sites in `QuantumFloat` with specific exception types (`TypeError`, `ValueError`, `NotImplementedError`).
- Fixed `q_mult` silently returning `None` for an unrecognized `method` argument; now raises a `ValueError` naming the valid options. Gave `q_mult` a full docstring (it had none).
- Reverted an internal refactor (extracting the six comparison dunders' shared logic into a `_compare()` helper) after discovering it silently broke `with qf == 0:`-style conditional-environment usage in CI, by adding a stack frame that `adaptive_condition`'s fixed-depth caller-line lookup depends on. Added a regression test (`test_comparison_as_conditional_environment`) covering this usage pattern.
- Fixed stale/broken docstring examples, typos, and grammar throughout `QuantumFloat`'s documentation (e.g. a broken `:meth:` cross-reference from `exp_shift`'s docstring to `quantum_bit_shift`, "amount" → "number" for qubit counts, "inplace" → "in-place" for consistency, missing/misplaced punctuation). Verified via a full Sphinx HTML rebuild that all edited text renders correctly with no new warnings.
- Added 24 new tests to `tests/core_tests/test_quantum_float.py` (covering `__truediv__`, `__floordiv__`, `__pow__`'s inversion branch, in-place arithmetic, classical-scalar branches, `msize`, `sb_poly()`, `jdecoder()`, `quantum_bit_shift()`, the module-level `create_output_qf`/polynomial-encoding path, and the comparison-as-conditional-environment regression) and 2 new tests to `tests/jax_tests/test_jasp_QuantumFloat.py` (comparisons and arithmetic operators under Jasp tracing, previously untested).
- Fixed a real name-shadowing bug and a function-redeclaration bug found while adding tests (local `x`/`y` sympy symbols shadowing imported gate functions; two test functions both named `main`).
- Trimmed excessive multi-line class/method docstrings in the test files to single lines.
- Consolidated the changelog down to 3 entries for this PR (Improvements, Bug Fixes, Compatibility).

## How was it tested?

| Test-ID | Status |
|---------|--------|
| `tests/core_tests/test_quantum_float.py` | ✅ |
| `tests/jax_tests/test_jasp_QuantumFloat.py` | ✅ |
| `test_conditional_environments_example`, `test_custom_control` (CI regression, root-caused and fixed) | ✅ |
| `pyright`, `ruff format --check`, `ruff check`, `pylint` on all changed files | ✅ (no new findings) |
| `python -m doctest` on `quantum_float.py` | ✅ (all 105 doctest examples content-correct) |
| Sphinx HTML doc build | ✅ (no new warnings; edited text spot-checked in rendered output) |

## Checklist

## Reviewer Notes

### DIFF
--- a/src/qrisp/qtypes/quantum_float.py
+++ b/src/qrisp/qtypes/quantum_float.py
@@ -237,7 +237,7 @@ class QuantumFloat(QuantumVariable):
 
     >>> d = a//b
     >>> print(d)
-    {3.0: 1.0}
+    {3: 1.0}
 
     Inversion:
 
@@ -659,8 +659,7 @@ class QuantumFloat(QuantumVariable):
             return output_qf
 
         raise TypeError(
-            "QuantumFloat multiplication for type " + str(type(other)) + ""
-            " not implemented (available are QuantumFloat and int)"
+            f"QuantumFloat multiplication for type {type(other)} not implemented (available are QuantumFloat and int)"
         )
 
     @gate_wrap(permeability="args", is_qfree=True)
@@ -682,7 +681,7 @@ class QuantumFloat(QuantumVariable):
             cx(self, res)
             res += other
             return res
-        raise TypeError("Addition with type " + str(type(other)) + " not implemented")
+        raise TypeError(f"Addition with type {type(other)} not implemented")
 
     @gate_wrap(permeability="args", is_qfree=True)
     def __sub__(self, other: QuantumFloat | FloatLike) -> QuantumFloat:
@@ -703,7 +702,7 @@ class QuantumFloat(QuantumVariable):
             cx(self, res)
             res -= other
             return res
-        raise TypeError("Subtraction with type " + str(type(other)) + " not implemented")
+        raise TypeError(f"Subtraction with type {type(other)} not implemented")
 
     __radd__ = __add__
     __rmul__ = __mul__
@@ -723,7 +722,7 @@ class QuantumFloat(QuantumVariable):
             x(res)
             res += other + 2**res.exponent
             return res
-        raise TypeError("Subtraction with type " + str(type(other)) + " not implemented")
+        raise TypeError(f"Subtraction with type {type(other)} not implemented")
 
     @gate_wrap(permeability="args", is_qfree=True)
     def __truediv__(self, other: QuantumFloat | FloatLike) -> QuantumFloat:
@@ -806,7 +805,8 @@ class QuantumFloat(QuantumVariable):
             polynomial_encoder(input_qf_list, self, poly)
 
         elif isinstance(other, (int, float, np.number)):
-            if not int(other / 2**self.exponent) == other / 2**self.exponent:
+            scaled = other / 2**self.exponent
+            if int(scaled) != scaled:
                 raise ValueError(
                     "Tried to perform in-place addition with invalid number. QuantumFloat precision too low."
                 )
@@ -817,7 +817,7 @@ class QuantumFloat(QuantumVariable):
             polynomial_encoder(input_qf_list, self, poly)
 
         else:
-            raise TypeError("In-place addition for type " + str(type(other)) + " not implemented")
+            raise TypeError(f"In-place addition for type {type(other)} not implemented")
 
         return self
 
@@ -839,7 +839,8 @@ class QuantumFloat(QuantumVariable):
             polynomial_encoder(input_qf_list, self, poly)
 
         elif isinstance(other, (int, float, np.integer, np.floating)):
-            if not int(other / 2**self.exponent) == other / 2**self.exponent:
+            scaled = other / 2**self.exponent
+            if int(scaled) != scaled:
                 raise ValueError(
                     "Tried to perform in-place subtraction with invalid number. QuantumFloat precision too low."
                 )
@@ -850,7 +851,7 @@ class QuantumFloat(QuantumVariable):
             polynomial_encoder(input_qf_list, self, poly)
 
         else:
-            raise TypeError("In-place substraction for type " + str(type(other)) + " not implemented")
+            raise TypeError(f"In-place subtraction for type {type(other)} not implemented")
 
         return self
 
@@ -970,22 +971,19 @@ class QuantumFloat(QuantumVariable):
         >>> print(a)
         {8: 1.0}
         >>> print(a.qs)
-
-        .. code-block:: none
-
-            QuantumCircuit:
-            --------------
-            a.0: ─────
-                 ┌───┐
-            a.1: ┤ X ├
-                 └───┘
-            a.2: ─────
-            <BLANKLINE>
-            a.3: ─────
-
-            Live QuantumVariables:
-            ---------------------
-            QuantumFloat a
+        QuantumCircuit:
+        ---------------
+        a.0: ─────
+             ┌───┐
+        a.1: ┤ X ├
+             └───┘
+        a.2: ─────
+        <BLANKLINE>
+        a.3: ─────
+        <BLANKLINE>
+        Live QuantumVariables:
+        ----------------------
+        QuantumFloat a
 
         """
         if not isinstance(shift, int):
@@ -1013,7 +1011,7 @@ class QuantumFloat(QuantumVariable):
 
         """
         if self.signed:
-            raise ValueError(r"Tried to add sign to signed QuantumFloat")
+            raise ValueError("Tried to add sign to signed QuantumFloat")
 
         self.extend(1, self.size)
         self.signed = True
@@ -1244,7 +1242,7 @@ class QuantumFloat(QuantumVariable):
 
         .. note::
 
-            Bit bit shifts based on a QuantumFloat are currently only possible
+            Bit shifts based on a QuantumFloat are currently only possible
             if both self and ``shift_amount`` are unsigned.
 
         .. warning::
@@ -1444,7 +1442,7 @@ def create_output_qf(operands: list[QuantumFloat], op: str | sp.Expr) -> Quantum
         msize = max_sig - min_sig
         exponent = min_sig
 
-        signed = bool(sum(int(operand.signed) for operand in operands))
+        signed = any(operand.signed for operand in operands)
 
         return QuantumFloat(msize, exponent=exponent, signed=signed)
 
@@ -1529,7 +1527,7 @@ def copy_qf(
 
     if qf2.signed:
         if not qf1.signed:
-            raise ValueError("Tried to copy signed into unsigend float")
+            raise ValueError("Tried to copy signed into unsigned float")
 
         qf1_len -= 1
         qf2_len -= 1
--- a/src/qrisp/qtypes/quantum_float.py
+++ b/src/qrisp/qtypes/quantum_float.py
@@ -111,6 +111,13 @@ class QuantumFloat(QuantumVariable):
     Here, the 3 indicates the number of mantissa qubits and the -1 indicates the
     exponent.
 
+    .. note::
+
+        ``msize`` and ``exponent`` also accept a ``jax.Array`` under Jasp
+        tracing, but only in the sense of a 0-d, traced scalar (e.g. a value
+        computed from a measurement inside ``@jaspify``/``make_jaspr``).
+        Never an actual multi-element array.
+
     For unsigned QuantumFloats, the decoder function is given by
 
     .. math::
@@ -327,6 +334,14 @@ class QuantumFloat(QuantumVariable):
         signed: bool = False,
     ) -> None:
         """Construct a QuantumFloat with the given mantissa size, exponent, and sign.
+
+        .. note::
+
+            The ``jax.Array`` accepted by ``msize`` and ``exponent`` is a
+            0-d, traced scalar under Jasp tracing (e.g. a value computed from
+            a measurement inside ``@jaspify``/``make_jaspr``), not an actual
+            multi-element array. Outside of tracing, both are plain Python
+            integers.
 
         Parameters
         ----------
--- a/src/qrisp/qtypes/quantum_float.py
+++ b/src/qrisp/qtypes/quantum_float.py
@@ -621,7 +621,7 @@ class QuantumFloat(QuantumVariable):
         super().encode(value, permit_dirtyness=permit_dirtyness)
 
     @gate_wrap(permeability="args", is_qfree=True)
-    def __mul__(self, other: QuantumFloat | FloatLike) -> QuantumFloat:
+    def __mul__(self, other: QuantumFloat | int | np.integer) -> QuantumFloat:
         """Multiply this QuantumFloat by another QuantumFloat or a classical int."""
         if check_for_tracing_mode():
             # NOTE: Local import to avoid a circular import (qrisp.alg_primitives.arithmetic imports from qrisp.qtypes).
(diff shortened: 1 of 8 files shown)


renezander90:
If I'm not mistaken, this PR also fixes #840 ?

There was a dedicated PR for this: #841 

However, this more comprehensive overhaul of the `QuantumFloat` is very welcome! Can #841 then be closed?
I would still prefer splitting the `create_output_qf` (sympy/float/modulus as done in #841).

src/qrisp/qtypes/quantum_float.py
```diff
@@ -455,54 +548,100 @@ def sb_poly(self, m=0):
...
+            qubits. The default is False.
 
+        Returns
+        -------
+        None
+
+        """
+        value = encoding_number
+        if rounding:
+            # Round value to closest fitting number. Vectorized equivalent of
```
renezander90: I think this exponentially expensive logic is not necessary. See `truncate` method.

PietropaoloFrisoni:
Thanks @renezander90 , I created this PR because I was studying `QuantumFloat` and I realized that I do not understand some of its components. I took the chance to open this PR in the meantime (there are many other improvements I am working on, always with the help of Claude). However this is not ready for review as there are many things I still need to check. If you want to go ahead with other PRs instead, feel free to go (I will resolve merge conflicts while they arise)

purva-thakre:
@PietropaoloFrisoni Just letting you know I am trying to understand the `QuantumModulus` class. I will likely have a PR soon because the methods of this class lack proper docstrings. Wanted to make sure we don't double our efforts by working on the same module.

PietropaoloFrisoni:
@purva-thakre Thanks for letting me know. I am focusing exclusively on `QuantumFloat` here (the changes in other files are just to suppress `pylance` warnings appearing here), so there should not be much overlap.

P.S. I am doing some refactorings to that module in [this PR](https://github.com/eclipse-qrisp/Qrisp/pull/827) as well

renezander90:
@purva-thakre This PR: #827 refactoring parts of the infrastructure surrounding `QuantumModulus` could still be relevant. Regarding, `QuantumModulus`, some confusion within the class could arise from the fact that it's using two different modular arithmetic implementations: `jasp_arithmetic/jasp_montgomery` (new, Jasp compatible) and `modular_arithmetic/modular_multiplcation`(legacy, not Jasp compatible). We should further consolidate this. (I just discussed this yesterday with GPT.) This PR: #710 will eliminate some discrepancy in the Montgomery shift between the static/dynamic pipelines. Also tagging @diehoq.

Bildschirmfoto 2026-09-02 um 14 15 36

This is of course unrelated to the present PR, thus I apologise for spamming here!

P.S.: We should use issues or draft PR's or discord to keep aligned on which parts of the code base different people are working on to avoid duplicate work.

renezander90:
Please also consider the issues that @positr0nium 's Claude pointed out on my PR: 
[repro_pr841.py](https://github.com/user-attachments/files/31747366/repro_pr841.py)

tests/jax_tests/test_jasp_QuantumFloat.py
```diff
@@ -60,13 +60,104 @@ def inner(qf):
...
+    check_signed_static_exponent_dynamic()
+
+
+def test_jasp_QuantumFloat_comparisons():
+    """Test QuantumFloat's comparison operators under Jasp tracing."""
+    from qrisp import QuantumFloat
+    from qrisp.jasp import terminal_sampling
+
+    @terminal_sampling
+    def compare(a_value, b_value, op):
```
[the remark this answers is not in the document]
PietropaoloFrisoni: This is false (checked locally). Resolving this

PietropaoloFrisoni:
@renezander90 @purva-thakre  I will not request Copilot's reviews anymore since it's basically a rabbit hole. It keeps finding issues, and most of them are pre-existing edge cases that will never come up in practical usage, or they are out of scope for this PR (which was meant to be a quick refactoring but it already costed too much time and tokens)

src/qrisp/qtypes/quantum_float.py
```diff
@@ -86,45 +97,44 @@ def _signed_int_iso_inv(y, n):
...
 
-# def signed_int_iso(x, n):
-#    if int(x) < -(2**n) or int(x) >= 2**n:
-#        raise Exception("Applying signed integer isomorphism resulted in overflow")
-
-#    if x >= 0:
-#        return x % 2**n
-#    else:
-#        return -abs(x) % 2 ** (n + 1)
+def trunc_poly(poly: sp.Expr, trunc_bounds: tuple[int, int]) -> sp.Expr:
```
renezander90: We should mark this as internal helper `_trunc_poly`

src/qrisp/qtypes/quantum_float.py
```diff
@@ -355,43 +395,103 @@ def __init__(self, msize, exponent=0, qs=None, name=None, signed=False):
...
-        # represent.
+    def mshape(self) -> tuple[int | Array, int | Array]:
+        """The (log2(min), log2(max)) bounds of the absolute values this QuantumFloat can represent.
+
+        Returns
+        -------
+        tuple[int, int]
+            The (minimal, maximal) exponent of the representable magnitude.
+
+        """
```
renezander90:
```suggestion 
       """Return the mantissa's significance bounds.

        For a ``QuantumFloat`` with exponent ``k`` and mantissa size ``n``,
        the result is ``(k, k + n)``. The lower bound is inclusive and the
        upper bound is exclusive: mantissa qubit ``i`` has significance ``k + i``.

        Returns
        -------
        tuple[int | jax.Array, int | jax.Array]
            The inclusive lower and exclusive upper significance bounds.

        """
```

src/qrisp/qtypes/quantum_float.py
```diff
@@ -355,43 +395,103 @@ def __init__(self, msize, exponent=0, qs=None, name=None, signed=False):
...
         Also validates that the input value can be represented within the bounds of the provided
         QuantumFloat in static mode.
+
+        .. note::
+
+            Unlike the base :meth:`QuantumVariable.encoder
+            <qrisp.QuantumVariable.encoder>`, this parameter is named ``i``
+            (not ``value``) for historical reasons specific to QuantumFloat.
+
+            Unlike the base's ``ScalarLike``, this does not accept ``complex``:
```
renezander90: I think the `base` is not typed at all currently? And it could also accept `str`(see `QuantumChar`). I would remove this part of the note. It should be cleat that one cannot instantiate a `QuantumFloat` with a complex number.

src/qrisp/qtypes/quantum_float.py
```diff
@@ -455,201 +553,259 @@ def sb_poly(self, m=0):
...
 
-        super().encode(encoding_number, permit_dirtyness=permit_dirtyness)
+            Unlike the base :meth:`QuantumVariable.encode
+            <qrisp.QuantumVariable.encode>`, this method additionally accepts
+            ``rounding`` (inserted before ``permit_dirtyness``, for
+            historical reasons specific to QuantumFloat).
 
-    @gate_wrap(permeability="args", is_qfree=True)
-    def __mul__(self, other):
+            Unlike the base's ``ScalarLike``, this does not accept ``complex``
```
renezander90: Same comment as for `encoder`.

purva-thakre:
@PietropaoloFrisoni Does this also fix https://github.com/eclipse-qrisp/Qrisp/issues/842 ?

src/qrisp/qtypes/quantum_float.py
```diff
@@ -455,201 +553,259 @@ def sb_poly(self, m=0):
...
-            else:
-                output_qf = QuantumFloat(
-                    self.msize + int(np.ceil(np.log2(abs(other)))),
-                    self.exponent,
-                    signed=False,
-                )
+            output_qf = QuantumFloat(
+                self.msize + int(np.ceil(np.log2(abs(other)))),
+                self.exponent,
+                signed=bool(self.signed or other < 0),
```
renezander90, line 645: This raises an interesting question (for the future): `signed` is static. Hence, Jasp multiplication cannot make this distinction.

src/qrisp/qtypes/quantum_float.py
```diff
@@ -455,201 +553,259 @@ def sb_poly(self, m=0):
...
+            Suppresses the error message when calling encode on dirty
+            qubits. The default is False.
+
+        Returns
+        -------
+        None
+
+        """
+        value = encoding_number
+        if rounding:
```
renezander90, line 600: Wouldn't it be a better policy to always round/truncate?
This would also avoid the situation, that we cannot raise an Error in tracing mode, if the value that should be encoded does not fit. (Likely out of scope for this PR). @purva-thakre @positr0nium

src/qrisp/qtypes/quantum_float.py
```diff
@@ -975,51 +1171,55 @@ def significant(self, k):
...
+            mode (depending on whether the exponent is non-negative, see
+            :meth:`decoder <qrisp.QuantumFloat.decoder>`), or a traced
+            ``jax.Array`` while tracing.
 
         Examples
         --------
-        We create a QuantumFloat and round a value to fit the encoder and subsequently
-        initiate:
+        We create a QuantumFloat and round a value to the closest one it can
+        represent. Note that directly encoding an unrepresentable value (like
```
renezander90, line 1196: Okay, this solves #842 @purva-thakre 

However, I think this "truncate silently" that is currently happening is not always "proper" truncation. This brings us back to the issue that it would be easier to always call `truncate` in the encoder (i.e. the rounding always happens).

src/qrisp/qtypes/quantum_float.py
PietropaoloFrisoni: You are right, in my head it was typed since I have a local branch with some type hints in it (`QuantumVariable` will hopefully be improved soon as we have an onboarding ticket for this)

src/qrisp/qtypes/quantum_float.py
PietropaoloFrisoni: This function is not actually used within this file at all, so it would be weird to mark it internal and only use it somewhere else. I turned it into an internal helper and moved it in the only file that uses it

src/qrisp/qtypes/quantum_float.py
PietropaoloFrisoni: I did not touch the default (non-rounding) path in this PR, so that is issue is not fully solved unfortunately. I only focused on improving the performance and correctness of the explicit `rounding=True` option.

If I understand correctly, closing that issue would mean changing the default behavior to reject non-representable values, which is probably a real compatibility change worth its own PR : )

src/qrisp/qtypes/quantum_float.py
PietropaoloFrisoni: I can't help with that as I am not familiar enough with quantum arithmetic (one of the components I still need to investigate properly). At least now I have a basic understanding of how encoding/decoding works and what multiplication, summation etc. with a `QuantumFloat` actually means

positr0nium:
> @renezander90 @purva-thakre  I will not request Copilot's reviews anymore since it's basically a rabbit hole. It keeps finding issues, and most of them are pre-existing edge cases that will never come up in practical usage, or they are out of scope for this PR (which was meant to be a quick refactoring but it already costed too much time and tokens)

https://github.com/user-attachments/assets/8795d639-4808-4529-bf43-38d3686e5686

positr0nium:
Jokes aside, I think we need a deeper refactoring to stop this whac-a-mole

PietropaoloFrisoni:
> Jokes aside, I think we need a deeper refactoring to stop this whac-a-mole

I think we should start from the most primitive objects in qrisp (such as `QuantumVariable`) and refactor them systematically:

- Applying `pylint`/`ruff`/`pyright` improvements as much as possible
- Applying type hints as much as possible
- Add proper test coverage for every piece of source code
- Revisit the entire structure and modularization of the test suite and testing pipeline
- Make sure that edge cases are considered and accounted for 
- (...)

Only the first 2 points made us discover so many bugs in this PR and in previous refactorings: 
- https://github.com/eclipse-qrisp/Qrisp/pull/770
- https://github.com/eclipse-qrisp/Qrisp/pull/671
- https://github.com/eclipse-qrisp/Qrisp/pull/526
etc.

One year ago, this would have been daunting for 3-4 people, but now we will soon be more, and we have AI tools. I think this is definitely an investment that is worth doing! I hope we can coordinate efforts in this direction with Fraunhofer as well @renezander90

src/qrisp/qtypes/quantum_float.py
```diff
@@ -334,12 +308,42 @@ class QuantumFloat(QuantumVariable):
...
 
-    def __init__(self, msize, exponent=0, qs=None, name=None, signed=False):
+    signed: bool
+    exponent: int | Array
+    traced_attributes: list[str]
+    static_attributes: list[str]
+
+    def __init__(
+        self,
+        msize: int | Array,
```
renezander90, line 330: The type hints a great! However, they could cause some confusion, and user could attempt to initialise a `QuantumFloat` with a "true" array. Maybe, we should add a note in the documentation that this is only relevant in JASP/tracing mode and refers to 0-d (traced) JAX arrays.

renezander90 (approved):
LGTM 🎊  (It might not solve all problems, but definitively a significant improvement. Thanks!)

purva-thakre:
@renezander90 @diehoq What are your thoughts on us working on a paper about Qrisp's quantum arithmetic module? I brought this up with @positr0nium @PietropaoloFrisoni @moar55. According to Raphael we have historically prioritized other things because the portions of quantum arithmetic shor's algorithm relies on work as expected; which is technically not a great method to gauge the health of a module. 

b0dl4g

The quantum arithmetic bugs we've been addressing are on an ad-hoc basis. I believe we should use this paper as motivation to resolve them comprehensively rather than continuing to deal with them one at a time. 

The paper could also be a good resource for the users who often stumble on `QuantumFloat` technically being `QuantumFixed`. 

Feel free to tag other folks who have worked on quantum arithmetic in Qrisp. Sorry, I do not know all their usernames.

PietropaoloFrisoni:
@purva-thakre I think it would be the perfect motivation/excuse to deep dive into this topic and to improve it in qrisp from scratch (that is, theoretical approach -> efficient implementation for each quantum type -> proper testing for each case that we plan to support)

renezander90:
@purva-thakre this sounds like a great idea! Yes, we always had a big place in our hearts for Shor's algorithm, and even the modular arithmetic that is outside of what is necessary for Shor has been less reliable (recently addressed by @diehoq ). Basically, this should happen while we are refactoring the arithmetic at the same time. (I.e., remove SBP arithmetic and reimplement the primitives that exclusively rely on that like quantum-classical multiplication, migrate all adders to a common interface and testing, consolidate the modular arithmetic, etc.) The goal should be to have the arithmetic in perfect shape and a paper about that (which at the same time can serve as an overview on quantum arithmetic in general) at some point next year.

PietropaoloFrisoni:
> Basically, this should happen while we are refactoring the arithmetic at the same time. (I.e., remove SBP arithmetic and reimplement the primitives that exclusively rely on that like quantum-classical multiplication, migrate all adders to a common interface and testing, consolidate the modular arithmetic, etc.)

We might also take the change to modularize efficiently the jasp pipeline v.s. eager Python code. Right now there are functions that are natively jittable, functions that are not jittable, function that are jittable with a separate pipeline for certain values of the arguments (and the logic is implemented at the same level with the `check_for_tracing_mode` check), etc. We could potentially revisit this structure and define a more consistent logic, at least for quantum arithmetic

src/qrisp/qtypes/quantum_float.py
PietropaoloFrisoni: I added this clarification. If you don't like it please let me know, otherwise I will merge this PR in the evening as you already approved : )

purva-thakre:
> We might also take the change to modularize efficiently the jasp pipeline v.s. eager Python code. Right now there are functions that are natively jittable, functions that are not jittable, function that are jittable with a separate pipeline for certain values of the arguments (and the logic is implemented at the same level with the check_for_tracing_mode check), etc. We could potentially revisit this structure and define a more consistent logic, at least for quantum arithmetic

Yes, this has been my plan as a part of working on #501

positr0nium:
Tagging also @bruno-ah-um

Issue #839, opened 2026-09-01
QuantumModulus with Cuccaro adder


renezander90:
### Priority

High

### Qrisp Version

0.9.7

### Python Version

3.11.15

### Operating System

macOS

### Short Description

Exception when using QuantumModulus with Cuccaro adder (found is scope of adding doctest workflow #838)

### Steps to Reproduce

```
from qrisp import QuantumModulus, cuccaro_adder

N = 13
a = QuantumModulus(N, inpl_adder = cuccaro_adder)
a[:] = 5
a *= 10
print(a)
```
> AttributeError: 'list' object has no attribute 'duplicate'

### Expected Behavior

Result of 5 * 10 % 13 = 11 should be returned.

### Actual Behavior

Exception is thrown.

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive |             |                 |
| T-002   | Negative |             |                 |
| T-003   | Boundary |             |                 |
| T-004   | Boundary |             |                 |

### Definition of Done

- [x] Root cause identified
- [ ] Fix implemented and peer-reviewed
- [ ] Regression test added
- [ ] All tests pass (CI green)
- [ ] Changelog entry added

renezander90:
**Root cause:** `QuantumModulus` passes a raw qubit list to the configured adder.

1. In jasp_montgomery.py, the combined target is created as:

```python
wqf = aux[:] + res[:]
```

This produces a `list[Qubit]`.

2. That list is passed to the custom adder at jasp_montgomery.py.

3. `cuccaro_adder` receives `(int, list[Qubit])`, but its classical-input branch calls `b.duplicate()` at cuccaro_adder.py. Lists do not have `duplicate()`. It also explicitly rejects non-`QuantumVariable` targets at cuccaro_adder.py.

This contradicts the documented in-place-adder interface, which allows `list[Qubit]` targets in utility.py.

Verified with the requested interpreter:

- Cuccaro: `AttributeError: 'list' object has no attribute 'duplicate'`
- Gidney: succeeds with `{11: 1.0}`
- Fourier: succeeds with `{11: 1.0}`
- Existing Cuccaro tests: `20 passed`, because they use `QuantumVariable` targets and do not cover the list-target contract.

The narrow fix is to make `cuccaro_adder` genuinely support `list[Qubit]` targets, rather than changing Montgomery multiplication.

purva-thakre:
[...]

PR #848, opened 2026-09-02
Cuccaro adder compatibility with all possible inputs

purva-thakre:
## Description

`cuccaro_adder` now works with `QuantumModulus` and other quantum types that pass plain qubit lists as targets, fixing #839 (`AttributeError: 'list' object has no attribute 'duplicate'`). It now accepts `QuantumVariable`, `list[Qubit]`, and `DynamicQubitArray` for both operands, and classical addends larger than the target register are truncated modulo `2**len(b)` so modulo addition behaves as documented. The adder body was refactored into small helper functions, tests were expanded to cover the new inputs and the issue regression, and a changelog entry was added. This branch also introduces a new `adder_utilities.py` module with a shared register check, which is intended as the starting point for a separate, future PR that consolidates utilities across all adder implementations.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No

## What was changed?

-
-

## How was it tested?

| Test-ID | Status |
|---------|--------|
| T-001   | ✅ / ❌ |
| T-002   | ✅ / ❌ |
| T-003   | ✅ / ❌ |
| T-004   | ✅ / ❌ |

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the Cuccaro adder for qubit lists and dynamic qubit arrays.
  - Classical addends larger than the target register are reduced modulo the register size.
  - Improved carry-in, carry-out, controlled-operation, and quantum-modulus arithmetic behavior.
  - Added clearer validation for unsupported register and carry types.
  - Fixed controlled operations so carry-out remains unchanged when control is disabled.

- **Documentation**
  - Expanded adder documentation with usage examples, supported input types, size behavior, and error conditions.

### DIFF
--- a/documentation/source/general/changelog/changelog-dev.rst
+++ b/documentation/source/general/changelog/changelog-dev.rst
@@ -69,6 +69,15 @@ Other New Features
 Bug Fixes
 ---------
 
+* Fixed :func:`cuccaro_adder <qrisp.cuccaro_adder>` to support arbitrary
+  quantum inputs, in particular ``list[Qubit]`` and ``DynamicQubitArray``
+  targets. Previously, a classical addend with a ``list[Qubit]`` target raised
+  ``AttributeError: 'list' object has no attribute 'duplicate'``, so
+  ``QuantumModulus(..., inpl_adder=cuccaro_adder)`` failed. Classical addends
+  larger than the target register are now truncated modulo ``2**len(b)``
+  (`PR #848 <https://github.com/eclipse-qrisp/Qrisp/pull/848>`_,
+  `issue #839 <https://github.com/eclipse-qrisp/Qrisp/issues/839>`_).
+
 * Fixed the precision of :meth:`get_unitary <qrisp.QuantumCircuit.get_unitary>`.
   Unitary matrices are now computed in ``complex128`` precision, removing the
   spurious ~1e-7 off-diagonal entries that previously appeared where a
--- a/src/qrisp/alg_primitives/arithmetic/adders/adder_utilities.py
+++ b/src/qrisp/alg_primitives/arithmetic/adders/adder_utilities.py
@@ -1,0 +1,34 @@
+# ********************************************************************************
+# * Copyright (c) 2026 the Qrisp authors
+# *
+# * This program and the accompanying materials are made available under the
+# * terms of the Eclipse Public License 2.0 which is available at
+# * http://www.eclipse.org/legal/epl-2.0.
+# *
+# * This Source Code may also be made available under the following Secondary
+# * Licenses when the conditions for such availability set forth in the Eclipse
+# * Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+# * with the GNU Classpath Exception which is
+# * available at https://www.gnu.org/software/classpath/license.html.
+# *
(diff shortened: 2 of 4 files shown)


renezander90 (dismissed):
LGTM 🎊

positr0nium:
**`c_out` is not gated on `ctrl`, so the carry is written even when the control is `|0>`**

`_apply_maj_gates` deliberately runs uncontrolled (only the UMA phase takes `ctrl`), so after the MAJ chain the top `a` qubit holds the true carry of `a + b` regardless of the control. `_apply_c_out` then copies it out unconditionally:

```python
def _apply_c_out(c_out, a):
    if c_out is not None:
        cx(a[-1], c_out)
```

On the `ctrl = |0>` branch `b` is correctly left alone, but `c_out` is flipped anyway.

Reproduction, which is `test_cuccaro_adder_static_cout_ctrl` with the single line `x(ctrl[0])` removed:

```python
from qrisp import QuantumFloat, QuantumBool, cuccaro_adder

a = QuantumFloat(3)
b = QuantumFloat(3)
a[:] = 6
b[:] = 6

c_out = QuantumBool()
ctrl = QuantumBool()          # left in |0>, so no addition should happen

cuccaro_adder(a, b, c_out=c_out, ctrl=ctrl)

assert b.get_measurement() == {6: 1.0}          # passes, b is untouched
assert c_out.get_measurement() == {False: 1.0}  # fails, measures {True: 1.0}
```

`6 + 6 = 12` overflows the 3-qubit register, so the carry is 1 and lands in `c_out` even though the addition was suppressed.

The classical-state case above is a wrong result on its own. The more damaging case is a `ctrl` in superposition: the same `cx` entangles `c_out` with the addends on the branch where the addition did not happen, so a later uncomputation of `c_out` will not close.

Both entry points are affected, the explicit `ctrl=` keyword and `with control(qbl): cuccaro_adder(a, b, c_out=c_out)` via `custom_control`.

Suggested fix, mirroring how `_apply_uma_gates` already handles the control:

```python
def _apply_c_out(c_out, a, ctrl):
    if c_out is not None:
        if ctrl is None:
            cx(a[-1], c_out)
        else:
            mcx([ctrl, a[-1]], c_out)
```

For reference, `gidney_adder` sidesteps the problem entirely by appending `c_out` as an extra MSB of `b`, so the controlled chain covers it without a separate copy.

Worth noting that the current suite cannot catch this: `test_cuccaro_adder_static_cout_ctrl` and `test_cuccaro_adder_dynamic_cout_ctrl` both turn the control on before calling, and `_mk_add.apply()` always calls `qbl.flip()`, so the dynamic sweeps only ever see `ctrl = |1>`. `test_cuccaro_adder_static_no_addition_ctrl` is the only `ctrl = |0>` case and it passes no `c_out`.

purva-thakre:
@coderabbit review

purva-thakre:
@coderabbit full review

purva-thakre:
@CodeRabbit help

purva-thakre:
@coderabbit full review

purva-thakre:
@coderabbit full review

Issue #798, opened 2026-08-11
Caching issue in trotterization


bastibock:
### Priority

Medium

### Responsible

Sebastian (FOKUS)

### Qrisp Version

0.9.6

### Python Version

independet of Python version

### Operating System

macOS

### Short Description

Each 'group_up' call (line 1808 and line 1824) in trotterization() is called for every trotter step but doesn't depend on t, only on the fixed operator structure. So they don't need to be run every time and the grouping does not need to be recomputed from scratch for every trotter step.

### Steps to Reproduce

```python
import time
from qrisp import QuantumVariable
from qrisp.operators import Z

N = 18  # dense coupling -> M = N(N-1)/2 = 153 two-qubit terms
H = sum(Z(i) * Z(j) for i in range(N) for j in range(i + 1, N))

U = H.trotterization()          # method="commuting_qw" (default)
qv = QuantumVariable(N)

t0 = time.perf_counter()
U(qv, t=1.0, steps=1)
t1 = time.perf_counter()
print(f"1st call:        {t1 - t0:.3f} s")

t0 = time.perf_counter()
for _ in range(10):
    U(qv, t=1.0, steps=1)
t1 = time.perf_counter()
print(f"10 more calls:   {t1 - t0:.3f} s  ({(t1 - t0) / 10:.3f} s/call)")
```

### Expected Behavior

Run 'group_up' once per trotter step

### Actual Behavior

'group_up' is run for every trotter step.

### Means of Verification

| Test-ID | Type     | Description | Expected Result |
|---------|----------|-------------|-----------------|
| T-001   | Positive |             |                 |
| T-002   | Negative |             |                 |
| T-003   | Boundary |             |                 |
| T-004   | Boundary |             |                 |

### Additional Context

I'm already working on a fix and will open up a PR on this soon.

### Definition of Done

- [x] Root cause identified
- [ ] Fix implemented and peer-reviewed
- [ ] Regression test added
- [ ] All tests pass (CI green)
- [ ] Changelog entry added

renezander90:
Changing this would require some substantial restructuring of the methods inside the QubitOperator. Please note that this caching problem is already solved with the Jasp pipeline, hence it has not been a priority to change this. For example:

```
from qrisp import *
from qrisp.operators import X, Y, Z

H = X(0)*X(1) + Y(1) + Z(2)

@qache
def U(qv, t, steps):
    H.trotterization(order=2)(qv, t, steps)


@terminal_sampling
[...]

PR #850, opened 2026-09-02
Fix trotter caching

bastibock:
## Description

Fixes two independent performance bottlenecks in COLD/LCD (`qrisp.algorithms.cold`): slow Hamiltonian *construction* (O(N^4) for a dense QUBO, before compilation even starts) and slow circuit *compilation* (redundant Trotter-grouping recomputation + generic per-term gate emission). Also fixes two latent correctness bugs in the `nc` (nested-commutator) AGP coefficient shape, surfaced while wiring in the compile-time fix.

## Related Issues

Related to #798 

## Type of Change

- [x] Bug Fix
- [x] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- **`qrisp.operators.qubit.QubitOperator`**
  - `trotterization()` (`commuting_qw`/`commuting` methods): cache term-grouping (`group_up`) results outside the per-call `trotter_step` closure instead of recomputing them on every Trotter timestep. Grouping depends only on the operator, not on `t`/`steps`.
  - New `QubitOperator.sum(operators)` classmethod: merges many operators in a single O(M) pass (M = total term count), instead of Python's built-in `sum()` folding pairwise via `__add__`, which copies the entire running-total dict on every call -- O(M^2). Preserves `__add__`'s existing near-zero-coefficient pruning (`threshold = 1e-9`).
- **`qrisp.algorithms.cold.fast_trotterization`** (new module)
  - `is_flat_ising_operator(H)` / `fast_trotterization(H, ...)`: drop-in replacement for `QubitOperator.trotterization()` that emits native `rx`/`ry`/`rz`/`rzz` gates directly (no per-term `QuantumEnvironment`/session-merge overhead) whenever `H` contains only identity, single-qubit Pauli, or `Z⊗Z` terms -- covers every built-in QUBO Hamiltonian. Transparently falls back to `H.trotterization(...)` for anything else, so it's always at least as general and at least as correct.
- **`qrisp.algorithms.cold.dcqo_problem.DCQOProblem`**
  - `apply_lcd_hamiltonian`/`apply_cold_hamiltonian`: `.trotterization()` calls replaced with `fast_trotterization(...)`.
  - Docstring: documented the fast-path behavior, `compile_U_cold`'s O(N^2 * N_steps) scaling, and replaced the class's own `H_prob` example with `QubitOperator.sum(...)` so it stops teaching the O(N^4) anti-pattern that started this investigation.
- **`qrisp.algorithms.cold.problems.QUBO`** (`create_COLD_instance`, `create_LCD_instance`)
  - `H_prob` construction (and, in `create_LCD_instance`, the `nested_commutators` AGP build) switched from nested `sum()` to `QubitOperator.sum(...)`.
  - Bug fix, `nc_uniform`: `alph = [N * [alph]]` -> `alph = [alph] * N`. Old code wrapped an N-element array one level too deep, so the uniform-AGP branch received a whole array where a scalar coefficient was expected.
  - Bug fix, `nc_nonuniform`: `alph = [solve_alpha(h, J, lam)]` -> `alph = solve_alpha(h, J, lam)`. Same double-wrap; `solve_alpha` already returns the length-N array that `coeffs[idx]` needs to index directly.
  - Both bugs were previously silent: the general `.trotterization()` path tolerated the malformed array shape without erroring (almost certainly computing wrong AGP coefficients for `agp_type="nc"`); `fast_trotterization`'s native gate calls correctly reject it with a `ValueError`, which is how they were caught (via full-suite regression testing after wiring in the compile-time fix, not by the fix itself).
- **Tests** (new)
  - `tests/algorithms_tests/cd_tests/test_fast_trotterization.py`: `is_flat_ising_operator` classification; fast-path vs. general-path statevector equivalence; exact-match fallback behavior for ineligible (ladder-operator) Hamiltonians.
  - `tests/algorithms_tests/cd_tests/test_qubo_instance_build.py`: `QubitOperator.sum` vs. naive `sum()` equivalence (including exact-zero-coefficient pruning); `create_COLD_instance`/`create_LCD_instance` `H_prob` equivalence against a naive-`sum()` reference; `create_LCD_instance(agp_type="nc")` equivalence (exercises both bug fixes above).

## How was it tested?
Ran the full `tests/algorithms_tests/cd_tests/` suite multiple times (COLD, LCD, fast-trotterization, QUBO-instance-build) locally: **18 passed, 0 failed**. Confirmed via `git stash` isolation that the two `nc` AGP failures were reproducible against the trotter-caching/`fast_trotterization` change alone (not the `QubitOperator.sum` work) before being fixed. Manually verified the corrected `dcqo_problem.py` docstring example produces a `terms_dict` identical to the original naive-`sum()` version.
Ran all tests locally.

| Test-ID | Status |
|---------|--------|
| T-001 (QubitOperator.sum matches naive sum, incl. zero-pruning) | ✅ |
| T-002 (create_COLD_instance H_prob matches naive build) | ✅ |
| T-003 (create_LCD_instance nc H_prob + AGP matches naive build, post bugfix) | ✅ |
| T-004 (fast_trotterization matches general path / correct fallback) | ✅ |

## Checklist


src/qrisp/algorithms/cold/_fast_trotterization.py
```diff
@@ -0,0 +1,140 @@
...
+                return False
+        elif n == 2:
+            if set(factor_dict.values()) != {"Z"}:
+                return False
+        else:
+            return False
+    return True
+
+
+def _emit_flat_trotter_step(H, qarg, t, steps, forward_evolution):
```
renezander90, line 61: Yes, this might be fast, but it also just applies the exponentials for the individual terms in the order they appear in the `terms_dict`. This can lead to significantly less efficient circuits (depth) compared to when grouping strategies are applied. 

The grouping should be fast (numba based) even for larger instances. Have you checked?

The overall logic for the `is_flat_ising_operator` case is much simpler, compared to all cases that we have to take into account in general. 

We should further investigate. It would be helpful to have a code example where this method is significantly faster than the integrated `trotterization`.

bastibock: I tried some tests for different sizes of Hamiltonians and the picture seems very consistent overall.
To give some more specific numbers, I created dense Ising Hamiltonians (N=60, seed=0) with:
```python
def build_dense_ising(N, seed):
    rng = np.random.default_rng(seed)
    J = rng.uniform(-1, 1, (N, N))
    J = (J + J.T) / 2
    h_field = rng.uniform(-1, 1, N)
    return (
        sum(X(i) for i in range(N))
        + sum(J[i][j] * Z(i) * Z(j) for i in range(N) for j in range(i + 1, N))
        + sum(h_field[i] * Z(i) for i in range(N))
    )
```

If I then compare the main branch with this branch I get the following numbers:

  | trotter_time | compile_time | raw depth | compiled depth
-- | -- | -- | -- | --
main | 92.71s | 9.70s | 930 | 1147
fix_trotter_caching | ~70-78s | ~5.7s | 930 | 1147

So the circuits stays the same and trotterization is a little faster. But not a huge improvement.
The speed-up in compile time is not consistent over different examples. The small trotter time speed-up seems consistent.
This looks different for counterdiabatic driving:

trotter_time | compile_time | raw depth | compiled depth
-- | -- | -- | --
main | 110.39s | 7.03s | 952 | 1169
fix_trotter_caching | 1.89s | 5.52s | 1093 | 1092

Here we can see a huge speed-up in trotterization time, and interesting behaviour for compiled and non-compiled depth.

So yes, the trotterization most definitely needs further investigation.
At least for CD the improvement in this PR seems quite good.

Suggestion: Don't change the trotterization for now but use the fast_trotterization for CD algorithms. Investigate trotterization further.

src/qrisp/algorithms/cold/_fast_trotterization.py
```diff
@@ -0,0 +1,142 @@
...
+        factor_dict = term.factor_dict
+        n = len(factor_dict)
+        if n == 0:
+            continue
+        elif n == 1:
+            if next(iter(factor_dict.values())) not in ("X", "Y", "Z"):
+                return False
+        elif n == 2:
+            if set(factor_dict.values()) != {"Z"}:
+                return False
```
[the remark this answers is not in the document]
renezander90: Yes, it's fast but does not use grouping which, will result in different (and possibly less efficient) unitaries in general. But, it's not a bug. Trotter unitaries are not unique, but depend on the term order.

documentation/source/general/changelog/changelog-dev.rst
```diff
@@ -37,6 +37,24 @@ Improvements
...
 
+- **Faster COLD/LCD circuit compilation and Hamiltonian construction**
+  :meth:`compile_U_cold <qrisp.cold.DCQOProblem.compile_U_cold>` and
+  :meth:`~qrisp.cold.DCQOProblem.run` no longer recompute Trotter term
+  grouping on every timestep, and Ising-type Hamiltonians (identity,
+  single-qubit Pauli, or :math:`Z \otimes Z` terms only — the case for every
+  built-in QUBO Hamiltonian) are now Trotterized via a new native-gate fast
+  path (``fast_trotterization``) that skips the general per-term
+  ``QuantumEnvironment``/session-merge machinery, falling back automatically
+  and transparently for any other operator. Separately, building large QUBO
```
renezander90:
```suggestion
  single-qubit Pauli, or :math:`Z \otimes Z` terms only — the case for every
  built-in QUBO Hamiltonian) are now Trotterized via a new fast path 
  (``fast_trotterization``). Separately, building large QUBO
```

We should keep changelog entries concise.

src/qrisp/algorithms/cold/_fast_trotterization.py
```diff
@@ -0,0 +1,142 @@
+# ********************************************************************************
+# * Copyright (c) 2024 the Qrisp authors
```
renezander90:
```suggestion
# * Copyright (c) 2026 the Qrisp authors
```

renezander90:
All files in `tests/algorithms_tests/cd_tests/` need a license header.

tests/algorithms_tests/cd_tests/test_fast_trotterization.py
```diff
@@ -0,0 +1,82 @@
...
+from qrisp.algorithms.cold._fast_trotterization import fast_trotterization, is_flat_ising_operator
+from qrisp.operators.qubit import A, C, P0, X, Y, Z
+
+
+def _up_to_global_phase_close(sv1, sv2, tol=1e-6):
+    overlap = np.vdot(sv1, sv2)
+    return abs(abs(overlap) - 1.0) < tol
+
+
+def test_is_flat_ising_operator():
```
renezander90, line 33: Please add a short one-line description docstring for each test.

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -57,6 +58,23 @@ class DCQOProblem:
         By default, the groundstate of the x-operator $\ket{-}^n$ is prepared.
 
 
+    Notes
+    -----
+    For performance, ``H_init``, ``H_prob``, ``A_lam``, and ``H_control`` are
+    Trotterized via :func:`fast_trotterization
```
renezander90: Since `_fast_trotterization` in an internal helper, I would not reference this function here explicitly.

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -308,6 +336,11 @@ def apply_cold_hamiltonian(self, qarg, N_steps, T, opt_params, CRAB=False):
         of the system Hamiltonian, the adiabatic gauge potential (AGP) and
         local pulses (given by ``H_control``) with optimized parameters.
 
+        ``H_init``, ``H_prob``, ``A_lam``, and (for COLD) ``H_control`` are each
+        Trotterized via :func:`fast_trotterization`, which automatically uses a
```
renezander90: I would not reference to an internal helper function.

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -372,6 +405,18 @@ def apply_cold_hamiltonian(self, qarg, N_steps, T, opt_params, CRAB=False):
...
         """Compiles the circuit that is created by the :meth:`apply_cold_hamiltonian <qrisp.cold.DCQOProblem.apply_cold_hamiltonian>` method.
 
+        Notes
+        -----
+        Compilation time is dominated by how many Trotter-step gates need to be
+        built and compiled, which scales like $O(N^2 \\cdot N_{steps})$ for a dense
+        QUBO. Ising-type Hamiltonians (the default for QUBO problems built via
+        :func:`create_COLD_instance <qrisp.algorithms.cold.problems.QUBO.create_COLD_instance>`)
+        are Trotterized via the fast native-gate path in
+        :func:`fast_trotterization <qrisp.algorithms.cold._fast_trotterization.fast_trotterization>`,
```
renezander90: Same as above.

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -372,6 +405,18 @@ def apply_cold_hamiltonian(self, qarg, N_steps, T, opt_params, CRAB=False):
...
+        which avoids per-term QuantumSession merging overhead that otherwise
```
renezander90: I would also avoid reference vague formulation referencing internal implementation details like "per-term QuantumSession merging overhead".

documentation/source/general/changelog/changelog-dev.rst
bastibock: Shortened the changelog entry.

carlotta-k (approved):

src/qrisp/algorithms/cold/dcqo_problem.py
```diff
@@ -255,9 +274,13 @@ def _precompute_opt_pulses(self, N_steps, T, t_list, N_opt, CRAB=False):
...
 
     def apply_lcd_hamiltonian(self, qarg, N_steps, T):
-        """Simulate the local counterdiabatic driving (LCD) Hamiltonian on a
-        quantum argument via trotterization. The LCD Hamiltonian consists
-        of the system Hamiltonian and the adiabatic gauge potential (AGP).
+        """Simulate the local counterdiabatic driving (LCD) Hamiltonian on a quantum argument via trotterization.
+
+        The LCD Hamiltonian consists of the system Hamiltonian and the adiabatic gauge potential (AGP).
+        ``H_init``, ``H_prob``, ``A_lam``, and (for COLD) ``H_control`` are each
+        Trotterized via :func:`fast_trotterization`, which automatically uses a
```
renezander90: There is still a reference to the internal `fast_trotterization` method. Is this intentional?

renezander90 (approved):

src/qrisp/algorithms/cold/dcqo_problem.py
bastibock: Good catch! I missed that one but removed it now. Thank you.

PR #820, opened 2026-08-26
Replacing headers with comments + proper docstrings across the codebase

PietropaoloFrisoni:
## Description

Raphael and I previously talked about this change and decided to postpone it after the latest release of qrisp (this happened during the previous cycle, so now would be the time to implement this). This suddenly became 'urgent' since `ruff` through `reviewdog` started to complain about this problem since it is technically a PEP violation. 
This PR converts every source file's Eclipse Public License header from a module-level docstring into a #-prefixed comment block, and ensures every file has a genuine one-line module docstring. The header was previously written as a docstring, which made ruff's pydocstyle checks treat the license box itself as the module's documentation and flag its formatting (D205, D212, etc.) on nearly every file in the package - this is what was polluting reviewdog with irrelevant warnings on unrelated PRs.

## Related Issues

Closes #
Related to #

## Type of Change

- [ ] Feature (new functionality)
- [ ] Change Request (modification of existing functionality)
- [ ] Bug Fix
- [x] Refactoring (no behavior change)
- [ ] Performance improvement
- [x] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [x] No

## What was changed?

- Converted the Eclipse Public License header in 396 source files from a triple-quoted docstring into a #-prefixed comment block, matching the format already established in src/qrisp/interface/{backend,batched_backend,job,measurement_result}.py and the provider/simulator backend files.
- Added a real one-line module docstring to every file that lacked one (312 authored from the file's actual contents; 17 had existing rich module documentation preserved verbatim, just reformatted to satisfy D205).
- Added a one-line module docstring to 7 files that had no license header at all and were therefore missing a docstring too, including one genuinely empty stray file (src/qrisp/operators/test.py).
- Ran ruff format --preview across every touched file to satisfy the reviewdog format check, and added a changelog entry.

## How was it tested?

No behavior changes - this only touches module-level docstrings/comments and formatting. Verified via:
- ruff check --select D100 src/ -> 0 errors (no file is missing a module docstring)
- ruff check src/ (lint) and ruff format --check --preview src/ (format), cross-checked line-by-line against this PR's diff -> 0 in-scope findings
- python -m py_compile on every touched file -> all pass
- import qrisp plus qrisp.jasp, qrisp.algorithms, qrisp.circuit, qrisp.operators, qrisp.block_encodings, qrisp.permeability, qrisp.simulator -> all import cleanly

| Test-ID | Status |
|---------|--------|
| N/A - no linked issue | - |

[...]

PR #866, opened 2026-09-04
Converting headers into docstrings (Part 2)

PietropaoloFrisoni:
This PR is simply the follow-up to [this PR](https://github.com/eclipse-qrisp/Qrisp/pull/820). The description and the changes are the same (therefore, in this special case, I avoid the entire PR template format).

This one simply converts all headers into comments (and adds a 1-line file description) in the test files.

After this PR, there are no more files that need to be modified.


PietropaoloFrisoni:
Hi @bastibock , is it ok if I merge this? I just forgot the test files in my previous PR 😅

PR #867, opened 2026-09-04
Refactor FOQCS-LCU tests: remove redundancy, randomness and dead code

renezander90:


tests/block_encodings_tests/test_block_encoding_foqcs_lcu.py
```diff
@@ -510,176 +462,34 @@ def main(BE):
...
+    manual = _post_selection(multi_measurement([operand] + ancillas))
+    rus = _sample_rus(be, psi, L)
 
     assert np.allclose(
-        [filtered_conditional[k] for k in sorted(filtered_conditional)],
-        [result_rus[k] for k in sorted(result_rus)],
+        [manual[key] for key in sorted(manual)],
+        [rus[key] for key in sorted(rus)],
         atol=1e-4,
     )
```
bastibock, line 495: This test might throw a similar error than the fix applies.
With the direct dict indexing we get a KeyError/shape mismatch, If key sets ever differ again (same crash as before only different place).

Suggestion below is the as Line 515 and following. (just used your approach there).
```suggestion
    keys = sorted(set(manual) | set(rus))
    assert np.allclose(
        [manual.get(key, 0) for key in keys],
        [rus.get(key, 0) for key in keys],
        atol=1e-4,
    )
```

bastibock (dismissed):
Thanks for the fast fix!

I really just have this one small suggestions everything else looks good to me 😌

bastibock (approved):

PR #882, opened 2026-09-14
Preliminary refactor of quantum_variable.py

moar55:
## Description
This PR is a first step towards refactoring the quantum_variable file.
Mainly the goal of the PR is to do general cleanup of quantum_variable  such as
explicitly declaring instance/class attributes of `QuantumVariable` as well as
adding missing typings in various places(will be added in other places). 
Most importantly this PR refactors the `QuantumVariable`'s `__init__ ` method and
delegates the qv's name generation to the `QuantumSession`.

**Note**:  I have used the same approach for name generation in `duplicate` that I used in `__init__`. This means if `name` argument is provided a numerical suffix should be added if the name already exists in the `QuantumSession`. Now since the quantum session there is a fresh one, this should never happen. This assumption that the `name` argument must be unique to the `QuantumSession` was explicit earlier(an exception is raised otherwise), now it is not really enforced. Now it is not for convenience of re-using the name generation method. Not sure if this is the best approach, I could also have a concrete check that the name is not a duplicate and raise an exception otherwise...

## Type of Change

- [x]  Change Request (modification of existing functionality) - Practically none but technically there is a slight change in logic for name generation in `duplicate`, see above note.
- [ ] Bug Fix
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI / Build

## Breaking Change?
- [ ] Yes
- [ ] No
- [x] Not sure - see above note.

## What was changed?

- Explicit instance/class attributes were added to `QuantumVariable` (`quantum_variable.py`)
- Typings added in the same file
- Some docstrings added
- Most importantly, name generation logic is now delegated to `QuantumSession`, and __init__ method is  more concise now.
- Necessary relevant changes were done in other files too

## How was it tested?
Some tests were ran locally. 
Will update this when CI passes.

## Checklist

## Reviewer Notes

- One change that required multiple file change, is the change of the attribute
`usre_given_name` to `is_fixed_name`. The rationale for changing the name is
that this attribute is used to determine if the QuantumVariable can be renamed or not. 
The earlier name doesn't seem to convey that semantics, but rather conveys why
the boolean is set to True or False which can be confusing at conditions
involving the attribute. Let me know if you disagree with this part ofc.
-  I set `reg` in `quantum_variable` to `Any` (should be a `list[Qubit] |
DynamicQubitArray`) because it is not possible to see
the methods that `DynamicQubitArray` exposes at compile-time. Maybe we need to
extend the compile-time typing of `DynamicQubitArray` accordingly, but for now I
think we must use an Any.
-  I think in the next steps it might make sense to set the attributes set in
`register_qv` explicitly in the __init__ file (perhaps by making `register_qv`
return them) although I am not sure how well that will work out.


src/qrisp/core/quantum_variable.py
```diff
@@ -1027,7 +1009,8 @@ def get_measurement(
             raise Exception("Tried to get measurement from deleted QuantumVariable")
 
         if self.size == 0:
-            return {"": 1.0}
+            # TODO: Create a const DecodedMeasurmentResult that describes this construct.
+            return {"": 1.0}  # pyright: ignore[reportReturnType]
```
moar55, line 1013: @PietropaoloFrisoni Thoughts about this?

PietropaoloFrisoni: I think this makes sense (we might create another constructor for `DecodedMeasurementResult`, but it makes sense to leave a TODO here and do it in a follow-up PR).

For context, this `DecodedMeasurementResult` was introduced recently in the backend refactoring process

moar55:
Next-follow up [PR](https://github.com/moar55/Qrisp/pull/1), will be opened against repo's `main` when this is merged.

PietropaoloFrisoni:
Thank you @moar55!

Do you mind if I ask for a Copilot review before I take a closer look? I recently started using it for my own PRs, and it seems to be quite useful (although it lacks context and/or hallucinates in some cases)

moar55:
> Thank you @moar55!
> 
> Do you mind if I ask for a Copilot review before I take a closer look? I recently started using it for my own PRs, and it seems to be quite useful (although it lacks context and/or hallucinates in some cases)

Sure :smile:

purva-thakre:
> Next-follow up [PR](https://github.com/moar55/Qrisp/pull/1), will be opened against repo's `main` when this is merged.

@moar55 You could stack your new PR on top of this. I am on a mission to make everyone use it 😅. In my defense, this should make each PR review short and self-contained. 

https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

moar55:
> > Next-follow up [PR](https://github.com/moar55/Qrisp/pull/1), will be opened against repo's `main` when this is merged.
> 
> @moar55 You could stack your new PR on top of this. I am on a mission to make everyone use it 😅. In my defense, this should make each PR review short and self-contained.
> 
> https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

I thought you can't stack prs from a fork :thinking:

purva-thakre:
> I thought you can't stack prs from a fork 🤔

Thanks for the reminder! I forgot you are not a committer yet.

src/qrisp/core/quantum_variable.py
```diff
@@ -470,12 +442,16 @@ def delete(self, verify=False, recompute=False):
...
         while i < len(QuantumVariable.live_qvs):
-            if QuantumVariable.live_qvs[i]() is None:
-                QuantumVariable.live_qvs.pop(i)
+            live_qv = QuantumVariable.live_qvs[i]()
+            if live_qv is None:
+                del QuantumVariable.live_qvs[i]
+                continue
+
+            if not live_qv.name:
                 continue
```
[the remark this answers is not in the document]
moar55: I change this whole loop in an upcoming PR, so I would rather not resolve this here right away.

src/qrisp/core/quantum_variable.py
```diff
@@ -505,6 +484,14 @@ def duplicate(self, name=None, qs=None, init=False, qubits=None):
...
             If set to True, the :meth:`init_from <qrisp.QuantumVariable.init_from>`
             method of the result will be called on self. The default is False.
+        qubits: list[Qubit] | DynamicQubitArray, optional
+            Qubits to inialize the duplicated quantum variable from.
+            If not provided ``self.size`` is used for qubits extraction.
+
+        Raises
+        ------
+        QuantumVariableNamingError
+            Name of qv is already used in this QuantumSession.
```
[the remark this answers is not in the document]
moar55: I changed to a more general statement since merging can raise for other reasons...

tests/core_tests/test_variable_naming.py
```diff
@@ -0,0 +1,230 @@
...
+* This Source Code may also be made available under the following Secondary
+* Licenses when the conditions for such availability set forth in the Eclipse
+* Public License, v. 2.0 are satisfied: GNU General Public License, version 2
+* with the GNU Classpath Exception which is
+* available at https://www.gnu.org/software/classpath/license.html.
+*
+* SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+********************************************************************************
+"""
+
```
PietropaoloFrisoni, line 17:
```suggestion
# ********************************************************************************
# * Copyright (c) 2026 the Qrisp authors
# *
# * This program and the accompanying materials are made available under the
# * terms of the Eclipse Public License 2.0 which is available at
# * http://www.eclipse.org/legal/epl-2.0.
# *
# * This Source Code may also be made available under the following Secondary
# * Licenses when the conditions for such availability set forth in the Eclipse
# * Public License, v. 2.0 are satisfied: GNU General Public License, version 2
# * with the GNU Classpath Exception which is
# * available at https://www.gnu.org/software/classpath/license.html.
# *
# * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
# ********************************************************************************

```

src/qrisp/core/quantum_variable.py
```diff
@@ -230,124 +277,54 @@ def __init__(self, size, qs=None, name=None):
...
-                raise Exception("Tried to trace Qrisp code using make_jaxpr (use make_jaspr instead)")
+        from qrisp.core.quantum_session import QuantumSession
 
-            self.qubit_cache = {}
-        elif qs is not None:
-            self.qs = qs
+        if isinstance(self.qs, QuantumSession):
+            declaration_stack_level = 1 if type(self) is QuantumVariable else 2
+            (self.name, self.is_fixed_name) = self.qs.generate_name(name, self, declaration_stack_level + 1)
         else:
```
PietropaoloFrisoni, line 308: could this be a single `elif name is None` directive? I believe `ruff` is flagging the same

src/qrisp/core/quantum_variable.py
```diff
@@ -540,39 +526,25 @@ def duplicate(self, name=None, qs=None, init=False, qubits=None):
...
+            (duplicate.name, duplicate.is_fixed_name) = new_qs.generate_name(
+                name if name is not None else self.name,
+                duplicate,
+                declaration_stack_level,
+                name is None,
+            )
         else:
-            duplicate.user_given_name = False
+            duplicate.name = name if name is not None else self.name + "_dupl"
+            duplicate.is_fixed_name = False
```
PietropaoloFrisoni, line 540: Should this mirror `__init__`'s handling? 

That is:

```
`is_fixed_name = not name.endswith("*")` / `name.removesuffix("*")`
```

src/qrisp/core/quantum_variable.py
```diff
@@ -955,7 +927,10 @@ def get_measurement(
         filename=None,
         precompiled_qc=None,
```
PietropaoloFrisoni, line 928: `precompiled_qc` is missing in the docstring (this is a pre-existing problem since this function is a bit messy)

src/qrisp/core/quantum_variable.py
```diff
@@ -505,6 +486,14 @@ def duplicate(self, name=None, qs=None, init=False, qubits=None):
         init : bool, optional
             If set to True, the :meth:`init_from <qrisp.QuantumVariable.init_from>`
             method of the result will be called on self. The default is False.
+        qubits: list[Qubit] | DynamicQubitArray, optional
+            Qubits to initialize the duplicated quantum variable from.
+            If not provided ``self.size`` is used for qubits extraction.
+
+        Raises
+        ------
+        Exception
```
PietropaoloFrisoni, line 495: Although AI technically considers this a breaking change, I think we should try to replace a generic `Exception` with a better error message class which eventually inherits from it (`RuntimeError`, `AttributeError`, `NotImplementedError`, etc.).

It must be said though that this can make the JAX tests blowup (a strange phenomenon I observed once and I never fully understood why), but this is a very rare case

src/qrisp/core/quantum_variable.py
```diff
@@ -490,8 +468,11 @@ def is_deleted(self):
                 return True
         return False
 
-    def duplicate(self, name=None, qs=None, init=False, qubits=None):
-        r"""Duplicates the QuantumVariable in the sense that a new QuantumVariable is
+    def duplicate(self, name: str | None = None, qs=None, init=False, qubits=None):
```
PietropaoloFrisoni, line 471: Can we add type hints for the other parameters?

If you believe we should add more custom types to the `qrisp.typing` module, please let me know

src/qrisp/core/quantum_variable.py
```diff
@@ -230,124 +277,54 @@ def __init__(self, size, qs=None, name=None):
...
-            # If this didn't work, generate a generic, unique name
-            if not name_found:
-                while True:
-                    try:
-                        self.name = self.get_unique_name()
-                        self.qs.register_qv(self, size)
-                        break
-                    except QuantumVariableNamingError:
-                        pass
+        from qrisp.jasp.tracing_logic import flatten_qv, unflatten_qv
```
PietropaoloFrisoni, line 317: Is this imported here because of a circular import issue? If yes, can we please leave a NOTE or comment? Might be useful when we will address this issue in a more general PR

src/qrisp/core/quantum_variable.py
```diff
@@ -26,12 +29,43 @@
 from qrisp.core.compilation import qompiler
 
 if TYPE_CHECKING:
+    from qrisp.core.quantum_session import QuantumSession
     from qrisp.interface.measurement_result import DecodedMeasurementResult
+    from qrisp.jasp import TracingQuantumSession
+
+
+def extract_quantum_session(qs: QuantumSession | None) -> QuantumSession | TracingQuantumSession:
```
PietropaoloFrisoni, line 37: Yes, we definitely need to resolve this entanglement issue with the JAX compilation pipeline.

Right now we have a lot of functions that return completely different objects at runtime depending on whether we are tracing or not, which makes everything more difficult to maintain.

This is a long standing issue and of course we should not address it here, but thanks for raising the problem

src/qrisp/core/quantum_variable.py
```diff
@@ -470,12 +447,13 @@ def delete(self, verify=False, recompute=False):
 
         i = 0
         while i < len(QuantumVariable.live_qvs):
-            if QuantumVariable.live_qvs[i]() is None:
-                QuantumVariable.live_qvs.pop(i)
+            live_qv = QuantumVariable.live_qvs[i]()
```
PietropaoloFrisoni, line 450: Thanks for the fix! Is this tested?

src/qrisp/core/quantum_session.py
```diff
@@ -189,35 +195,146 @@ def __init__(self, backend=None):
...
         """
-        if qv.name in [temp_qv.name for temp_qv in self.qv_list + self.deleted_qv_list]:
-            raise QuantumVariableNamingError("Variable name " + str(qv.name) + " already exists in quantum session")
+        if not self._is_fresh_name(qv.name):
+            raise QuantumVariableNamingError(f"Variable name {qv.name} already exists in quantum session")
 
-        # Hand qubits to quantum variable
-        qv.reg = self.request_qubits(size, name=qv.name)
+        # Hand qubits to quantum variable if size is provided.
+        if size is not None:
```
PietropaoloFrisoni, line 330: Thanks for the fix (now we should have parity between `TracingQuantumSession` and `QuantumSession`). Is this already tested?

src/qrisp/core/quantum_session.py
```diff
@@ -31,7 +33,11 @@
 )
 from qrisp.core.quantum_variable import QuantumVariable
 from qrisp.core.session_merging_tools import multi_session_merge
-from qrisp.misc import get_depth_dic
+from qrisp.misc import find_calling_line, get_depth_dic
+
+
+class QuantumVariableNamingError(Exception):
```
PietropaoloFrisoni, line 39: It's a bit strange that we have such a custom error only for the quantum variable name (why not for many other cases)?. Still, this was pre-existing and certainly not a problem is this PR

src/qrisp/core/quantum_variable.py
```diff
@@ -884,12 +856,12 @@ def extend(self, amount, position=-1):
...
             for i in range(amount):
-                insertion_qubits[i].identifier = (
-                    self.name + "_ext_" + str(self.qs.qubit_index_counter[0]) + "." + str(self.size)
+                insertion_qubits[i].identifier = (  # pyright: ignore
+                    self.name + "_ext_" + str(self.qs.qubit_index_counter[0]) + "." + str(self.size)  # pyright: ignore
                 )
                 self.reg.insert(position + i, insertion_qubits[i])
 
-    def reduce(self, qubits, verify=False):
+    def reduce(self, qubits: Any, verify=False):
```
PietropaoloFrisoni, line 864: I introduced a `QubitLike` type in the typing module precisely for such case(s). Can we use it here, or does `pyright`/`pylance` complain?

src/qrisp/core/quantum_variable.py
```diff
@@ -1012,6 +991,11 @@ def get_measurement(
         {1.0: 0.5, 3.0: 0.5}
 
         """
+        from qrisp.jasp import TracingQuantumSession
```
PietropaoloFrisoni, line 994: Same comment as above regarding the nested import

src/qrisp/core/quantum_variable.py
```diff
@@ -1012,6 +991,11 @@ def get_measurement(
...
+
+        if isinstance(self.qs, TracingQuantumSession):
+            raise Exception("Tried to get measurement of a QuantumVariable in tracing mode")
```
PietropaoloFrisoni, line 997: Same comment as above, I think here we should raise a more specific error than a generic `Exception`

src/qrisp/core/quantum_variable.py
```diff
@@ -1307,6 +1387,12 @@ def uncompute(self, do_it=True, recompute=False):
 
 
         """
+
+        from qrisp.jasp import TracingQuantumSession
+
+        if isinstance(self.qs, TracingQuantumSession):
+            raise Exception("Tried to uncompute a QuantumVariable in tracing mode")
```
PietropaoloFrisoni, line 1394: same as above

tests/core_tests/test_variable_naming.py
PietropaoloFrisoni, line 1: In general I would like to keep the test files matching the files in the source code. That is, every test should be called `test_<name of the source file>`. But in this case I think it makes sense to have a test file just for naming. I think the issue is that these classes `QuantumVariable`, `QuantumCircuit`, etc. are too large, so a test file for each one would be way, way too large at present.

src/qrisp/core/quantum_variable.py
PietropaoloFrisoni, line 989: I tried to run this example but it's broken as the parameter is `compilation_kwargs` instead of `transpilation_kwargs`. Can we please fix this?

src/qrisp/core/quantum_variable.py
```diff
@@ -26,12 +29,43 @@
...
 if TYPE_CHECKING:
+    from qrisp.core.quantum_session import QuantumSession
     from qrisp.interface.measurement_result import DecodedMeasurementResult
+    from qrisp.jasp import TracingQuantumSession
+
+
+def extract_quantum_session(qs: QuantumSession | None) -> QuantumSession | TracingQuantumSession:
+    """Determine the QuantumSession a new QuantumVariable should be registered in.
+
+    While tracing (e.g. within jasp), the active :class:`TracingQuantumSession`
```
PietropaoloFrisoni, line 40: I tried building the documentation but `:class:`TracingQuantumSession` is a broken link (the class is not part of the documentation). The same link appears below

PietropaoloFrisoni (commented):
Thanks a lot @moar55 , this is great 🚀

tests/core_tests/test_variable_naming.py
moar55: I don't get this suggestion.
Why should I put '#' at the start of each line.

src/qrisp/core/quantum_variable.py
moar55: Yes good point, but `is_fixed_name = name is None or name.endswith("*")` *

src/qrisp/core/quantum_session.py
moar55: This should be handled in a diff pr with the design we agreed on offline.

src/qrisp/core/quantum_variable.py
moar55: I handled this in a follow [pr](https://github.com/moar55/Qrisp/pull/2/changes#diff-f4aae05bc295e8f9598abf9d65c633ac896d435d64f326b7bf0c4ca79847e530), btw lemme know if you think those should be merged together.

src/qrisp/core/quantum_session.py
moar55: Also about the test, there are tests for register_qv with None in the tracing_quantum_session variant. But I didn't add an explicit one should I?

src/qrisp/core/quantum_variable.py
moar55: This is only executed in static mode if i understand correctly and i typed it properly in the affermentioned [pr](https://github.com/moar55/Qrisp/pull/2)

src/qrisp/core/quantum_variable.py
moar55: This change shouldn't affect logic, but I don't see tests covering `delete`. Should I add ones?

tests/core_tests/test_variable_naming.py
moar55: Yeah we could have further modules in the future...

src/qrisp/core/quantum_variable.py
moar55: It is necessary for circular import indeed.
The __init__.py of tracing_logic imports tracing_quantum_session which in turn imports QuantumVariable.

src/qrisp/core/quantum_variable.py
moar55: I agree, but in this case the exception is raised by merge so the change should be there which is out of the scope of this pr.
Also i noticed init_from could raise if `qubits` are passed to `duplicate` that are not fresh or if `init`  is set to true when `duplicate` is called in tracing mode. I am not sure if this is clear enough to the user :thinking:  Maybe we should add a big warn in the docstring in the arguments section?

src/qrisp/core/quantum_variable.py
moar55: Necessary as explained above, but will add a comment.

